PR #1, opened 2019-09-26
Refactor modules and integrate JavaCPP to map the C API

saudet:
This basically integrates JavaCPP in the way discussed previously at https://github.com/karllessard/tensorflow-java/pull/2, in addition to building with MKL enabled and upgrading to TF 2.0.0-rc2. (Maybe `src/gen/java` is too much? It should change only little by little from now on though.)

I also took the liberty to refactor tentatively the modules a bit, especially the naming. The module that launches Bazel is "tensorflow-core-api", whose parent is "tensorflow-core", so we don't end up with something odd like "parent-core", which also doesn't match with the subdirectory name. Other modules in "tensorflow-core" also use it as prefix, "tensorflow-core-platform" and "tensorflow-core-examples", which contains the LabelImage example that I moved there and a HelloWorld one for the C API. I've also moved the single source file from "annotation-processor" to "tensorflow-core-api" since we can execute all that in the same module. What do you think about that? It's not supposed to be used for anything else, right?

This naming scheme also works for others: "tensorflow-utils" -> "tensorflow-utils-nio", "tensorflow-frameworks" -> "tensorflow-frameworks-keras", but we can still deviate a bit and do something like "tensorflow-keras". (BTW, I think we should use a different name than "keras", like "kerasj", for example.) If people follow even approximately these rules, I think we should get a pretty nice and consistent repository such as the one for Spring Boot: https://github.com/spring-projects/spring-boot

The project builds and runs fine on Linux and Mac here, but I still get a weird compiler error on Windows. I'll eventually get it working, but other than that I think it's ready for a review! (BTW, I had to copy `.bazelrc` in there. Is there a better way to achieve this?)


saudet:
@googlebot I signed it!

karllessard:
> I also took the liberty to refactor tentatively the modules a bit, especially the naming. The module that launches Bazel is "tensorflow-core-api", whose parent is "tensorflow-core", so we don't end up with something odd like "parent-core", which also doesn't match with the subdirectory name. Other modules in "tensorflow-core" also use it as prefix, "tensorflow-core-platform" and "tensorflow-core-examples", which contains the LabelImage example that I moved there and a HelloWorld one for the C API. 

I like those new names as well. But I would remove completely the `tensorflow-core-examples` folder, I was planning to move those samples to our second repo (github.com/tensorflow/java-models).

> I've also moved the single source file from "annotation-processor" to "tensorflow-core-api" since we can execute all that in the same module. What do you think about that? It's not supposed to be used for anything else, right?

Like we discussed on the call, I'm pretty sure the annotation processor needs to be in its own artifact but we need to depend on it when building the `tensorflow-core-api` jar. See how it is done [here](https://github.com/tensorflow/java/blob/ddf0b569368966ebadc8abb852c2c08b16a20014/core/tensorflow-core/pom.xml#L19) and [here](https://github.com/tensorflow/java/blob/ddf0b569368966ebadc8abb852c2c08b16a20014/core/tensorflow-core/pom.xml#L70). Let me know you find another way to make it work, but I suggest that we run that processor outside Bazel even in that first phase.


> (BTW, I had to copy `.bazelrc` in there. Is there a better way to achieve this?)

Looks like TensorFlow Serving [is doing the same](https://github.com/tensorflow/serving/blob/master/.bazelrc) so I guess it is OK?

tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -0,0 +1,31 @@
...
+# Allows us to use ccache with Bazel
+export BAZEL_USE_CPP_ONLY_TOOLCHAIN=1
+export BAZEL_VC="C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/"
+export PYTHON_BIN_PATH=$(which python.exe)
+if [[ ! -f $PYTHON_BIN_PATH ]]; then
+    export PYTHON_BIN_PATH=$(which python3)
+fi
+
+# Build TensorFlow itself
+bazel build --python_path="$PYTHON_BIN_PATH" --config=mkl --output_filter=DONT_MATCH_ANYTHING --verbose_failures @org_tensorflow//tensorflow:tensorflow @org_tensorflow//tensorflow/java:tensorflow
```
karllessard: Can we build only targets that we actually need for the Java part? We don't have to build the whole client, afaik we just need the operation wrappers, which are provided by this target `//tensorflow/java:java_op_gen_sources`, in replacement to `//tensorflow/java:tensorflow`.

tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -0,0 +1,31 @@
...
+TENSORFLOW_DYLIB=(bazel-bin/external/org_tensorflow/tensorflow/libtensorflow.?.?.?.dylib)
+if [[ -f $TENSORFLOW_DYLIB ]]; then
+    ln -sf $(basename $TENSORFLOW_DYLIB) bazel-bin/external/org_tensorflow/tensorflow/libtensorflow.dylib
+    ln -sf $(basename $TENSORFLOW_DYLIB) bazel-bin/external/org_tensorflow/tensorflow/libtensorflow.2.dylib
+fi
+
+# Copy generated Java source files
+mkdir -p src/gen/java/
+cp -r bazel-genfiles/external/org_tensorflow/tensorflow/java/ops/src/main/java/* src/gen/java/
+cp -r bazel-genfiles/external/org_tensorflow/tensorflow/java/_javac/tensorflow/libtensorflow_sourcegenfiles/* src/gen/java/
```
karllessard: This last line should be remove if we run the annotation processor from Maven instead (see [this comment](https://github.com/tensorflow/java/pull/1#issuecomment-536228429))

tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -0,0 +1,321 @@
...
+      <artifactId>guava</artifactId>
+      <version>28.0-jre</version>
+      <optional>true</optional> <!-- for compilation only -->
+    </dependency>
+    <dependency>
+      <groupId>com.squareup</groupId>
+      <artifactId>javapoet</artifactId>
+      <version>1.11.1</version>
+      <optional>true</optional> <!-- for compilation only -->
+    </dependency>
```
karllessard: javapoet and guava are just required for the annotation-processor, which should be move to its own artifact (unless you find a way to make it work where you put it).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/examples/BUILD
```diff
@@ -0,0 +1,30 @@
...
+            "**/METADATA",
+            "**/OWNERS",
+            # Exclude any .class files that might have been generated
+            # by manual invocations of javac as described in
+            # tensorflow/java/README.md
+            "**/*.class",
+        ],
+    ),
+    visibility = ["//tensorflow:__subpackages__"],
+)
```
karllessard: Let's remove that whole `org.tensorflow.examples` package

tensorflow-core/tensorflow-core-examples/src/main/java/org/tensorflow/examples/LabelImage.java
```diff
@@ -0,0 +1,235 @@
...
+    private <T> Output<T> binaryOp(String type, Output<T> in1, Output<T> in2) {
+      return g.opBuilder(type, type).addInput(in1).addInput(in2).build().<T>output(0);
+    }
+
+    private <T, U, V> Output<T> binaryOp3(String type, Output<U> in1, Output<V> in2) {
+      return g.opBuilder(type, type).addInput(in1).addInput(in2).build().<T>output(0);
+    }
+    private Graph g;
+  }
+}
```
karllessard: As discussed [here](https://github.com/tensorflow/java/pull/1#issuecomment-536228429), let's remove the `tensorflow-core-examples` artifact, we'll keep our samples in `github.com/tensorflow/java-models` instead.

In addition, this sample is super outdated and shouldn't be followed as an example anymore...

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Buffer.java
```diff
@@ -0,0 +1,69 @@
...
+        return b;
+    }
+
+    /**
+     * Calls the deallocator, if registered, otherwise has no effect.
+     */
+    public void delete() {
+        deallocate();
+    }
+}
```
karllessard: Is this class (and all other classes found in `src/main/java/org/tensorflow/c_api`) actually used by your PR? If not, I would prefer we wait for the second phase before adding them.

tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -0,0 +1,321 @@
...
+          <execution>
+            <id>javacpp-parser</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>parse</goal>
+            </goals>
+            <configuration>
+              <skip>${javacpp.parser.skip}</skip>
+              <outputDirectory>${project.basedir}/src/gen/java</outputDirectory>
+              <classOrPackageName>org.tensorflow.c_api.presets.*</classOrPackageName>
```
karllessard: The TF Java code is based upon the Google Java Style Guide, which states that package names [shouldn't have underscores in it](https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names)...

Can we rename that package to something else, like `nativeapi`?

(BTW would be great that generated class names also sticks to camelcase, with no underscores, but I guess that could be a challenge to change this, right? e.g `TFEContext` instead of `TFE_Context`, or `DeallocatorPointerLongPointer` instead of `Deallocator_Pointer_long_Pointer`)

karllessard:
> (Maybe `src/gen/java` is too much? It should change only little by little from now on though.)

Yeah, I'm still not sure what we should do with those generated files neither... not having them committed though is pretty annoying when you want to write unit tests that uses them, for example... let's keep it there and agree that we might change our mind later.

tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -0,0 +1,321 @@
...
+                <include>org/tensorflow/processor/operator/OperatorProcessor.java</include>
+              </includes>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.bytedeco</groupId>
+        <artifactId>javacpp</artifactId>
+        <version>1.5.2-SNAPSHOT</version>
```
karllessard: Do we all have access to JavaCPP 1.5.2-SNAPSHOT? Can we use a released version of JavaCPP available on Maven central instead?

pom.xml
```diff
@@ -4,11 +4,11 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>org.tensorflow</groupId>
-  <artifactId>parent</artifactId>
+  <artifactId>tensorflow-java</artifactId>
   <version>2.0.0-SNAPSHOT</version>
```
karllessard: It has been suggested in our last call that our new artifacts should start from the beginning, in alpha-mode, as we won't guarantee complete stability for a while. I like this idea as well, what do you think?

Therefore, it would mean that `2.0.0-SNAPSHOT` be renamed to `0.1.0-SNAPSHOT` I guess?

karllessard:
An idea for the annotation processor artifact: when we'll migrate the Java operators generator to Java, we could probably move that code to the same artifact as this processor. So we can give it a more generic name, like `tensorflow-op-generator`

tensorflow-core/tensorflow-core-api/build.sh
saudet: No, we can't, unfortunately, that is, not without patching TensorFlow, those targets are not public. Though, I've had to add a patch for MKL, since the fix isn't in 2.0.0. We can add more patches if we want to go that way.

tensorflow-core/tensorflow-core-api/build.sh
saudet: We can, but what does it give us? Until everything is moved to Java, it seems like one more failure point that can get out of sync and that we'll need to watch for.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/examples/BUILD
saudet: Oops, that wasn't supposed to be there.

tensorflow-core/tensorflow-core-examples/src/main/java/org/tensorflow/examples/LabelImage.java
saudet: Hum, ok, I was thinking that we could leave it here until we move it somewhere else, but if we're going to delete it anyway, let's remove it I suppose?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Buffer.java
saudet: Well, they are the base classes of `TF_Buffer`, etc. They are functional, and we can modify them later as well. What are you worried about?

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Well, we can start customizing names like that yes, but the idea with JavaCPP is to map the underlying API to something that's as close as possible to the native API. If we start renaming things randomly like that, it increases the chances of conflicts and it makes it harder to understand the mapping. Besides, the C API is very hard to use, it's not meant to be used by end users, and making it even harder to use isn't going to help. :)

tensorflow-core/tensorflow-core-api/pom.xml
saudet: I think I can rebase them on a released version yes. In any case, all snapshots are available on Sonatype OSS as well: https://oss.sonatype.org/content/repositories/snapshots/org/bytedeco/

pom.xml
saudet: Yes, something like that. Do you want to do it as part of this PR?

tensorflow-core/tensorflow-core-api/pom.xml
saudet: AFAIK, it's working just fine, yes. If we uncomment the lines that start with `<!-- Bazel currently runs this`, it gets executed just fine. However, it outputs an error because it doesn't want to overwrite existing classes:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project tensorflow-core-api: Fatal error compiling: java.lang.AssertionError: javax.annotation.processing.FilerException: Attempt to recreate a file for type org.tensorflow.op.NnOps -> [Help 1]
```

saudet:
> An idea for the annotation processor artifact: when we'll migrate the Java operators generator from C++ to Java, we will probably move the code in the same artifact as this processor. So we can give it a more generic name, like `tensorflow-op-processor` or anything better you may find

Good point. We can make the name even more generic too, like `tensorflow-core-processor` to stuff there anything we don't need at runtime.

tensorflow-core/tensorflow-core-api/build.sh
karllessard: Yeah, I think we can have @sjamesr commit this very quickly in the main repository, I'll check with him but not super important neither, we don't have to wait for this.

tensorflow-core/tensorflow-core-api/build.sh
karllessard: The annotation processor does not depend on any native code and is actually already [working pretty well](https://github.com/tensorflow/java/tree/master/core/annotation-processor) outside Bazel right now, it is pure Java. I don't see any good reason for taking a step back at this stage just to restore it later, let's keep it instead, what we want from the Bazel build is the code generated by the C++ engine only.

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Yes, that's because you copied those files (`NnOps` & friends) in your bash script previously, if you remove that last line in the script I pointed out, it should be fine... so you are saying that the annotation-processor runs successfully directly from the core-api jar?

tensorflow-core/tensorflow-core-examples/src/main/java/org/tensorflow/examples/LabelImage.java
karllessard: yes let's remove it, like I said anyway, it is now a very bad example to look at.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Buffer.java
karllessard: I just don't find any additional value for having classes in the code base that are actually not used, it creates unnecessary confusion and they can be easily reviewed and added later as part of the whole migration of the JNI code.

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Since the classes won't be public, I'm ok to just rename the package as well.

pom.xml
karllessard: yeah, why not 👍

tensorflow-core/tensorflow-core-api/build.sh
karllessard: Just thought of something, we actually **need** the annotation processor to be ran outside Bazel, because otherwise it will not generate bindings based on the [op classes](https://github.com/tensorflow/java/tree/master/core/tensorflow-core/src/main/java/org/tensorflow/op/core) inside our repo but on those found in the old code base of the main repository, which is not what we want.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Yes, that's correct. If I delete `src/gen/java`, it ends up generating them in `target/generated-sources/annotations/`. Is there a way to change that?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Buffer.java
saudet: But they are used. Why do you say that they are not?

tensorflow-core/tensorflow-core-api/pom.xml
saudet: My thought was that having a package name like `c_api` makes it less likely for the casual to want to use it. I also thought about `capi`, but I just don't find it as obvious as `c_api`, which is used by TensorFlow in the docs making it clearer that it's related: https://www.tensorflow.org/install/lang_c

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Ah, here we go: https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#generatedSourcesDirectory So that should work alright...

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Ok then

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Buffer.java
karllessard: Oops, sorry I didn’t read carefully that last reply of yours, all good then.

So generated classes can extend from human-written ones? Just for my knowledge, where this magic is happening?

karllessard: Ok got it: https://github.com/saudet/tensorflow-java/blob/18ea0b7939442d4539a61c8c0a164bc15f9ead51/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/presets/tensorflow.java#L118

SidneyLann:
The java TF 2.0 available now? have more than 90% functionalities of TF 2.0?

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Ok, done! But I haven't been able to get it to output to `src/gen/java`. I tried to hack it a dozen ways, but `maven-compiler-plugin` is hard coded to not compile anything from the `generatedSourcesDirectory` directory, which makes sense I guess, but it would be nice if we could use the `includes` and `excludes` filters instead. Anyway, it's now outputting to `src/gen/annotations`. Maybe we can do the same with the others like `src/gen/javacpp` and `src/gen/ops`...

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Maybe let's get the opinion of at least one other person? In a lot of places where JavaCPP just uses `_`, I know Panama uses `$` because, you know, `$` isn't a valid character in C/C++ identifiers, but it is in Java!! (I wonder how they are going to get this passed the CSR group...) @Craigacp Any opinions about this?

Craigacp: I think for internal package names which we won't expose to the user then `_` is fine. Panama uses `$` for internally generated things because it guarantees no conflicts, but in this case the header file is actually called `c_api.h`, so mirroring the name is fine. Panama's `jextract` gives it the same name (see here https://hg.openjdk.java.net/panama/dev/raw-file/c359a9e944de/doc/panama_foreign.html#using-tensorflow-c-api-mac-os).

karllessard:
Good job @saudet , we are almost good to go, just want to follow up with you about [my previous suggestion](https://github.com/tensorflow/java/pull/1#issuecomment-536228429) to remove completely `tensorflow-core-examples` from this repo. 

I think it could be nice to have all the sample code in the `java-models` repo instead, to replicate what the core team is doing (sample code [is found](https://github.com/tensorflow/models/tree/master/samples) in `/tensorflow/models` instead of `/tensorflow/tensorflow`).

Otherwise, everything looks fine to me

karllessard:
> The java TF 2.0 available now? have more than 90% functionalities of TF 2.0?

@SidneyLann, I think the closest TF Java is to TF 2.0 is that it supports eager execution. Work is currently in progress to add a Keras-like API on top of it and and to handle functional graphs.

SidneyLann:
> > The java TF 2.0 available now? have more than 90% functionalities of TF 2.0?
> 
> @SidneyLann, I think the closest TF Java is to TF 2.0 is that it supports eager execution. Work is currently in progress to add a Keras-like API on top of it and and to handle functional graphs.

https://github.com/tensorflow/swift/blob/master/docs/WhySwiftForTensorFlow.md, this link said java can't be used for Graph Program Extraction because java can't do static analysis, need to use swift, can you talk about this problem?

Craigacp:
@SidneyLann Those are different issues. Functional graphs as implemented in TF2 in python are different from extracting a graph program directly from the Swift program. As I understand it the Swift one is more precise, whereas the Python one requires tracing the annotated functions to determine the currently executing codepath. It's possible to do the tracing in Java, that's kinda how eager mode works for backprop.

Java's type system isn't quite powerful enough to express the kind of generics you'd like to use for static compile time typing of tensors (including their shapes), but to get a type system that powerful requires lots of other features which make it harder to use and can lead to a lot of complexity. I'm not sure if Swift's is powerful enough for that, but the Swift for Tensorflow project encompasses changes to the Swift language and core libraries, which allows you to do more than is possible in a normal library.

saudet:
> Good job @saudet , we are almost good to go, just want to follow up with you about [my previous suggestion](https://github.com/tensorflow/java/pull/1#issuecomment-536228429) to remove completely `tensorflow-core-examples` from this repo.
> 
> I think it could be nice to have all the sample code in the `java-models` repo instead, to replicate what the core team is doing (sample code [is found](https://github.com/tensorflow/models/tree/master/samples) in `/tensorflow/models` instead of `/tensorflow/tensorflow`).
> 
> Otherwise, everything looks fine to me

Ah, I didn't realize you wanted to remove everything even my little "HelloWorld" :) I was thinking we could leave snippets and maybe even some documentation there for "core developers". Do you see a better place for those? They wouldn't go in another repo... Maybe the wiki? It doesn't appear to be enabled on TensorFlow repos though.

karllessard:
@saudet, I am discussing this with @tzolov, we’ll definitely have a page for developers, need to figure out where, but let’s keep that topic pending for now and remove this example artifact from the current PR, I don’t think your “Hello World” will be missed anyway :D 

But if you feel sentimental about it, maybe you could move it as an integration test in the `src/test` folder of the core-api until we figure this out?

karllessard:
... at the same time, it’s up to you, we can keep it there and move it later as well, let me know.

saudet:
Yes, unit tests as documentation, that works too. Ok, it's done!

PR #5, opened 2019-11-30
TensorFlow Tools

karllessard:
This PR merges the first draft of Tensor NIO into our master repository so that it can be tried out and tested by a larger audience. Here is a minimalist example of usage (more examples can be also found in the unit tests):

```java
    IntNdArray heapData = NdArrays.vector(0, 1, 2, 3);

    // Creates a 2x2 matrix
    try (Tensor<TInt32> tensor = TInt32.tensorOfShape(2, 2)) {
      IntNdArray tensorData = tensor.data();

      // Copy first 2 values of the vector to the first row of the matrix
      tensorData.set(heapData.slice(Indices.range(0, 2)), 0);

      // Copy values at an odd position in the vector as the second row of the matrix
      tensorData.set(heapData.slice(Indices.odd()), 1);

      assertEquals(0, tensorData.getInt(0, 0));
      assertEquals(1, tensorData.getInt(0, 1));
      assertEquals(1, tensorData.getInt(1, 0));
      assertEquals(3, tensorData.getInt(1, 1));

      // Read rows of the tensor in reverse order
      IntNdArray reversedTensorData = tensorData.slice(Indices.all(), Indices.flip());

      assertEquals(1, reversedTensorData.getInt(0, 0));
      assertEquals(0, reversedTensorData.getInt(0, 1));
      assertEquals(3, reversedTensorData.getInt(1, 0));
      assertEquals(1, reversedTensorData.getInt(1, 1));

      try (EagerSession session = EagerSession.create()) {
        Ops tf = Ops.create(session);

        // Compute the power of the tensor by itself
        Constant<TInt32> x = tf.constant(tensor);
        IntNdArray result = tf.math.pow(x, x).tensorData();

        // Validate result by computing the same operation in Java
        tensorData.scalars().forEachIndexed((coords, s) ->
          assertEquals(Math.pow(s.getInt(), s.getInt()), result.getInt(coords), 1e-7f)
        );
      }
    }
```

Some additional notes:
* The major breaking feature is that now, the `Tensor` class does not carry a Java boxed type, like `Integer`, but a custom tensor type, like `TInt32`. This allows us to support more types in Java while still guaranteeing compile-time safety check of operand type compatibility. A lot of the changes seen in this PR are related to this.
* The `c_api` package has been moved to `org.tensorflow.internal`, next to the new package `buffer` that holds new Tensor buffer mapping utilities.
* Right now, it looks like we do not control the order we process the `@Operator` annotation for generating `*Ops` classes. For this reason, methods are moved within the classes at each compilation, as you can observe in this PR.


tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -31,6 +31,11 @@
       <version>${project.version}</version>
       <optional>true</optional> <!-- for compilation only -->
     </dependency>
+    <dependency>
+      <groupId>org.tensorflow</groupId>
+      <artifactId>nio-utils</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
saudet: That feels strange. Why do we need this dependency here? Is it just because that we can't have a separate module until we move away from the code generator in C++?

tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -150,6 +151,7 @@
             <compilerOption>${project.basedir}/src/main/native/server_jni.cc</compilerOption>
             <compilerOption>${project.basedir}/src/main/native/session_jni.cc</compilerOption>
             <compilerOption>${project.basedir}/src/main/native/tensorflow_jni.cc</compilerOption>
+            <compilerOption>${project.basedir}/src/main/native/tensor_buffers_jni.cc</compilerOption>
```
saudet: I thought everyone was OK with moving away from writing JNI manually. Why do we need this?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
```diff
@@ -0,0 +1,3 @@
+package org.tensorflow.types.family;
+
+public interface TType {}
```
saudet: Why do we need empty interfaces like this? That's always a code smell to me... They probably have *some* functionality in common. BTW, we have default methods too starting with Java SE 8. Can we use those?

tensorflow-utils/nio-utils/src/main/java/org/tensorflow/nio/buffer/DataBuffers.java
```diff
@@ -0,0 +1,519 @@
...
+  }
+
+  private static boolean unsafeAvailable;
+
+  static {
+    try {
+      Class.forName("sun.misc.Unsafe");
+      unsafeAvailable = true;
+    } catch (ClassNotFoundException e) {
+      unsafeAvailable = false;
```
saudet: This is misleading. `sun.misc.Unsafe` is often available in other implementations of Java SE, but it doesn't always contain all the methods from OpenJDK. We need to make sure it has all the methods we want like this:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/indexer/UnsafeRaw.java

saudet:
What's the copyright on the 1500x916.jpg file? I didn't see it mentioned anywhere, but that needs to be clear to put anything in a repository from Google.

karllessard:
> Does that mean we can now rename `getShort()`, `getInt()`, `setByte()`, etc to just `get()` and `set()`? I think that would be a lot nicer.

On the `DataBuffer`, they could but on the `NdArray`, they are already being used for getting/setting subelements of an array. 

I was inspired by this: if you think that `get()/set()` are replaceable by brackets (like Kotlin does), for a given 2D matrix, it might be natural to retrieve its first row simply with `matrix[0]` instead of something like `matrix.getElement(0)`. On the other hand, if `get()/set()` were used to return a value type, then only the right number of brackets that leads to a scalar would be allowed (`matrix[0][0]` in this case), which is a bit awkward.

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: `Tensor` are now based on `NdArray` for mapping their memory and allowing direct access to it, so it must depend on `nio-utils`. I'm not sure to understand what is strange, you meant that memory mapping should occur outside the core API?

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Because I need to create an instance of `ByteBuffer` out of a tensor native address and size in case `Unsafe` is not available, and the only way I know of is to do it from JNI. Do you have another way to do it?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
karllessard: This is just to make sure that not all Java types can be passed as a generic type of a `Tensor`, like it is enforced here:
```
 public static <T extends TType> Tensor<T> create(Object obj, DataType<T> dtype) {...}
```
It very soft compile-time type safety but better than nothing, I think.

tensorflow-utils/nio-utils/src/main/java/org/tensorflow/nio/buffer/DataBuffers.java
karllessard: Thanks, I'll revise this

karllessard:
> What's the copyright on the 1500x916.jpg file? I didn't see it mentioned anywhere, but that needs to be clear to put anything in a repository from Google.

Oops! I had a reminder to change this one for one of my own picture but clearly I forgot, thanks for pointing that out!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
Craigacp: In a future version of Java we will be able to seal this interface so it only has the implementations that we specify, which will make it more useful & reliable (as we'll be able to switch on the subclasses with exhaustiveness checking, and users won't be able to do their own implementations that our code can't work with).

saudet:
> > Does that mean we can now rename `getShort()`, `getInt()`, `setByte()`, etc to just `get()` and `set()`? I think that would be a lot nicer.
> 
> On the `DataBuffer`, they could but on the `NdArray`, they are already being used for getting/setting subelements of an array.
> 
> I was inspired by this: if you think that `get()/set()` are replaceable by brackets (like Kotlin does), for a given 2D matrix, it might be natural to retrieve its first row simply with `matrix[0]` instead of something like `matrix.getElement(0)`. On the other hand, if `get()/set()` were used to return a value type, then only the right number of brackets that leads to a scalar would be allowed (`matrix[0][0]` in this case), which is a bit awkward.

Ah, so the `get()` methods basically return another array? That's interesting. That certainly makes it feel like NumPy, and that's pretty much how ND4J went on to do it too. It hasn't proven to be too confusing, so that sounds good. I think we should keep it like that then.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: I think it's just the names that are confusing me. When I read something like `tensorflow-core`, I think like this means the actual "core" without dependencies on other modules. Maybe it's just that `tensorflow-nio` should really be named something else and not "TensorFlow"? That might also help with adoption from other projects. That's probably something we should continue on the mailing list you created about that and one of the first things to clear up with guys from MXNet, DL4J, etc.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: This is going to do the same as the JNI code in that file:
```java
TF_TensorData(nativeTensor).capacity(TF_TensorByteSize(nativeTensor)).asByteBuffer()
```
Along with a couple of null checks there too if needed.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
saudet: I see, that sounds useful.

karllessard:
> Ah, so the `get()` methods basically return another array? That's interesting. That certainly makes it feel like NumPy, and that's pretty much how ND4J went on to do it too. It hasn't proven to be too confusing, so that sounds good. I think we should keep it like that then.

Ok. Like I said, we could rename those in `*DataBuffer` though but I'm tempted to leave them like that as well just to be consistent with `*NdArray`

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Exactly, that's why I renamed the artifact `tensorflow-nio-utils` to `nio-utils` (though I'm not a huge fan of that name neither so if anyone comes up with a better one...). So the `tensorflow` you've noticed there is just to mention our organization in the group and does not name the artifact itself.

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Ok I'll give it a try, thanks

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
karllessard: Question for you @Craigacp, will the sealed interface work in Java the same way as in Kotlin, meaning that will only interfaces from the same package (or the same file for Kotlin) be allowed to extend from it or will there be a different safety mechanism?

karllessard:
BTW group, I was thinking to push a new version that renames the following methods (and their siblings):

* `TInt32.scalar(10)` to `TInt32.scalarOf(10)`
* `TInt32.vector(1, 2, 3, 4)` to `TInt32.vectorOf(1, 2, 3, 4)`
* `TInt32.tensorOfShape(2, 2)` to `TInt32.ofShape(2, 2)`
* `Operand.tensorData()` to `Operand.data()`
* `shape(2, 2)` to `shapeOf(2, 2)`

Do you have any preference?

saudet:
@karllessard No preferences, but we should try to go with unwritten conventions people are used to:
https://blog.joda.org/2011/08/common-java-method-names.html

tensorflow-core/tensorflow-core-api/pom.xml
Craigacp: `ndarray-core`?  I think `nio-utils` is likely to make people think of `java.nio.*` and as the language moves away from those classes (and they are 15+ years old at this point), calling them "new" is something of a misnomer.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java
Craigacp: This is the current JEP for the sealed interfaces feature - https://openjdk.java.net/jeps/360. It requires them to be in the same package or module, and listed in the `permits` clause of the interface/class. If the subclasses are all in the same java file, then they don't have to be listed in the permits clause (but that's probably not how we'll do it).

Craigacp:
I'm fine with those renamings, they seem fairly idiomatic.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Or `ndarray-utils`? Or the other way around `utils-ndarray`? Sounds more Java-ish and in line with the name of the parent artifact

karllessard: yeah so I'm convinced as well, `nio` is out. `ndarray-utils` I like, I understand the logic but it sounds more natural than `utils-ndarray`. Or just `ndarray`? `nd-utils`? 

There are `DataBuffer` as well but they are mainly there to support Nd-arrays.

karllessard: yeah, I don't dislike simply `org.tensorflow:ndarray`, wdyt?

Craigacp: I think I prefer `ndarray` then `ndarray-utils`. I think both `ndarray-utils` and `utils-ndarray` seem a little odd given this module is actually the base classes for ndarrays.

karllessard: Ok, just for the sake of picking something, `ndarray` it is, thanks guys!

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Great, that works, bye bye JNI...

tensorflow-core/tensorflow-core-api/pom.xml
saudet: I don't think it belongs in "tensorflow" anyway, whether is goes under `tensorflow-utils` or not is a moot point IMO.

saudet: We could rename that parent package to `tensorflow-misc` and just put whatever doesn't fit in there. That would kind of make more sense to me...

karllessard: > I don't think it belongs in "tensorflow" anyway

That's a valid point. Actually, we can host that code elsewhere since it does not depend on TF but it is the other way around, like JavaCPP. Should we?

saudet: Yes, I think everything should be in JavaCPP :) But seriously, this is something that needs to be discussed with people from other projects too.

As for the matter at hand, the name of the artifact, looking back at how Spring does it at https://github.com/spring-projects/spring-framework maybe we shouldn't try to categorize all the modules. `tensorflow-core` has the Bazel build with settings for JavaCPP and everything so that can stay like that, but I think we could have `tensorflow-ndarray` along with `tensorflow-jkeras`, etc, all as main modules under `tensorflow-java`, without a kind of "dump" like `tensorflow-frameworks` or `tensorflow-utils`. That sounds nice to me. In any case, it's probably going to be temporary if we are able at some point to migrate to a version that comes from another repo and that is accepted by other organizations such as Apache MXNet and Eclipse Deeplearning4j as well.

karllessard: I agree, we will group artifacts if there is a need but right now, I cannot think of any other utility library than this one.

saudet: I think there should be a functional need to group things together. If it's just to put them in abstract categories, it will make it hard to decide which category each module belongs to, possibly without any actual benefits. Maven doesn't even put the parent name in artifact coordinates, so unless we do it explicitly like with `tensorflow-core` -> `tensorflow-core-api`, etc then the benefits are even more marginal.

karllessard: Ok, here is what I ended up to do: I preserved the `tensorflow-utils` name but now it contains the code of the library itself, including `DataBuffer` and `NdArray` APIs. 

So if in the future it happens that we have more small utility classes like these that we would like to share with the world and that are independent from TF runtime, they will end up in this library as well (kind of a Guava for ML). 

There is a lot of presumptions and questions that makes it hard to pick the right name for this library right away, so let's simply rename it in the future if we need to.

karllessard:
Just to let you know that I've just pushed a new version of the library with all your recommendations.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Sounds good, but let's name it `tensorflow-util` just to be consistent with the package name? Or inversely, let's name the package `org.tensorflow.utils`? I don't have preference for either, I just think that consistency is a good thing to have whenever it makes sense.

karllessard: I think it is fine to have them mismatch here, `tensorflow-utils` sounds more natural for the name of library (as there is more than one utility class in it) while package names are often singular by convention (e.g. `java.util.*`). 

In addition, we don't have a `org.tensorflow.core` package neither in our core artifacts.

saudet:
Yes, so what about org.tensorflow.ndarray? That would make it more
consistent in a way.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -26,6 +26,16 @@
...
+import org.tensorflow.types.TDouble;
+import org.tensorflow.types.TFloat;
+import org.tensorflow.types.TInt32;
+import org.tensorflow.types.TInt64;
+import org.tensorflow.types.TString;
+import org.tensorflow.types.TUInt8;
+import org.tensorflow.types.family.TType;
 
 /**
  * A statically typed multi-dimensional array whose elements are of a type described by T.
```
Craigacp: Can we more strongly type `Tensor` to `Tensor<T extends TType>`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -261,23 +251,33 @@
     } else {
       nremaining = data.remaining();
     }
-    Tensor<?> t = allocateForBuffer(dtype, shape, nremaining);
+    Tensor<T> t = allocateForBuffer(dtype, shape, nremaining);
     t.buffer().put(data);
     return t;
   }
 
+  public static <T extends TType> Tensor<T> allocate(DataType<T> dtype, Shape shape) {
```
Craigacp: Needs some javadoc. Looks like this makes an empty buffer, but what does that mean for types like String? Are they populated by the null reference or the empty string etc?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -287,21 +287,20 @@
   // Helper function to allocate a Tensor for the create() methods that create a Tensor from
   // a java.nio.Buffer.
   // Requires: dataType matches T
-  private static <T> Tensor<T> allocateForBuffer(DataType dataType, long[] shape, int nBuffered) {
-    final int nflattened = numElements(shape);
+  private static <T extends TType> Tensor<T> allocateForBuffer(DataType dataType, long[] dimSizes, int nBuffered) {
```
Craigacp: Can we type `DataType` with `<T>` here? And in a bunch of other places. Or alternatively `<? extends TType>`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
```diff
@@ -15,49 +15,35 @@
...
-  /**
+   * @param name readable-name for this type
    * @param value must match the corresponding TF_* value in the TensorFlow C API.
    * @param byteSize size of an element of this type, in bytes, -1 if unknown
+   * @param tensorMapper method for mapping tensor memory to elements of this type
    */
-  DataType(int value, int byteSize) {
-    this.value = value;
-    this.byteSize = byteSize;
+  public static <T> DataType<T> create(String name, int value, int byteSize, TensorMapper<T> tensorMapper) {
```
Craigacp: Stronger type? `<T extends TType>`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
```diff
@@ -15,49 +15,35 @@
...
+  public interface TensorMapper<T> {
+
+    /**
+     * Maps tensor memory to a data structure for manipulating elements of this type.
+     *
+     * @param nativeTensor pointer to the native tensor
+     * @param shape the shape of the tensor
+     * @return data structure of elements of this type
+     */
+    T apply(TF_Tensor nativeTensor, Shape shape);
```
Craigacp: This exposes the JavaCPP `TF_Tensor` type into the public API. Can we fix that?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
```diff
@@ -15,49 +15,35 @@
...
-  /** 32-bit signed integer. */
-  INT32(3, 4),
-
-  /** 8-bit unsigned integer. */
-  UINT8(4, 1),
+import org.tensorflow.internal.c_api.TF_Tensor;
+import org.tensorflow.util.Shape;
+
+/** Represents a type of elements in a {@link Tensor} */
+public final class DataType<T> {
```
Craigacp: Stronger type bound for DataType, `<T extends TType>`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
```diff
@@ -67,50 +53,39 @@ public int byteSize() {
...
+   * @return data structure of elements of this type
    */
-  public static DataType fromClass(Class<?> c) {
-    DataType dtype = typeCodes.get(c);
-    if (dtype == null) {
-      throw new IllegalArgumentException(
-          c.getName() + " objects cannot be used as elements in a TensorFlow Tensor");
-    }
-    return dtype;
+  T map(Tensor tensor) {
```
Craigacp: This takes a bare `Tensor` type, can it have a stronger bound? Either <? extends TType>  or even `<T extends TType>`.

tensorflow-utils/src/main/java/org/tensorflow/util/StaticApi.java
```diff
@@ -0,0 +1,708 @@
...
+  }
+
+  /**
+   * Creates a vector (rank 1) initialized with the given values.
+   *
+   * @param values vector values
+   * @return new vector
+   * @throws IllegalArgumentException if values is null
+   */
+  static <T> NdArray<T> vectorOfObjects(T... values) {
```
Craigacp: This is just a wrapper around references right? Is it clear that we need that kind of loose bound? A multidimensional array of references doesn't give us the locality or performance that we're looking for, but people might think it will.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/jdk/ByteJdkDataBuffer.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.nio.ByteBuffer;
+import org.tensorflow.util.buffer.ByteDataBuffer;
+import org.tensorflow.util.buffer.DataBuffer;
+import org.tensorflow.util.buffer.impl.Validator;
+
+/**
+ * A buffer of bytes using a JDK {@link ByteBuffer} for storage.
+ * <p>
+ * Since JDK buffers supports only 32-bits indexation, the size of this buffer type cannot
```
Craigacp: "32-bits indexation" -> "32-bit indexing", similarly elsewhere in the javadoc.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/DataBuffer.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.nio.ReadOnlyBufferException;
+
+/**
+ * A container of data of a specific type.
+ *
+ * <p>Instances of {@code DataBuffer} map native or heap memory segments to a linear view that
+ * supports:
+ * <ul>
+ *  <li>64-bits indexation, allowing to work with a single buffer with a size up to
```
Craigacp: It's probably limited to 2^48 as there are only 48 address bits in current x86_64 hardware. The top address bits tend to be used for access control or pointer colouring, and so we can't usually get into them.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/DataBuffers.java
```diff
@@ -0,0 +1,509 @@
...
+  }
+
+  /**
+   * Wraps an array of bytes into a data buffer.
+   *
+   * @param array array to wrap
+   * @param readOnly true if the buffer created must be read-only
+   * @return a new buffer
+   */
+  public static ByteDataBuffer wrap(byte[] array, boolean readOnly) {
```
Craigacp: These methods should note if they copy or present a view, as views can be modified through other access points (e.g. the input array here). That's likely to cause odd bugs, so we need to document it so we can point at the documentation when people run into the bugs. It might also be worth adding a `boolean makeCopy` argument to all these methods to make it explicit.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/FloatDataBuffer.java
```diff
@@ -0,0 +1,153 @@
...
+public interface FloatDataBuffer extends DataBuffer<Float> {
+
+  /**
+   * Reads the float at the given index.
+   *
+   * @param index the index from which the float will be read
+   * @return the float at the given index
+   * @throws IndexOutOfBoundsException if index is negative or not smaller than the buffer size
+   */
+  float getFloat(long index);
```
Craigacp: I like these methods, as it allows us to make an FP16 one that can have an FP16 view (either as as short or a value type) and a float view. So I don't think they should be renamed to just `get/set`.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArraySequence.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+/**
+ * Iterates through a sequence of elements of an N-dimensional array.
+ *
+ * @param <T> data type of the array being iterated
+ */
+public interface NdArraySequence<T extends NdArray<?>> {
```
Craigacp: Can this be an iterator? Or implement Iterable?

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArraySequence.java
```diff
@@ -0,0 +1,29 @@
...
+   *
+   * @param consumer method to invoke for each elements
+   */
+  void forEach(Consumer<T> consumer);
+
+  /**
+   * Visit each elements of this iteration and their respective coordinates.
+   *
+   * <p><i>Important: the consumer method should not keep a reference to the coordinates
+   * as they might are mutable and reuse during the iteration to improve performance.</i>
```
Craigacp: "might be mutable and reused"

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
```diff
@@ -0,0 +1,87 @@
+package org.tensorflow.util.buffer.impl.raw;
+
+import java.lang.reflect.Field;
+import sun.misc.Unsafe;
+
+/**
+ * Holds a valid reference to an instance of sun.misc.Unsafe.
+ */
+public final class UnsafeReference {
```
Craigacp: Does this class need to be public?

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
```diff
@@ -0,0 +1,87 @@
...
+   * Creates a reference to {@code unsafe}
+   *
+   * The reference can be created only if the instance exposes the methods required by
+   * this library, otherwise an exception is thrown
+   *
+   * @param unsafe instance of {@code Unsafe}
+   * @return a valid reference to {@code unsafe}
+   * @throws IllegalArgumentException if {@code unsafe} instance does not meet requirements for this library
+   */
+  public static UnsafeReference from(Object unsafe) {
```
Craigacp: Does this method need to be public? And can it take a stronger type than Object?

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArray.java
```diff
@@ -0,0 +1,356 @@
...
+   * independently, without affecting each other.
+   *
+   * @param dst the destination array
+   * @param offset the index of the first element to write in the destination array
+   * @return this array
+   * @throws java.nio.BufferOverflowException if the destination array cannot hold the content of
+   * this array
+   * @throws IllegalArgumentException if offset is greater than dst length or is negative
+   */
+  NdArray<T> read(T[] dst, int offset);
```
Craigacp: Shouldn't `offset` be a long? And similarly for write. I think this method is rather confusing as to know the offset you need to know what the backing store is and if it's row or column major etc. I also couldn't easily find a way of calculating the offset.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/index/From.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public long mapCoordinate(long coordinate, Dimension dim) {
+    return start + coordinate;
+  }
+
+  From(long start) {
+    this.start = start;
+  }
+
+  private long start;
```
Craigacp: These fields should be final.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/index/Index.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * An index used for slicing a view out of an N-dimensional array.
+ *
+ * <p>A slice, i.e. a reduced view, of an N-dimensional array is obtain by calling
+ * {@link NdArray#slice(Index...)}, given a list of indices
+ * that select which elements on a given dimension should be included/excluded
+ * from that view.
+ */
+public interface Index {
```
Craigacp: Is there a way to make this interface functional or more constant? I'm a little worried that the JVM won't be able to optimise all these small objects properly (until they become value types). Maybe we could make them `ValueBasedClass`es? They provide conceptually similar functionality to `VarHandle` from Java 9, which Paul wrote, so maybe his input would be useful on these classes.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dimension/UnknownDimension.java
```diff
@@ -0,0 +1,56 @@
...
+    return "Unknown";
+  }
+
+  @Override public int hashCode() {
+    return System.identityHashCode(this); // All unknown dimensions are distinct
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return false; // All unknown dimensions are distinct
```
Craigacp: Shouldn't they have reference equality? The hashcode does.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dimension/AbstractDimension.java
```diff
@@ -0,0 +1,43 @@
...
+package org.tensorflow.util.ndarray.impl.dimension;
+
+abstract class AbstractDimension implements Dimension {
+
+  /**
+   * Dimensions are known to be equal if they have the same number of elements
+   */
+  @Override
+  public int hashCode() {
+    return (int) numElements();
```
Craigacp: If we care about hashcode for this we should apply some mixing function first. Integer's hashcode is unfortunately terrible but I think it's dictated by the spec.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dense/DoubleDenseNdArray.java
```diff
@@ -0,0 +1,104 @@
...
+    implements DoubleNdArray {
+
+  public static DoubleNdArray create(DoubleDataBuffer buffer, Shape shape) {
+    Validator.denseShape(buffer, shape);
+    return new DoubleDenseNdArray(buffer, shape);
+  }
+
+  @Override
+  public double getDouble(long... indices) {
+    return buffer().getDouble(positionOf(indices, true));
```
Craigacp: Some of these methods go through `buffer()`, some access the `buffer` field directly. It should pick one way, or document that they behave differently (as `DoubleDenseNdArray` isn't final, so could be subclassed).

karllessard:
> Yes, so what about org.tensorflow.ndarray? That would make it more consistent in a way.

@saudet , I still think that the actual naming is better. Right now, the package tree looks like this:
```
- org.tensorflow.util
  |_ ndarray
  |_ buffer
```
It make sense because the buffer API can actually be used with no ndarrays. Also, I have moved the `Shape` class directly under `org.tensorflow.util` since it it is also found in other contexts than ndarrays (unmapped tensors, ops operands, etc.)

Also, just to clarify my previous post, I brought the counterexample of `tensorflow-core-api` to point out that we already don't enforce too much strictness between the name of the artifact and its packages. But I do personally prefer that each artifact have at least their discriminative base package, hence `org.tensorflow.util` in this case.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: That is what I tried at first but to discover than this enforcement became quickly viral and that it needs to be everywhere in our methods, classes, wrappers, frameworks and, ultimately,  user code. 

But since `TType` does not expose any method and since we enforce it at the beginning of the chain (i.e. at the creation of the `Tensor`), I realized that there is no real gain carrying it afterward. We just want to make sure that tensors are created using one of our supported data types.

I think our users will appreciate that we relax a bit this rule.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: I have to admit that the need to add a proper Javadoc to all endpoints affected by this PR is on my TODO list. I was waiting for some people to try it our first but then the SIG decided we should better merge it and fix it after. Are you still OK with this?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: It should definitely extend `<T>`, those are left overs from previous implementation, I'm surprised my IDE is not giving me any warning about it... I'll fix it, thanks!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
karllessard: Mmh, I don't remember if that one was as hemorrhagic as with `Tensor`, I'll give it a try and let you know.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
karllessard: It is not "really" exposed. This interface is only used internally but it has to be public unfortunately because it is shared between packages. 

What I can do is to move it under `org.tensorflow.internal.buffer` so at least it is known to be internal. And since it does not return a `TF_Tensor` but accepts one, there is no major "leak" neither.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
karllessard: why I get no warning about this stuff... :-/ I'll double-check everywhere, thanks)

tensorflow-utils/src/main/java/org/tensorflow/util/StaticApi.java
karllessard: This vector won't be performant, true, but a tensor of strings is a legitimate multidimensional array though and this endpoint is mainly there to support these tensors. 

For example, since you can do:
```
TInt32.ofShape(2, 2)
    .set(vectorOf(10, 20), 0)
    .set(vectorOf(30, 40), 1);
```
ideally you should also be able to do:
```
TString.ofShape(2, 2)
    .set(vectorOf("a", "b"), 0)
    .set(vectorOf("c", "d"), 1);
```
But there is another problem: Because of how a tensor of string is structured internally, its data must be known at allocation time and has to be read-only after (as opposed to other type of tensors). Thus, we need to build up the N-dimensional array in memory first and then pass it at tensor creation, like this:
```
TString.copyOf(ndArrayOf(String.class, shapeOf(2, 2))
    .set(vectorOf("a", "b"), 0)
    .set(vectorOf("c", "d"), 1)
);
```
So in the end, as you can see, we need those non-performant reference multidimensional arrays. 

I was  forced to add the `Objects` suffix to the name of some methods to avoid clashes with those of a primitive type. Are you suggesting that maybe we make these names even more explicit about the nature of the created data structure? Like `ndArrayOfReferences`, `vectorOfReferences`, etc.?

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/DataBuffers.java
karllessard: Yes, Javadoc needs to be improved here as well. I'll check for the `makeCopy` as well, sounds like a good addition.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArraySequence.java
karllessard: I think that *now*, it should (as I changed its implementation many times...). I'll update it if that's the case.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
karllessard: Yes, because `tensorflow-core-api` creates an instance of it. But there is nothing that can be done with it other than passing it back to a `*RawDataBuffer`. Its purpose is to guarantee that the reference holder was able to provide a valid instance of `sun.misc.Unsafe`.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
karllessard: yes, needs to be public, see my previous comment. The reason of accepting any `Object` here is to avoid uncaught `ClassNotFoundException` on a JDK that does not have `sun.misc.Unsafe`... but I just realized that I forgot that exception type in the `catch` clause below, I'll fix it :)

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArray.java
karllessard: That `offset` is the offset in the array passed in argument (therefore 32-bits only). That is to mimics methods found in `java.nio.Buffer`, I don't know how useful they are but it was not hard to support. I admit though that the double usage of `offset` might be confusing... maybe I can rename it to `dstOffset`? (and `srcOffset` for write)?

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/index/Index.java
karllessard: Ok, let's do this as a follow up.

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dimension/UnknownDimension.java
karllessard: This class is a remnant of a previous implementation and is not used anymore, I'll remove it

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dimension/AbstractDimension.java
karllessard: Totally agree, I'll fix it

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/impl/dense/DoubleDenseNdArray.java
karllessard: `buffer()` is only there to let the superclass access the `buffer` stored in the subclass. Again, this has changed so many times... I'll sanitize it, thanks!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
Craigacp: The advantage of doing it is that the erased type bound of `Tensor` becomes `Tensor<TType>` which makes it tricker for people to skirt the generics to do things that they shouldn't be doing. I agree it seems like it's everywhere in the codebase, and it does ripple a little into user code, but there is some benefit to it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
Craigacp: Yeah, I'm fine with you merging it without the extra javadoc. We need to make sure it's there before there is a release with this code in though, so maybe file an issue for it?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
Craigacp: We could turn on compiling with warnings, then maven would spray them everywhere.

tensorflow-utils/src/main/java/org/tensorflow/util/StaticApi.java
Craigacp: Ok. I'd go with `somethingOfReferences` to make it clear, no preference between ndArray or vector.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
Craigacp: Could we have a static factory somewhere that deals with that, so it could remain private?

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArray.java
Craigacp: Ah, ok. So it starts reading/writing from the start of the `NdArray` and fills it in whatever order the backing store uses? I still think that we should document the orientation of the backing store (i.e. row or column major) and if we want it to be adjustable (e.g. to allow passing matrices to fortran code) then it should be an enum stored in an instance field and available on all the `NdArray` subtypes.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
Craigacp: It's times like this that I understand why C++ has friend classes.

karllessard:
> Other general points:
> 
> * Should a scalar ndarray have a tag interface?

@Craigacp : might be hard to enforce in all cases. For example, for a 2x2 matrix, both `matrix.get(0)` and `matrix.get(0, 0)` returns an `NdArray` but only the second will be a scalar. Do you see an actual benefit for having it?

> * Might want to use the code gen pattern from Java Vector API (https://github.com/openjdk/panama/blob/vectorIntrinsics/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/gen-src.sh) to emit the buffers/ndarrays of a particular kind (raw, jdk, etc) from a single source file. Reduces the amount of things that can go wrong in a refactor.

Sounds interesting, I'll take a look, thanks

> * Can we get remove UnsafeReference from the public API? I'm uncomfortable with public methods that need it as an argument.

`UnsafeReference` is not `Unsafe` but just a validator/wrapper to it. We are not exposing a free access to `Unsafe` as it needs to be provided by the caller itself. In addition, those methods are not exposed by the "public API", composed of the public interfaces and their static helpers, but found in the internal `*.impl` packages, for what it's worth.

> * Is there a relationship between buffer.Validator and ndarray.Validator? They feel like they have a lot of duplicated code.

I guess the difference can be subtle, I will check

> * Needs more javadoc (in the impl classes mainly). I understand how irritating it is to say that too, trust me, but bringing in new people to work in this codebase will be easier with it.

Yes, like I said in a previous comment, I postponed that task for later as this work is still known as "in progress" but I agree it should become a priority.

Craigacp:
> > Other general points:
> > 
> > * Should a scalar ndarray have a tag interface?
> 
> @Craigacp : might be hard to enforce in all cases. For example, for a 2x2 matrix, both `matrix.get(0)` and `matrix.get(0, 0)` returns an `NdArray` but only the second will be a scalar. Do you see an actual benefit for having it?
Ah, yeah I'd forgotten that the accessor took a varargs so you couldn't have a sharper return type. Not that the type system is quite powerful enough to do the best thing there anyway. As scalars don't have dimension then I feel like that should be represented in the type system so you can specialise when being passed a scalar (i.e. immediately unboxing it), but it's hard to get methods that emit the right types given the way the library is setup. 
> > * Can we get remove UnsafeReference from the public API? I'm uncomfortable with public methods that need it as an argument.
> 
> `UnsafeReference` is not `Unsafe` but just a validator/wrapper to it. We are not exposing a free access to `Unsafe` as it needs to be provided by the caller itself. In addition, those methods are not exposed by the "public API", composed of the public interfaces and their static helpers, but found in the internal `*.impl` packages, for what it's worth.
I understand, but anything that's public is part of the public API as far as users are concerned. I've written impl classes with big warnings on them, but unless you can lock them off with a module system or another access control people will still see them in the javadoc and use them for things you weren't expecting.

saudet:
> > Yes, so what about org.tensorflow.ndarray? That would make it more consistent in a way.
> 
> @saudet , I still think that the actual naming is better. Right now, the package tree looks like this:
> 
> ```
> - org.tensorflow.util
>   |_ ndarray
>   |_ buffer
> ```
> 
> It make sense because the buffer API can actually be used with no ndarrays. Also, I have moved the `Shape` class directly under `org.tensorflow.util` since it it is also found in other contexts than ndarrays (unmapped tensors, ops operands, etc.)
> 
> Also, just to clarify my previous post, I brought the counterexample of `tensorflow-core-api` to point out that we already don't enforce too much strictness between the name of the artifact and its packages. But I do personally prefer that each artifact have at least their discriminative base package, hence `org.tensorflow.util` in this case.

I think the extra package is fine (it allows to split the module into `tensorflow-util-ndarray`, `tensorflow-util-buffer`, etc more easily if it makes sense, which we could do right away FWIW), it's just the naming inconsistency that bothers me. If we search for "util" and "utils" with https://search.maven.org, we actually get more hits for "util"!  Although we get a lot more hits for "tools" than "tool", so to follow conventions, maybe that's the term you would be looking for? How does `org.tensorflow.tools` sound to you as package name?

karllessard:
> How does `org.tensorflow.tools` sound to you as package name?

Yes, I like `tools` as an artifact and package name, better than `utils/util`. I'll make the change.

*...and we would like to thank all our supporters, friends and family* :)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: good idea, I'll create an issue right after merging

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: I'm aware of the benefits of enforcing these safety checks, I've always been a defender of it, but I think we can make an exception here. I don't care that much of the verbosity in our code base but I'm concerned about the user code. 

I guess writing down a quick example may help us to take the right decision. Meanwhile, I'm still not sure to understand what wrong can happen if we leave the generic type parameter unbounded after creating the tensor, other than not being a good practice. Do you have a specific scenario in mind?

karllessard:
> but it's hard to get methods that emit the right types given the way the library is setup.

It would be possible to add a different method (something like `ndarray.scalar(long...)`) to return a `Scalar` type and also make sure that the number of indices provides is enough to reach one. This won't break the contract with the actual API. I suggest to do this though in a second phase, if we identify a real need for having it.

> I understand, but anything that's public is part of the public API as far as users are concerned. I've written impl classes with big warnings on them, but unless you can lock them off with a module system or another access control people will still see them in the javadoc and use them for things you weren't expecting.

I can shuffle things a bit to mark these methods as `protected` instead, would that work for you? With this and the actual protection that forces the caller to do the dirty reflective job to retrieve the instance of `Unsafe`, I think we can consider that our library has no major security breach, what do you think?

Craigacp:
> > but it's hard to get methods that emit the right types given the way the library is setup.
> 
> It would be possible to add a different method (something like `ndarray.scalar(long...)`) to return a `Scalar` type and also make sure that the number of indices provides is enough to reach one. This won't break the contract with the actual API. I suggest to do this though in a second phase, if we identify a real need for having it.

Sure

> 
> > I understand, but anything that's public is part of the public API as far as users are concerned. I've written impl classes with big warnings on them, but unless you can lock them off with a module system or another access control people will still see them in the javadoc and use them for things you weren't expecting.
> 
> I can shuffle things a bit to mark these methods as `protected` instead, would that work for you? With this and the actual protection that forces the caller to do the dirty reflective job to retrieve the instance of `Unsafe`, I think we can consider that our library has no major security breach, what do you think?

Protected would work. I don’t consider this a security thing, more an api cleanliness one. We don’t want people to start incorporating UnsafeReference as an field or argument for their classes as then we can’t remove it without it being a breaking change, and it’s really an internal implementation detail.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
karllessard: You just made me realize that there were a ton of lint checks applied in the previous Bazel build that we forgot to migrate to Maven, we need to restore them... I'll create a separate issue about this.

saudet:
> Protected would work. I don’t consider this a security thing, more an api cleanliness one. We don’t want people to start incorporating UnsafeReference as an field or argument for their classes as then we can’t remove it without it being a breaking change, and it’s really an internal implementation detail.

We need *something* to let users implement those interfaces for libraries other than TensorFlow. It's either that, JavaCPP's `Pointer` class, or a couple of `long` values for pointer addresses, which are just as unsafe. Which one do you prefer that we "leak"? I think the latter is the least worst option, but Karl doesn't like it.

Craigacp:
> > Protected would work. I don’t consider this a security thing, more an api cleanliness one. We don’t want people to start incorporating UnsafeReference as an field or argument for their classes as then we can’t remove it without it being a breaking change, and it’s really an internal implementation detail.
> 
> We need _something_ to let users implement those interfaces for libraries other than TensorFlow. It's either that, JavaCPP's `Pointer` class, or a couple of `long` values for pointer addresses, which are just as unsafe. Which one do you prefer that we "leak"? I think the latter is the least worst option, but Karl doesn't like it.

If you're in a subclassing relationship then you could see it if it's protected. As you can't have protected interface members, then it might need to become an abstract class (which might prove tricky), but at that point you could hide it effectively in the Raw classes.

karllessard:
> but at that point you could hide it effectively in the Raw classes.

I'm wondering, since I accept the `Unsafe` instance as an `Object` first (which I validate then), maybe having a public endpoint accepting the `Object` could be enough, i.e. something like `RawMemoryMapper.create(Object)` where the object passed must be an instance of `sun.misc.Unsafe`. Then the `RawMemoryMapper` will have endpoints to map memory segments from their address. So there won't be any `UnsafeReference` anymore, as the `Unsafe` instance would be kept in the `RawMemoryMapper`.

saudet:
> > but at that point you could hide it effectively in the Raw classes.
> 
> I'm wondering, since I accept the `Unsafe` instance as an `Object` first (which I validate then), maybe having a public endpoint accepting the `Object` could be enough, i.e. something like `RawMemoryMapper.create(Object)` where the object passed must be an instance of `sun.misc.Unsafe`. Then the `RawMemoryMapper` will have endpoints to map memory segments from their address. So there won't be any `UnsafeReference` anymore, as the `Unsafe` instance would be kept in the `RawMemoryMapper`.

Actually, you're already passing the pointer address to your objects there, so I don't understand why we need to expose Unsafe at all...? If it's available, we use it, if it's not available, we just use something else like ByteBuffer or whatever OpenJDK comes up with eventually. Users don't need to know what we're doing to access their data, do they? We could expose an Enum or something to give them a choice of which implementation gets used, and I think that's enough. What other use cases are you thinking about?

saudet:
> > > but at that point you could hide it effectively in the Raw classes.
> > 
> > 
> > I'm wondering, since I accept the `Unsafe` instance as an `Object` first (which I validate then), maybe having a public endpoint accepting the `Object` could be enough, i.e. something like `RawMemoryMapper.create(Object)` where the object passed must be an instance of `sun.misc.Unsafe`. Then the `RawMemoryMapper` will have endpoints to map memory segments from their address. So there won't be any `UnsafeReference` anymore, as the `Unsafe` instance would be kept in the `RawMemoryMapper`.
> 
> Actually, you're already passing the pointer address to your objects there, so I don't understand why we need to expose Unsafe at all...? If it's available, we use it, if it's not available, we just use something else like ByteBuffer or whatever OpenJDK comes up with eventually. Users don't need to know what we're doing to access their data, do they? We could expose an Enum or something to give them a choice of which implementation gets used, and I think that's enough. What other use cases are you thinking about?

Ah, I remember what you said. You're basically considering the `Unsafe` object like we consider `ByteBuffer` to be "safe", so we'd be adding a new backend for others like `MemorySegment` and what not. Fair enough. It's more work for people to implement, so I suspect not everyone is going to be alright with that, but we can always add an "unsafe" backend at some point if this library becomes popular and everyone gets tired of copy/pasting the same thing around.

Craigacp:
Well it's only the DataBuffer classes that need to be implemented separately for things like `MemorySegment`, the `NdArray` classes on top of them should be agnostic. So it's not that much of an implementation burden right?

The `MemorySegment` API was merged into 14's tree today before the feature freeze, so it should be in the 14 release in March.

karllessard:
The idea of passing the `Unsafe` object in parameter is also to avoid having a public endpoint which is simply `DataBuffer mapMe(long address, long length)`. While it is tempting on the API and usability level, it is completely unsafe since it allows to pass any arbitrary number as an address, which can screw up the whole JVM. The problem though is that we do need this endpoint for mapping tensor buffers from `tensorflow-core-api`. So by forcing the caller pass `Unsafe` himself at the initial memory mapping, we prevent that users "just wandering around" starts to play with this functionality too easily, while still allowing entitled libraries to use it by adding the required piping, which is still minimal.

We all know that there is no perfect solution here neither. I talked with Maurizio from Panama and on their side, they focus to enforce the safety of native memory mapping from the `MemoryAddress` itself. That will be interesting when it will come. We could try to do it the same way with JavaCPP pointers but that will require us to create another artifact like `tensorflow-tools-javacpp`. Or tensor memory could also be mapped directly in `tensorflow-core-api`, as it was initially. This is the safest way to achieve this as we can guarantee that the address being mapped is one of a tensor, but then we lose completely reusability and we end up reimplementing all raw buffers in another libraries. Like @Craigacp pointed out, it might not be such a big deal, but it is not necessarily what we want neither.

Now I'm experimenting something new: having a non-final `RawDataBufferFactory`, located in the raw buffers implementation package, that exposes public endpoints for creating raw (unsafe) buffers from arrays but keep protected those for mapping native memory. Then a library can simply extend from this factory and expose/use the "hidden" endpoints in the safest way for its specific case. Since we won't have public `mapMe(address, length)` endpoints anymore, it is now OK to retrieve the `Unsafe` instance directly from this factory and relieve other libraries from this responsibility. Of course, anyone can again simply extend from this class and start mapping native memory with arbitrary addresses but by locating the factory under a `impl` package and by hidding the native memory mapping endpoint from the JavaDoc since they are protected, we might obfuscate it enough for wandering users.

saudet:
That sounds really nice, but you guys have to realize that all this running around in circles prevents end users from actually using those tools for any libraries not explicitly supported, including all those distributed with the JavaCPP Presets:
https://github.com/bytedeco/javacpp-presets/

One of the reasons why NumPy works so well is that it facilitates data sharing between libraries, but this implementation of ndarray here doesn't...If I'm missing something though, please enlighten me.

karllessard:
> That sounds really nice, but you guys have to realize that all this running around in circles prevents end users from actually using those tools for any libraries not explicitly supported, including all those distributed with the JavaCPP Presets:
> https://github.com/bytedeco/javacpp-presets/
> 
> One of the reasons why NumPy works so well is that it facilitates data sharing between libraries, but this implementation of ndarray here doesn't...If I'm missing something though, please enlighten me.

It does not prevent any of this, libraries just need to extend from this factory class and make use of the `mapNative` endpoints the way they want. Here is how TF is doing it:
```java
class TensorRawDataBufferFactory extends RawDataBufferFactory {

  static ByteDataBuffer mapTensorToBytes(Pointer tensorMemory) {
    return mapNativeBytes(tensorMemory.address(), tensorMemory.capacity(), false);
  }

  static IntDataBuffer mapTensorToInts(Pointer tensorMemory) {
    return mapNativeInts(tensorMemory.address(), tensorMemory.capacity(), false);
  }

  static LongDataBuffer mapTensorToLongs(Pointer tensorMemory) {
    return mapNativeLongs(tensorMemory.address(), tensorMemory.capacity(), false);
  }

  static FloatDataBuffer mapTensorToFloats(Pointer tensorMemory) {
    return mapNativeFloats(tensorMemory.address(), tensorMemory.capacity(), false);
  }

  static DoubleDataBuffer mapTensorToDoubles(Pointer tensorMemory) {
    return mapNativeDoubles(tensorMemory.address(), tensorMemory.capacity(), false);
  }

  static StringTensorBuffer mapTensorToStrings(Pointer tensorMemory, long numElements) {
    long offsetByteSize = numElements * Long.BYTES;
    LongDataBuffer offsets = mapNativeLongs(tensorMemory.address(), numElements, false);
    ByteDataBuffer data = mapNativeBytes(
        tensorMemory.address() + offsetByteSize,
        tensorMemory.capacity() - offsetByteSize,
        false);
    return new StringTensorBuffer(offsets, data);
  }
}
```

Craigacp:
> That sounds really nice, but you guys have to realize that all this running around in circles prevents end users from actually using those tools for any libraries not explicitly supported, including all those distributed with the JavaCPP Presets:
> https://github.com/bytedeco/javacpp-presets/

The aim is to make a Java type for ndarrays, not to wrap every C library's version of an ndarray into the same interface. Supporting every C library's view of memory is likely to cause issues (e.g. at the moment it doesn't support different storage orders).

> 
> One of the reasons why NumPy works so well is that it facilitates data sharing between libraries, but this implementation of ndarray here doesn't...If I'm missing something though, please enlighten me.

All the Python libraries use NumPy arrays, or implement the same interface (via duck typing). So the libraries are built on numpy and opted into it's representation. This API will also require opting into it's representation by implementing it's interfaces if you want to store memory in a different way to how the current implementation does (which should at most be the buffer interface) so I'm not sure what the difference is?

karllessard:
> All the Python libraries use NumPy arrays, or implement the same interface (via duck typing). So the libraries are built on numpy and opted into it's representation. This API will also require opting into it's representation by implementing it's interfaces if you want to store memory in a different way to how the current implementation does (which should at most be the buffer interface) so I'm not sure what the difference is?

Ok, I might have misunderstood the original question then. FYI, it is also possible to change the layout of your data in memory just by providing a [`DataAdapter`](https://github.com/karllessard/tensorflow-java/blob/tensor-nio/tensorflow-utils/src/main/java/org/tensorflow/util/buffer/adapter/DataAdapter.java), without the need of reimplementing the whole `DataBuffer` interface.

Craigacp:
> > All the Python libraries use NumPy arrays, or implement the same interface (via duck typing). So the libraries are built on numpy and opted into it's representation. This API will also require opting into it's representation by implementing it's interfaces if you want to store memory in a different way to how the current implementation does (which should at most be the buffer interface) so I'm not sure what the difference is?
> 
> Ok, I might have misunderstood the original question then. FYI, it is also possible to change the layout of your data in memory just by providing a [`DataAdapter`](https://github.com/karllessard/tensorflow-java/blob/tensor-nio/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/adapter/DataAdapter.java), without the need of reimplementing the whole `DataBuffer` interface (and by the way, I should probably rename that interface to `DataLayout`).

But for things like GPU memory vs CPU memory (where you need to apply a native hook to get the pointer into the right address space) it's probably best that they implement a new `DataBuffer`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
karllessard: Here: https://github.com/tensorflow/java/issues/7

tensorflow-utils/src/main/java/org/tensorflow/util/StaticApi.java
karllessard: Ok, I did not liked having `*References` or `*Refs` suffix in the name of our methods, sounds a bit awkward, so I replaced this by `*Objects` and the upcoming documentation will be self-explanatory to explain the layout of those buffers and their impact on performances.

tensorflow-utils/src/main/java/org/tensorflow/util/buffer/impl/raw/UnsafeReference.java
karllessard: That is what I ended up doing in the end, and I'll probably do the same for all other type of buffers

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArray.java
karllessard: You mean the enum should be in the `NdArray` to tell what is the orientation of the data in its buffer or it should be provided by the caller to ask `NdArray` to read/write its data to the src/dest array in a specific order?

Craigacp: I think both, as if they differ then the call will be very slow, so users need to be able to see and control the orientation. For example, a row wise view is important when you are doing SGD, but a column wise view is useful for doing feature selection, coordinate ascent (e.g. in an ElasticNet) or building a decision tree.

saudet:
> > That sounds really nice, but you guys have to realize that all this running around in circles prevents end users from actually using those tools for any libraries not explicitly supported, including all those distributed with the JavaCPP Presets:
> > https://github.com/bytedeco/javacpp-presets/
> 
> The aim is to make a Java type for ndarrays, not to wrap every C library's version of an ndarray into the same interface. Supporting every C library's view of memory is likely to cause issues (e.g. at the moment it doesn't support different storage orders).
> 
> > One of the reasons why NumPy works so well is that it facilitates data sharing between libraries, but this implementation of ndarray here doesn't...If I'm missing something though, please enlighten me.
> 
> All the Python libraries use NumPy arrays, or implement the same interface (via duck typing). So the libraries are built on numpy and opted into it's representation. This API will also require opting into it's representation by implementing it's interfaces if you want to store memory in a different way to how the current implementation does (which should at most be the buffer interface) so I'm not sure what the difference is?

No, that's not true. It's very easy to create a NumPy ndarray on top of any buffer:
https://jakevdp.github.io/blog/2014/05/05/introduction-to-the-python-buffer-protocol/
I just wanted to point out the inconsistency in your goals, not criticizing. You won't be able to eat your cake and have it too!

Craigacp:
> > > That sounds really nice, but you guys have to realize that all this running around in circles prevents end users from actually using those tools for any libraries not explicitly supported, including all those distributed with the JavaCPP Presets:
> > > https://github.com/bytedeco/javacpp-presets/
> > 
> > 
> > The aim is to make a Java type for ndarrays, not to wrap every C library's version of an ndarray into the same interface. Supporting every C library's view of memory is likely to cause issues (e.g. at the moment it doesn't support different storage orders).
> > > One of the reasons why NumPy works so well is that it facilitates data sharing between libraries, but this implementation of ndarray here doesn't...If I'm missing something though, please enlighten me.
> > 
> > 
> > All the Python libraries use NumPy arrays, or implement the same interface (via duck typing). So the libraries are built on numpy and opted into it's representation. This API will also require opting into it's representation by implementing it's interfaces if you want to store memory in a different way to how the current implementation does (which should at most be the buffer interface) so I'm not sure what the difference is?
> 
> No, that's not true. It's very easy to create a NumPy ndarray on top of any buffer:
> https://jakevdp.github.io/blog/2014/05/05/introduction-to-the-python-buffer-protocol/
> I just wanted to point out the inconsistency in your goals, not criticizing. You won't be able to eat your cake and have it too!

But the equivalent of the buffer protocol is what this PR provides? Java doesn't have a low level buffer which has multidimensional indexing, that's why Karl wrote one. It's as much effort to implement the Python buffer as it is to implement this one so I'm not sure how it's easier in Python (beside the fact that everyone has already written the code in Python, and we're just starting to write it here).

tensorflow-utils/src/main/java/org/tensorflow/util/ndarray/NdArray.java
karllessard: I think it is a good idea but changing the orientation of an array may have a significant impact on the actual solution. What can be done meanwhile is simply to permute dimensions of your tensor using TF before performing the copy (e.g. transposing a 2d matrix).

Craigacp: I agree in TF I have other ways of doing it, but not within the ndarray interface itself, and it's a property of the ndarray not just an ndarray as applied to TF. I think it needs to allow the user some control over the layout, but it doesn't have to be in this initial PR.

karllessard:
FYI, I just pushed a new version with all requested changes except the bounded types in `Tensor` and `DataType` generics as suggested by @Craigacp. I want to write down a few examples first to compare the complexity/verbosity impacts on the user code. But everything that we agreed else should be in there.

For your concerns @saudet , I'm still not sure what TF tools cannot handle that you are proposing; with the buffer and data layout interfaces, you can customize pretty much the backing store of an ndarray the way you like. But again it might be my misunderstanding, so I suggest that once the code is merged, we can try to map some buffers of the C libraries you were referring to and then it might become more clear if something is missing and what it is.

karllessard:
> FYI, I just pushed a new version with all requested changes except the bounded types in `Tensor` and `DataType` generics as suggested by @Craigacp. I want to write down a few examples first to compare the complexity/verbosity impacts on the user code

Ok, I've bound types in generics to `TType` finally, the verbosity impacts are less that I remember and by doing it, I've detected some illegal casts that were remaining in the unit tests, which self-proved its usefulness.

So if everyone agrees on those changes, I'll merge the branch by the end of this week, thanks

karllessard:
PR has been merged based on other members reviews.

PR #6, opened 2019-12-10
Add extensions to build separate artifacts with MKL and/or CUDA enabled

### LINKED ISSUE #4: Current default build is MKL
This is painful for machines which don't have MKL installed. The MKL version bound against is the one using Intel's OpenMP, which is even less likely to be available on a machine (plus it's MKL-ML, not regular MKL nor MKL DNN - now DNNL).

We should make the build system a little more configurable, or at least set the default build to be something that works everywhere (even if the performance is worse).

I keep running into this when testing things on different machines here, none of which have MKL, and I have to get approvals for new inbound libraries.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux
- TensorFlow installed from (source or binary): source
- TensorFlow version: 2.0

saudet:
Fixes #4

Usage should be clear from the modifications to the README.md. I've tested this on Linux, all builds pass and it runs, but some of the unit tests fail with CUDA. I think it just has to do with limited memory on my GPU though.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Tensor.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Tensor.java
@@ -32,14 +32,7 @@ public abstract class AbstractTF_Tensor extends Pointer {
     /** TensorFlow crashes if we don't pass it a deallocator, so... */
     protected static Deallocator_Pointer_long_Pointer dummyDeallocator = new Deallocator_Pointer_long_Pointer() {
         @Override public void call(Pointer data, long len, Pointer arg) { }
-    };
-
-    static {
-        PointerScope s = PointerScope.getInnerScope();
-        if (s != null) {
-            s.detach(dummyDeallocator);
-        }
-    }
+    }.retainReference();
 
     /** A reference to prevent deallocation. */
     protected Pointer pointer;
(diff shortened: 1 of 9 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/c_api/AbstractTF_Tensor.java
```diff
@@ -32,14 +32,7 @@
...
         @Override public void call(Pointer data, long len, Pointer arg) { }
-    };
-
-    static {
-        PointerScope s = PointerScope.getInnerScope();
-        if (s != null) {
-            s.detach(dummyDeallocator);
-        }
-    }
+    }.retainReference();
```
Craigacp: This change doesn't look related to the rest of the PR?

saudet: It's needed for JavaCPP 1.5.2, which changed a bit how `PointerScope` works to accommodate for reference counting. It doesn't affect anything since it's not being used right now, and it might never be either, if that gets fixed upstream and this hack is no longer needed.

karllessard:
This seems good to me, @sjamesr do you also want to take a look? Basically it allows us to distribute TF in 4 flavors: "vanilla", "w/mkl", "w/gpu" and "w/mkl+gpu"

PR #8, opened 2019-12-20
Add initial build workflow for CI using GitHub Actions

saudet:
It took me a while to determine which versions of everything work, to research all the flags needed to work around various issues, and otherwise get all the magical incantations just right, but with these updates GitHub Actions is pretty much working on Linux, Mac, and Windows including MKL and CUDA for CI purposes. I had to upgrade to TensorFlow 2.1.0-rc2 because 2.0.0 doesn't build correctly when using Visual Studio 2019 and that's the only version available on the hosted virtual machines. The result looks something like this on my fork: https://github.com/saudet/tensorflow-java/commit/4d7340eea7bd354b8da2283e61c276360852e290/checks?check_suite_id=377172023

The only remaining issue is the small size of the page file on Windows that makes MSVC run out of heap space: https://github.community/t5/GitHub-Actions/Error-quot-The-paging-file-is-too-small-for-this-operation-to/m-p/41713#M4723

I've tried to work around that by using options like `--local_cpu_resources=1` and `--local_ram_resources=2048` for Bazel, and `/Zm10` for the C++ compiler, but it's not helping. We'll need to wait until they fix this, or if Google is going to give us beefier machines with GPUs and everything to run this on anyway, we might as well do this now. @ewilderj What's the status on that front?

In any case, after we figure out something for the credentials on Sonatype OSS, we can start uploading snapshots automatically from these builds. @sjamesr How should we proceed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,105 @@
+name: CI jobs
+on: [push, pull_request]
+
+jobs:
+  linux-x86_64:
+    runs-on: ubuntu-latest
+    container: nvidia/cuda:10.1-cudnn7-devel-centos7
+    strategy:
+      matrix:
+        ext: ["", -mkl, -gpu, -mkl-gpu]
+    steps:
+      - name: Install environment
+        run: |
+          yum -y update
+          yum -y install centos-release-scl-rh epel-release
+          yum -y install java-1.8.0-openjdk-devel devtoolset-7 rh-git218 rh-maven35 patch python36-devel python36-pip python36-six
+          echo Downloading Bazel
+          curl -L https://github.com/bazelbuild/bazel/releases/download/0.29.1/bazel-0.29.1-installer-linux-x86_64.sh -o bazel.sh --retry 10
+          bash bazel.sh
+      - name: Checkout repository
+        uses: actions/checkout@v1
+      - name: Build project
+        run: |
+          source scl_source enable devtoolset-7 rh-git218 rh-maven35 || true
+          git --version
+          gcc --version
+          mvn -version
+          bazel version
+          df -h
+          echo Executing Maven without Wagon HTTP pool that fails under Docker
+          mvn clean install -B -U -e -Dmaven.wagon.http.pool=false -Djavacpp.platform=linux-x86_64 -Djavacpp.platform.extension=${{ matrix.ext }}
+          df -h
+  macosx-x86_64:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        ext: ["", -mkl]
+    steps:
+      - name: Install environment
+        run: |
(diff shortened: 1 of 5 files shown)


sjamesr:
Hey Samuel, I can get you added to the Sonatype repository ACLs, though we
probably need to create a new snapshot repo. I assume that's a matter of
filing a ticket with the Sonatype folks.

saudet:
@sjamesr My username on Sonatype OSS is also `saudet`. If you can give me permission to deploy for `org.tensorflow` that would be great. We just need to file a ticket for requests like that here: https://issues.sonatype.org/

To have GitHub Actions upload artifacts tough, we need to add the credentials with this:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets
I can do that for my own fork, but not here. I guess I could test it on my fork with my own personal credentials, and when it gets merged here, you can add the ones you've been using until now in the settings there. Sounds good?

BTW, for OSS projects Sonatype lets us use their own repo for snapshots as well, so that should be OK for now anyway, but sure, if you have or will have another Nexus instance somewhere that would be more appropriate, we can use that easily.

sjamesr:
@saudet I just filed https://issues.sonatype.org/browse/MVNCENTRAL-5426

saudet:
> @saudet I just filed https://issues.sonatype.org/browse/MVNCENTRAL-5426

Thanks! BTW, I don't have access to that issue. It looks like OSSRH and MVNCENTRAL are 2 different "projects"...

Anyway, do you think we should merge what we have here right away, or should we first wait until we get builds on Windows working and deployment going?

saudet:
@karllessard You'll need to be more specific about what changes you're requesting. You could also make those changes directly on my fork. That's probably going to be easier.

karllessard:
@saudet Actually I don't request any changes, I just did it to prevent the PR to be updated for deployment and merged before we address this issue (which seems to be the organic evolution of your previous discussions :) )

Ok so what about we merge this one as is and make another PR for deploying the artifacts, once the credentials are set and licenses are fixed @sjamesr ?

saudet:
I see, so that would be an additional issue that needs to be resolved before we can start deploying snapshots. Since running the builds and tests is still useful by itself, instead of trying to address all these issues here, it might make more sense to merge this PR right away. Actually, we should probably follow a policy of not trying to address multiple issues like that so PRs don't become stale and end up with merge conflicts all the time.

karllessard:
I agree, I was just waiting for James to merge it since he started to review it but I guess he might be off for holidays so to speed up things, I'll do it right away, I think everyone is on the same page here.

karllessard:
Created this issue to follow up on licensing issue that (might) prevent us to enable automatic deployments: https://github.com/tensorflow/java/issues/10

PR #11, opened 2020-01-04
Extend data layouts to map all type of buffers

karllessard:
Right now, data layouts allows to map any user type to one or more bytes. In this pull request, user types can be mapped to any buffer type and not only bytes. While being more generic, it can also reduce the number of I/O operations required for user type that scales over more than one bytes.

For an example of usage, check out the provided unit tests. Also, I have an upcoming PR that will add a `TFloat16` datatype to tensorflow, where floats are mapped directly to short (16-bit) values.


saudet:
> For an example of usage, check out the provided unit tests. Also, I have an upcoming PR that will add a `TFloat16` datatype to tensorflow, where floats are mapped directly to short (16-bit) values.

These types are not specific to TensorFlow, so they shouldn't start with "TF". I don't think TF itself uses any types that were not already widely used, except in Java land that is. :)

karllessard:
> These types are not specific to TensorFlow, so they shouldn't start with "TF". I don't think TF itself uses any types that were not already widely used, except in Java land that is. :)

I don't think we should spend a lot of time to discuss this in this PR since it is a bit unrelated but briefly, the prefix is not "TF" but "T" and stands for "Tensor of". The reason it is not simply called `Float16Tensor` is because it is carried as a generic type for a `Tensor` and `Tensor<Float16Tensor>` is a bit weird. As `Tensor<TFloat16>` is a reference to a tensor of float16, an instance of `TFloat16` is the actual tensor data mapped to an `NdArray<Float>`.

I'll update the documentation in a near future to better explain this design.

karllessard:
I think we are now good to go with this one.

PR #13, opened 2020-01-12
Add API definitions for new V2 operations

karllessard:
When we migrated from TF 1.15 to 2.0 (and now 2.1), we did not generated the [API definitions](https://github.com/tensorflow/java/tree/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def) for the new operations that came in these versions, so they all ended up in the "core" package at the root of the ops hierarchy.

This PR make some attempts to organize the new endpoints in the current hierarchy (note that picking the right group for an operation is a bit arbitrary since there is no clear guidelines on how to do it... @sjamesr ?). 

The PR also include the following changes:

- Actually, it is possible to access a operation group either via a getter method (i.e. `ops.math().add(...)`) or via a public field (i.e. `ops.math.add(...)`), which leads to inconsistency. I removed the methods so only the public fields now remain.

- An operation can now be classified into subgroups, like packages. For example, annotating an operation wrapper with `@Operator(group="group.experimental")` will generate an endpoint accessible from `ops.group.experimental.myOp()`


Craigacp:
Looks like there are some JavaCPP related changes in the middle of this PR? Is that due to an update to TF 2.1?

Otherwise it looks good.

karllessard:
@Craigacp I think the JavaCPP stuff is there due to the fact that when its version was updated in [this PR](https://github.com/tensorflow/java/pull/6/files), the changes in the generated files were not committed.

Some comments on this: It is useful to have generated files in the sources for reference but it comes with its lot of inconveniences, such as making PRs very large on small changes in the generator code, or having Git always reporting local changes just by compiling the library (since the order of processing in the annotation-processor is non-deterministic). Maybe we should reevaluate this decision, to be discussed.

PR #14, opened 2020-01-14
Rename TFloat/TDouble to TFloat32/TFloat64

karllessard:
As voted by the SIG, we change the datatype names to be closer to [those used in Python](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType?version=stable) rather than trying to match them with a Java primitive type.


saudet:
Was there a reason we couldn't name those without a "T" in prefix like Float32, Float64, etc? Why can't we do that? It sounds a lot nicer.

karllessard:
> Was there a reason we couldn't name those without a "T" in prefix like Float32, Float64, etc? Why can't we do that? It sounds a lot nicer.

See reply in group thread: https://groups.google.com/a/tensorflow.org/d/msg/jvm/Zp362xTQahc/Ho6iUEJfAgAJ

PR #15, opened 2020-01-14
Add 16-bit float tensor types

karllessard:
This PR shows how to make use of `DataLayouts` to implement new tensor data types not directly supported by Java primitive types. It adds 2 new tensor types to the Java bindings: `TFloat16` and `TBFloat16`. The former implements the IEEE-754 half-precision floating-point specification while the latter is for truncated 32-bit floats, called "Brain float-16". 

In addition, you'll notice that the PR also include a lot of javadoc updates, which is a work continuously in progress.


Craigacp:
Javadoc looks good.

saudet:
Intel capitalizes it "Bfloat16" so it might be better to do it like that here as well?
https://software.intel.com/en-us/download/bfloat16-hardware-numerics-definition
There are also more instances of "Bfloat16" than "BFloat16" in TensorFlow's main repository,
but it's not too clear cut...

BTW, again, why do we need to prefix all those types with "T"? What's wrong with just `Float16`, etc?

Craigacp:
I think the prefixing helps to distinguish them from things that implement `Number`. At the moment they are more like tag types to make things flow through the type system properly rather than data carriers.

karllessard:
For the prefix, please see my reply in the group thread as this question has been asked in many different places :) https://groups.google.com/a/tensorflow.org/d/msg/jvm/Zp362xTQahc/Ho6iUEJfAgAJ

For the `BFloat16` vs `Bfloat16`, I'm fine with both as well so let's follow @samuel suggestion then, I'll rename it.

karllessard:
What about `TUInt8` vs `TUint8`? Again, both are used in TF core.

karllessard:
Haha, worst, all of `QUInt8`, `QUint8` and `Quint8` are present in TF.

I guess if we go with `TBfloat16`, we should apply the same pattern of capitalizing only the first letter for all types. It looks like that is what the `Go` client is doing. So if everyone is ok, I'll make the change for `TUint8` as well.

PR #16, opened 2020-01-16
String tensor charsets and raw bytes

karllessard:
In previous implementation, UTF-8 character encoding was forced for all string tensors. Furthermore, it was not possible to work directly with sequences of raw bytes.

This PR adds the ability for a user to select which charset to use when encoding/decoding values from a tensor. It is also possible now to read directly the raw bytes of a string tensor and/or to create a tensor from an array of raw bytes. For example:
```java
NdArray<String> strings = NdArrays.vectorOfObjects("Pretty", "vacant");

Tensor<TString> utf8Tensor = TString.copyOf(strings);  // encodes using default UTF-8 charset
assertEquals("Pretty", utf8Tensor.data().getObject(0));  // decodes using default UTF-8 charset

Charset charset = StandardCharsets.UTF_16;
Tensor<TString> utf16Tensor = TString.copyOf(charset, strings); // encodes using UTF-16 charset
assertEquals("Pretty", utf16Tensor.data().use(charset).getObject(0));  // decodes using UTF-16 charset

NdArray<byte[]> utf16Bytes = utf16Tensor.data().asBytes();  // returns raw bytes of the UTF-16 tensor
assertEquals("Pretty", new String(utf16Bytes.getObject(0), charset));

Tensor<TString> utf16TensorCopy = TString.copyOfBytes(utf16Bytes);  // create a tensor from raw bytes
```
Also added to this PR is a safety catch of any exception that might occur when implicitly initializing the data of a tensor, which releases the tensor memory in case of failure.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,7 @@
   </modules>
 
   <properties>
+    <project.build.sourceEncoding>ASCII</project.build.sourceEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <junit.version>4.12</junit.version>
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
@@ -26,6 +26,7 @@ import java.nio.IntBuffer;
 import java.nio.LongBuffer;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.function.Consumer;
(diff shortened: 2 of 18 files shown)


Craigacp:
Looks good. It's a pity there isn't a place to store the charset in the Java Tensor object.

karllessard:
> Looks good. It's a pity there isn't a place to store the charset in the Java Tensor object.

We could but that would only work for tensors that we allocate ourselves. If your string tensor is, let say, the result of an operation or a graph execution, then its charset won't be automatically set.

Therefore, it's better to have a consistent behaviour where the user must know the charset to use in all cases.

PR #18, opened 2020-01-20
Refactor JNI code in C++ into Java code with JavaCPP

saudet:
This is a WIP to remove the manually written JNI code from tensorflow-core-api. In the first commit for SavedModelBundle.java, I replaced about 100 lines of JNI code in C++ with about 40 lines of code in Java, and the unit tests still pass. We need to do the same with the rest of the JNI code, which shouldn't take me that long to do by myself, but if anyone is interested in helping, please let me know and I will give you write access to my fork!

There is also a lot of code in the Java section of the wrappers that is redundant with JavaCPP and that we will be able to remove in a second phase.

/cc @tzolov


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/AbstractTF_Session.java
```diff
@@ -53,6 +63,25 @@ public static TF_Session newSession(TF_Graph graph, TF_SessionOptions opts, TF_S
...
+    public static TF_Session loadSessionFromSavedModel(TF_SessionOptions session_options, TF_Buffer run_options,
+        String export_dir, String[] tags, TF_Graph graph, TF_Buffer meta_graph_def, TF_Status status) {
+        TF_Session s = TF_LoadSessionFromSavedModel(session_options, run_options,
+                new BytePointer(export_dir), new PointerPointer(tags), tags.length, graph, meta_graph_def, status);
+        if (s != null) {
+            s.graph = graph;
+            s.opts = session_options;
+            s.run_options = run_options;
+            s.meta_graph_def = meta_graph_def;
+            s.status = status;
```
karllessard: I'm not sure to understand why we need to set explicitly all the fields here, aren't they initialized properly by the `TF_LoadSessionFromSavedModel` call?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -163,8 +169,46 @@ private static SavedModelBundle fromHandle(
     return new SavedModelBundle(graph, session, metaGraphDef);
   }
 
-  private static native SavedModelBundle load(
-      String exportDir, String[] tags, byte[] config, byte[] runOptions);
+  private static SavedModelBundle load(
+      String exportDir, String[] tags, byte[] config, byte[] runOptions) {
+    SavedModelBundle bundle = null;
+
+    try (PointerScope scope = new PointerScope()) {
```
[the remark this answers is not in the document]
saudet: Yes, the deallocators are added to ReferenceQueue for that purpose, but as usual there is no guarantees that the GC will run at all, so we shouldn't rely on it other than as safety net.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -163,8 +169,46 @@ private static SavedModelBundle fromHandle(
...
+
+      // allocate parameters for TF_LoadSessionFromSavedModel
+      TF_SessionOptions opts = TF_SessionOptions.newSessionOptions();
+      if (config != null && config.length > 0) {
+        TF_SetConfig(opts, new BytePointer(config), config.length, status);
+        status.throwExceptionIfNotOK();
+      }
+      TF_Buffer crun_options = null;
+      if (runOptions != null && runOptions.length > 0) {
+        crun_options = TF_Buffer.newBufferFromString(runOptions);
```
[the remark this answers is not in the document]
saudet: It's not my approach, I just left what was there from the JNI code to make it easier to understand where it comes from, but if that doesn't matter, we can change those names as well.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -163,8 +169,46 @@ private static SavedModelBundle fromHandle(
...
+
+      // handle the result
+      if (metagraph_def.length() > Integer.MAX_VALUE) {
+        throw new IndexOutOfBoundsException("MetaGraphDef is too large to serialize into a byte[] array");
+      } else {
+        byte[] jmetagraph_def = new byte[(int)metagraph_def.length()];
+        new BytePointer(metagraph_def.data()).get(jmetagraph_def);
+        bundle = fromHandle(graph.address(), session.address(), jmetagraph_def);
+        graph.retainReference().deallocate(false);
+        session.retainReference().deallocate(false);
```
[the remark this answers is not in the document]
saudet: If we keep around the original `Pointer` objects, yes that's the basic idea. But ideally we shouldn't try to manage references manually. We can add `PointerScope` fields to `Session` and `Graph` and do something like `attach(graph)` and `attach(session)` to those, and whatever else is associated with those objects, and call `scope.close()` in their `close()` methods. Anyway, like I said, this is all stuff we can look back at in a second phase.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/AbstractTF_Session.java
saudet: No, `TF_LoadSessionFromSavedModel()` is just a C function, it can't keep references to Java objects.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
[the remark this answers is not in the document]
karllessard: Yes, I just wanted to know how you were foreseeing it, cool

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/AbstractTF_Session.java
karllessard: ... oh ok, those fields are declared in the Java class, not the original C++ struct, got it.

saudet: But again, it's not really required, it's just to try to "rein in" the garbage collector a bit more...

karllessard:
@saudet , this PR is still marked as WIP, looks like it's ready to be merged now?

saudet:
I was thinking we could do all of it together before merging? It shouldn't take me more than a few days to do...

karllessard:
As you wish, we can do it iteratively with smaller PRs as well

saudet:
Let's talk about it at the meeting tomorrow :)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
```diff
@@ -131,43 +145,104 @@ public Shape shape(int outputIndex) {
...
 
-  private static native int numDims(long handle);
+  private static int outputListLength(TFE_Op handle, String name) {
+    requireOp(handle);
+    try (PointerScope scope = new PointerScope()) {
+      TF_Status status = TF_Status.newStatus();
+      int length = TFE_OpGetOutputLength(handle, name, status);
+      status.throwExceptionIfNotOK();
+      return length;
+    }
```
karllessard: Instead of starting new scopes in all of these methods, could it be simpler and more efficient to just create the status in a `try-with-resource` block when there are not other resource allocated?
```java
try (TF_Status status = TF_Status.newStatus()) { 
    ... 
}
```

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -448,38 +470,227 @@ public void remove() {
...
+      TF_Operation[] condOutputHandles = new TF_Operation[1];
+      int[] condOutputIndices = new int[1];
+      for (int i = 0; i < ninputs; i++) {
+          condInputHandles[i] = condInputsOutput.position(i).oper();
+          condInputIndices[i] = condInputsOutput.position(i).index();
+      }
+      condOutputHandles[0] = condOutputOutput.oper();
+      condOutputIndices[0] = condOutputOutput.index();
+
+      Object[] cond_output_handles_and_indices =
```
karllessard: Nit: camel case

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -448,38 +470,227 @@ public void remove() {
...
+      TF_Operation[] bodyOutputHandles = new TF_Operation[ninputs];
+      int[] bodyOutputIndices = new int[ninputs];
+      for (int i = 0; i < ninputs; i++) {
+          bodyInputHandles[i] = bodyInputsOutput.position(i).oper();
+          bodyInputIndices[i] = bodyInputsOutput.position(i).index();
+          bodyOutputHandles[i] = bodyOutputsOutput.position(i).oper();
+          bodyOutputIndices[i] = bodyOutputsOutput.position(i).index();
+      }
+
+      Object[] body_output_handles_and_indices =
```
karllessard: Nit: camel case

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -448,38 +470,227 @@ public void remove() {
...
+      params.name(new BytePointer(name));
+
+      // build the while loop, storing loop outputs in `outputs`
+      TF_Output outputs = new TF_Output(ninputs);
+      TF_FinishWhile(params, status, outputs);
+
+      status.throwExceptionIfNotOK();
+
+      // returned array contains both op handles and output indices, in pair
+      Object[] output_handles_and_indices = new Object[ninputs * 2];
```
karllessard: Nit: camel case

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -147,25 +162,92 @@ Shape shape(int outputIdx) {
...
-  private static native String name(long handle);
+  private static void requireHandle(Pointer handle) {
+    if (handle == null || handle.isNull()) {
+      throw new IllegalStateException("close() has been called on the Graph this Operation was a part of");
+    }
+  }
+
+  private static String name(TF_Operation handle) {
+    requireHandle(handle);
+    return TF_OperationName(handle).getString();
```
karllessard: `BytePointer` returned by these methods don't need to be freed/scoped?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -305,13 +324,13 @@ private Run runHelper(boolean wantMetadata) {
...
       for (Output<?> o : inputs) {
-        inputOpHandles[idx] = o.getUnsafeNativeHandle();
+        inputOpHandles[idx] = (TF_Operation)o.getUnsafeNativeHandle();
         inputOpIndices[idx] = o.index();
         idx++;
       }
       idx = 0;
       for (Output<?> o : outputs) {
-        outputOpHandles[idx] = o.getUnsafeNativeHandle();
+        outputOpHandles[idx] = (TF_Operation)o.getUnsafeNativeHandle();
```
karllessard: instead of all those explicit casts everywhere, maybe we can add 2 methods to `Output`: `getGraphOperationUnsafeNativeHandle` and `getEagerOperationUnsafeNativeHandle`? I think the code would end up cleaner.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -440,15 +459,63 @@ public Runner runner() {
...
+    }
+  }
+
+  private static void resolveHandles(String type, Pointer[] src, PointerPointer dst, int n) {
+    if (src.length != n) {
+      throw new IllegalArgumentException("expected " + n + ", got " + src.length + " " + type);
+    }
+    for (int i = 0; i < n; ++i) {
+      if (src[i] == null || src[i].isNull()) {
+        throw new NullPointerException("invalid " + type + " (#" + i + " of " + n + ")");
```
karllessard: I don't know what was thrown by the original JNI code but I think we should throw ourselves NPE, `IllegalStateException` sounds a better choice. Same thing in `allocate2` below.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -477,15 +544,49 @@ public Runner runner() {
...
+      TF_SessionRun(handle, runOpts, inputs, inputValues, ninputs,
+                    outputs, outputValues, noutputs, targets, ntargets,
+                    runMetadata, status);
+      status.throwExceptionIfNotOK();
+
+      for (int i = 0; i < noutputs; ++i) {
+        outputTensorHandles[i] = outputValues.get(TF_Tensor.class, i);
+      }
+
+      return runMetadata != null ? runMetadata.get() : null;
```
karllessard: Question: did you ever ran a benchmark showing that all these context switches between the JVM and the native code (i.e. at each JavaCPP generated method) ends up to be as performant as when only one call was made (`run` in this case)?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -821,35 +848,505 @@ private void throwExceptionIfTypeIsIncompatible(Object o) {
...
+        assert Loader.sizeof(LongPointer.class) == 8 :
+               "Java long not compatible with TF_INT64";
+        return 8;
+      default:
+        return 0;
+    }
+  }
+
+  /** Write a Java scalar object (java.lang.Integer etc.) to a TF_Tensor. */
+  private static void writeScalar(Object src, int dtype, BytePointer dst, long dstSize) {
```
karllessard: Most of the stuff here and below in this class should ultimately be dropped and replaced with the usage of TF Tools to read or write data from/to the tensor.

I guess we can migrate it to JavaCPP meanwhile and I can take care of the cleanup after.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -821,35 +848,505 @@ private void throwExceptionIfTypeIsIncompatible(Object o) {
...
+        throw new IllegalStateException("invalid DataType(" + dtype + ")");
+    }
+    return sz;
+  }
+
+  private static long writeNDArray(Object src, int dtype, int dimsLeft,
+                                   BytePointer dst, long dstSize) {
+    if (dimsLeft == 1) {
+      return write1DArray(src, dtype, dst, dstSize);
+    } else {
```
karllessard: No need of an `else` condition

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -821,35 +848,505 @@ private void throwExceptionIfTypeIsIncompatible(Object o) {
...
+      writeNDArray(value, dtype, numDims, data, sz);
+    }
+  }
 
-  private static native float scalarFloat(long handle);
+  private static float scalarFloat(TF_Tensor handle) {
+    requireHandle(handle);
+    if (TF_NumDims(handle) != 0) {
+      throw new IllegalStateException("Tensor is not a scalar");
+    } else if (TF_TensorType(handle) != TF_FLOAT) {
```
karllessard: No need of `else` conditions (applicable to all methods below)

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -535,10 +535,10 @@ public void eagerTensorIsReleasedAfterSessionIsClosed() {
       Output<?> x = TestUtil.constant(session, "Const1", 10);
       Output<?> y = TestUtil.constant(session, "Const2", 20);
       sum = TestUtil.<TInt32>addN(session, x, y).tensor();
-      assertNotEquals(0L, sum.getNativeHandle());
+      assertNotEquals(null, sum.getNativeHandle());
```
karllessard: `assertNotNull`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
```diff
@@ -131,43 +145,104 @@ public Shape shape(int outputIndex) {
...
+  private static void requireOp(TFE_Op handle) {
+    if (handle == null || handle.isNull()) {
+      throw new IllegalStateException("Eager session has been closed");
+    }
+  }
 
-  private static native long resolveTensorHandle(long handle);
+  private static void requireTensorHandle(TFE_TensorHandle handle) {
+    if (handle == null || handle.isNull()) {
+      throw new IllegalStateException("EagerSession has been closed");
```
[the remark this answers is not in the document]
saudet: I just copy/pasted...

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
```diff
@@ -131,43 +145,104 @@ public Shape shape(int outputIndex) {
...
+  }
 
-  private static native int dataType(long handle);
+  private static TF_Tensor resolveTensorHandle(TFE_TensorHandle handle) {
+    requireTensorHandle(handle);
+    try (PointerScope scope = new PointerScope()) {
+      TF_Status status = TF_Status.newStatus();
+      TF_Tensor tensor = TFE_TensorHandleResolve(handle, status);
+      status.throwExceptionIfNotOK();
+      return tensor;
```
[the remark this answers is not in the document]
saudet: The convention here is that we leave original C/C++ functions behave like they do in C/C++, where they return an integer for pointer addresses and what not. Then we add manually factory methods like `TF_Status.newStatus()` that wrap things up with deallocators to support garbage collection and reference counting, calling them them only when we want those features.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
saudet: Yes, it would be more efficient, but it would also make it more error prone when we start creating other objects in there that may start doing temporary allocations.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
saudet: No, as explained above, by convention, JavaCPP doesn't perform any allocation for directly mapped C/C++ functions.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
saudet: Well, it's not that bad. Though, I agree that all those weird methods relying on `Object` references should be totally refactored/removed, but let's leave them like they are for now instead of coming up with workarounds?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
saudet: Calls from the JVM to native code are very efficient, in the order of 30 ns. On the other hand, calls from native code to the JVM are very expensive, in the order of 300 ns, so it's almost certain that the new code here is going to faster. I'll run a simple benchmark and post the results here just to confirm, but if you're worried performance, we should think about writing a whole set of benchmarks to make sure there is never any regression in performance anywhere.

saudet:
I ran the `initTensorByArrays` benchmark from pull https://github.com/tensorflow/java/pull/20#issuecomment-577983662 with JMH's default settings and here is the kind of result I am getting on my machine:
- Before (manually written JNI code):
    ```
    Benchmark                        Mode  Cnt   Score   Error  Units
    MyBenchmark.initTensorByArrays  thrpt   25  93.895 ± 1.580  ops/s
    ```
- After (JNI refactored into Java with JavaCPP):
    ```
    Benchmark                        Mode  Cnt    Score   Error  Units
    MyBenchmark.initTensorByArrays  thrpt   25  110.598 ± 1.012  ops/s
    ```
So it does look like this refactoring also increases performance!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
karllessard: Yeah... I would still prefer we go with the most efficient approach but it's up to you if you want to make the changes or not, we can merge it like this too.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
karllessard: Ok, we can do this later if it gets really nasty.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
karllessard: I've started to write some of them as well (like [this one](https://github.com/tensorflow/java/blob/master/tensorflow-tools/src/test/java/org/tensorflow/tools/benchmark/NdArrayBenchmark.java)), I think with high-performance projects like TF we should have a set of benchmarks, not only it is useful to detect regression but also is a very good indicator of where we should spend time for optimization. 

If you can add some, that would be great, but we can merge without if you want to create these later.

PR #24, opened 2020-01-29
Adds a (previously published) fix for Ubunu 19.10 (eoan)

dhruvrajan:
See this [patch](https://github.com/clearlinux-pkgs/tensorflow/blob/master/Add-grpc-fix-for-gettid.patch)  from clearlinux which enables building of tensorflow on Ubuntu 19.10. Adding this patch to be used during tensorflow-java's build.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/WORKSPACE
+++ b/tensorflow-core/tensorflow-core-api/WORKSPACE
@@ -11,7 +11,8 @@ http_archive(
     name = "org_tensorflow",
     patches = [
         ":tensorflow-windows.patch", # https://github.com/tensorflow/tensorflow/issues/25213
-        ":tensorflow-api-def.patch"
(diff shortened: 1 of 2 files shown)


dhruvrajan:
This is a bugfix; I used to get a build error (matching https://github.com/clearlinux/distribution/issues/1151) when building the `java` repo; adding their fix solved the issue. If there's a way we can get around adding this patch file by changing a library version, that does sound like a better approach! This seemed to be simple fix for anyone developing currently on 19.10.

karllessard:
Interesting thread to follow on this topic: https://github.com/tensorflow/tensorflow/issues/33758

Still pretty active :) Looks like they are having trouble to upgrade grpc, I'll merge this PR but keeping a look at their progress, if an upgrade finally fix the problem, will revert it and apply new changes.

PR #27, opened 2020-02-07
Deterministic ops generation

Craigacp:
This enforces a deterministic order on the ops generation by passing each `OpsSpec`'s method collection through a comparator before building the Java output. The comparator first compares the method name using String's natural ordering, then compares the number of arguments to the methods, then compares the name of each method argument at a specific position using String's natural ordering.

The only real edit in this commit is to `OperatorProcessor.java` the rest is just the rearrangement of the ops according to the above comparator. Unfortunately this means it does rearrange the ops again, but it should be for the final time.

### DIFF
--- a/tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/processor/operator/OperatorProcessor.java
+++ b/tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/processor/operator/OperatorProcessor.java
@@ -160,19 +160,18 @@ public final class OperatorProcessor extends AbstractProcessor {
         return 1;
       } else if (o1.parameters.size() < o2.parameters.size()) {
         return -1;
-      } else {
-        List<ParameterSpec> firstParams = o1.parameters;
-        List<ParameterSpec> secondParams = o2.parameters;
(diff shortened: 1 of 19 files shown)


tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -154,17 +155,39 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
   }
 
   private static class OpsSpec {
+    private static final Comparator<MethodSpec> PARAMETER_SPEC_COMPARATOR = (o1, o2) -> {
+      if (o1.parameters.size() > o2.parameters.size()) {
+        return 1;
+      } else if (o1.parameters.size() < o2.parameters.size()) {
```
karllessard: no need of `else` blocks if you return in the previous `if`

Craigacp: Ok, I'll remove it.

PR #29, opened 2020-02-08
Endpoint annotation

karllessard:
This adds a new `@Endpoint` annotation that must be added to the methods of a class annotated with `@Operator` to be added as an endpoint in one of the `Ops` classes.

Actually, this was done by matching any factory method called `create`, which lacks in flexibility. The usage of `@Endpoint` is more explicit to identify which methods are added to the `Ops` classes. Plus, it adds a lot more flexibility, for instance by allowing overriding the default group and name for these endpoints. For example:

```java
@Operator
public class MyOp implements Op {

    @Endpoint
    public static MyOp create(Scope scope, ...) {
         ....
    }

    @Endpoint(group="exec", name="computeMyOp")
    public static Float compute(Scope scope, ...) {
        ....
     }
}
```
will create the two following endpoints, given `tf` is an intance of `Ops`: 
* `MyOp tf.myOp(...)`
* `Float tf.exec.computeMyOp(...)`.

Please take note that while being pretty large, this PR principally modifies the `op_generator.cc` and `OperatorProcessor.java` classes, other changes are result of the generated classes being updated.

CC: @Craigacp can you please take a look if that could help for returning a instance of `Variable` in your new `VariableWithInit` operator?


Craigacp:
The code works just fine. The generated Javadoc is not ideal.
```
   /**
-   * Builds an {@link VariableWithInit} operation
+   * Builds an {@link CoreOps} operation
    *
    * @param init The op to use to initialise this variable.
    * @param options carries optional attributes values
    * @return a new instance of Variable
-   * @see org.tensorflow.op.core.VariableWithInit
+   * @see org.tensorflow.op.core.CoreOps
    */
   public <T extends TType> Variable<T> variableWithInit(Operand<T> init,
       Variable.Options... options) {
-    return VariableWithInit.create(scope, init, options);
+    return CoreOps.createVariableWithInit(scope, init, options);
   }
 ```
The `@see` tag should link to the specific method, and the "Builds an XXX operation" bit should give the return type of the method (or we add an additional annotation field which specifies what appears in that bit of the Javadoc).

**Comment by Craigacp** [comment]
For reference the CoreOps file is:
```

@Operator
public class CoreOps {

    /**
     * Factory method to create a new Variable with it's initializer.
     *
     * @param scope current scope
     * @param init The op to use to initialise this variable.
     * @param options carries optional attributes values
     * @return a new instance of Variable
     */
    @Endpoint(name="variableWithInit")
    public static <T extends TType> Variable<T> createVariableWithInit(Scope scope, Operand<T> init, Variable.Options... options) {
        Output<T> initOutput = init.asOutput();
        Variable<T> newVar = Variable.create(scope,initOutput.shape(),initOutput.dataType(),options);
        Assign<T> assignOp = Assign.create(scope,newVar,init);
        ExecutionEnvironment exEnv = scope.env();
        if (exEnv instanceof Graph) {
            Graph graph = (Graph) exEnv;
            graph.addInitializer(assignOp);
        }

        return newVar;
    }
}
```

karllessard:
> The code works just fine. The generated Javadoc is not ideal.

Sounds good, I will take a look how to improve the Javadoc, thanks

karllessard:
> The `@see` tag should link to the specific method

Ok now I remember why it is done like that, it's because the interesting documentation of the generated operators (i.e. the one that comes from TF kernel doc directly) is at the class level, not the method level. I suggest that we leave it like that for now, I'll just fix the returned type.

Craigacp:
But in the case where we have multiple helper methods in the same class it's not going to be particularly useful. The `variableWithInit` method doesn't really warrant it's own class, and I suspect we'll have more methods like it while we build out things. Maybe we could put a flag on the `@Endpoint` annotation that means it inserts a link to the method rather than the class, and default it to false?

karllessard:
> The variableWithInit method doesn't really warrant it's own class, and I suspect we'll have more methods like it while we build out things. Maybe we could put a flag on the @Endpoint annotation that means it inserts a link to the method rather than the class, and default it to false?

We could yes... at the same time, I don't dislike the fact that all operators have their own class, it allows you to quickly retrieve the source file implementing them just by guessing the name (hence `VariableWithInit.java` in this case). 

But nonetheless, I think we could play a bit with the Javadoc. What about copying the whole class or method doc instead of linking to it? It will generate huge source files but it is probably more convenient for quick consultation in an IDE.

karllessard:
I just pushed a new version that copies either the description of the method annotated with `@Endpoint` (default behaviour) or the class annotated with `@Operator` (optional behaviour) to the endpoint documentation.

This way, the javadoc of the endpoints is way more relevant for quick referrals from a IDE, avoiding to jump to the operator class from the `@see` tag. The new attribute `describeByClass` can be used to switch from one description source to the other (default is false).

karllessard:
Ok, I think we are fine going with this one, everything that we agreed on has been implemented, and it will help unblock the other PRs.

PR #31, opened 2020-03-10
NdArrays and DataBuffers equality and hash codes

karllessard:
This PR adds the ability to compare `NdArray` instances (or their underlying `DataBuffer`) for equality check, so it is now possible to do something like:
```java
IntNdArray array = NdArrays.ofInts(2).setInt(10, 0).setInt(20, 1);
assertEquals(array, StdArrays.ndCopyOf(new int[]{10, 20}));
```
The comparison algorithms are not optimal for performances but it should be ok for a first revision and we can improve it later (e.g. by the use of native `memcmp` whenever possible).

Still, I added a `DataStorageVisitor` that allows us to optimize a few paths for data copy and comparison (e.g. by invoking direction `java.nio` methods whenever possible). 

Finally, there's been a few method renaming in `tensorflow-tools` as well, for consistency, and I have dropped the `StaticApi` helper class, which was duplicating code without providing real benefits. 

CC: @Craigacp


tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/AbstractDataBuffer.java
```diff
@@ -38,9 +41,145 @@
...
+    while (++idx < size()) {
+      if (!Objects.equals(other.getObject(idx), getObject(idx))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private int arrayHashCode(long startIdx, Class<?> arrayClass) {
+    Function<Long, Integer> hashCoder;
```
Craigacp: Can we stash these lambdas in static variables?

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/adapter/DoubleDataBufferAdapter.java
```diff
@@ -56,16 +56,48 @@ public DoubleDataBuffer write(double[] src, int offset, int length) {
     return this;
   }
 
+  @Override
+  public DoubleDataBuffer copyTo(DataBuffer<Double> dst, long size) {
+    Validator.copyToArgs(this, dst, size);
+    if (dst instanceof DoubleDataBuffer) {
+      for (long idx = 0L; idx < size; ++idx) {
+        ((DoubleDataBuffer)dst).setDouble(getDouble(idx), idx);
```
Craigacp: Might be best to hoist the cast out of the for loop in case the JIT doesn't figure it out. Pity we can't use the pattern matching from JDK 14 yet.

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/NdArray.java
```diff
@@ -293,4 +291,20 @@ default long size() {
    * @see DataBuffer#size()
    */
   NdArray<T> write(DataBuffer<T> src);
+
+  /**
+   * Checks equality between n-dimensional arrays.
+   *
+   * <p>An array is equal to another object if this object is another {@link NdArray} of the
```
Craigacp: This should probably note "the same shape, type and the elements are equal and in the same order".

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
```diff
@@ -223,4 +223,37 @@
...
+   * @return the same value returned by the visitor
+   */
+  default <R> R accept(DataStorageVisitor<R> visitor) {
+    return visitor.otherwise();
+  }
+
+  /**
+   * Checks equality between data buffers.
+   *
+   * <p>A data buffer is equal to another object if this object is another {@link DataBuffer} of the
```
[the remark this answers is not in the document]
karllessard: It does depend on the type as well, that was missing in the doc, I'll add it

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataStorageVisitor.java
```diff
@@ -0,0 +1,147 @@
...
+   * Fallback method called if the visitor implementation does not support the type of backing storage
+   * for a given {@link DataBuffer}
+   *
+   * <p>The implementor of this interface must override the {@code visit} methods for type of storage
+   * it supports. If {@link DataBuffer#accept(DataStorageVisitor)} is called on a buffer
+   * using a different type of storage, the invocation will fallback to this method.
+   *
+   * @return any value
+   */
+  R otherwise();
```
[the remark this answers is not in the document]
karllessard: This method is called when the visitor does not handle a given type of data storage, so that it can fallback to a generic task if needed. I still prefer `otherwise()` to `defaultVisit()` because basically when this method is called, it means that you are not visiting any data storage. I used it for example like this:

```java
    return otherBuffer.accept(new DataStorageVisitor<Boolean>() {

      @Override
      public Boolean visit(DoubleBuffer otherStorage) {
        return myStorage.equals(otherStorage);
      }

      @Override
      public Boolean otherwise() {
        for (int idx = 0; idx < size(); ++idx) {
          if (otherBuffer.getDouble(idx) != getDouble(idx)) {
            return false;
          }
        }
        return true;
      }
    });
```

So when `otherwise()` is call, I can only build a routine using the public API of a `DataBuffer` and nothing internal like other `visit()` methods can do.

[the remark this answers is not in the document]
Craigacp: I'd still consider that to be visiting every index in the storage, but I don't write that many visitors to be honest. Maybe `fallback`? The name `otherwise` to me suggests there should be some condition that makes it go down that path that the user could control, or is data dependent. The use case here seems to be type dependent (in that it'll be chosen statically at compile time).

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/dense/AbstractDenseNdArray.java
```diff
@@ -110,6 +113,18 @@ protected AbstractDenseNdArray(DimensionalSpace dimensions) {
 
   abstract protected DataBuffer<T> buffer();
 
+  protected DataBuffer<T> trimBuffer() {
```
Craigacp: This is disappointing, is there no way to make `equals` and `hashCode` just look at the view without having to actually perform an operation on them?

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataStorageVisitor.java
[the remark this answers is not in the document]
karllessard: Ok no problem for "fallback", as long as it is not "visit" I'm fine.

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/dense/AbstractDenseNdArray.java
karllessard: We could avoid it by trimming the buffer every time we slice the array but this function is more on the critical performance path and I was hesitating changing it only to support equals/hashCode, which should be used less frequently. 

Trimming the buffer when the array is not segmented (i.e. caused by the usage of some of the `Index` classes) is quite easy to do but substantial changes need to be performed in the other case.

I'm not too happy with this last-minute discovery as well so would you be comfortable if we keep it like that and I'll add a TODO notice to take a deeper look at how to improve this aspect later?

Craigacp: Yeah, I'm fine with it staying as is with a note to fix it later.

Craigacp: Maybe we could add a flag on the NdArray noting it's not continuous in it's backing store, then have equals and hashcode fall back to something slower?

karllessard: Oh, we already have this flag (`dimensions.isSegmented()`), so I would now assume the buffer is trimmed if this flag is not set? I think it is better but still not perfect, I'll do this but still leave a note to review for later.

karllessard: Ok @Craigacp , I just pushed new substantial changes, this was required if I wanted to limit the impact on performances when slicing an array. So I added a new `DataBuffer.slice()` method, instead of using the double-allocating `DataBuffer.offset(i).narrow(s)` (running `NdArraBenchmark` show a significative improvement by doing this).

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/nio/FloatNioDataBuffer.java
```diff
@@ -17,9 +17,11 @@
 
 package org.tensorflow.tools.buffer.impl.nio;
 
+import java.nio.DoubleBuffer;
 import java.nio.FloatBuffer;
 import org.tensorflow.tools.buffer.DataBuffer;
 import org.tensorflow.tools.buffer.DataStorageVisitor;
+import org.tensorflow.tools.buffer.DoubleDataBuffer;
```
Craigacp: Looks like some unnecessary imports got dragged in.

PR #34, opened 2020-03-19
Cleanup tensor factories and methods

karllessard:
This is the (hopefully) last PR related to the integration of TF Tools with the core API. It cleans up many obsolete `Tensor` factories and reconfigure unit tests for making use of the new `DataBuffer` and `NdArray` classes. 

Important breaking changes are:

* Complete removal of the `Tensors` helper class to instantiate tensors from Java primitives data structure

     * This has been done with the assumption that in most cases, `Tensor` are allocated to feed graph with variable data, not static one (such as `new int[][]{{1, 2}, {3, 4}}`). Static constants like these are normally used for creating `Constant` ops instead, and can still be done using `tf.val()`
     * As we start to support more datatypes, user should be more explicit on the desired type of tensor he wants to allocate when feeding a graph, by using one of the `TType` factory instead (e.g. `TInt32.tensorOf(Shape.of(2, 2))`. 

* Removal of the 'tensor to scalar' endpoints in the `Tensor` class (e.g. `Tensor.floatValue()`)

     * The reason is because those endpoints were not type safe like all other endpoints, and would fail at runtime if the wrong accessor was invoked on a tensor of a given type.
     * The safe way to do it now is `Tensor.data().getFloat()`, assuming that the `Tensor` instance is typed properly (when type of tensor is unknown, `Tensor.expect()` should be called prior to accessing the data).


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -82,241 +39,142 @@
...
+   * @param size size, in bytes, of the tensor
+   * @return an allocated but uninitialized tensor
+   * @see #of(DataType, Shape)
    */
-  public static Tensor<TInt32> create(long[] shape, IntBuffer data) {
-    Tensor<TInt32> t = allocateForBuffer(TInt32.DTYPE, shape, data.remaining());
-    t.buffer().asIntBuffer().put(data);
+  public static <T extends TType> Tensor<T> of(DataType<T> dtype, Shape shape, long size) {
+    Tensor<T> t = new Tensor<>(dtype, shape);
+    TF_Tensor nativeHandle = allocate(t.dtype.nativeCode(), shape.asArray(), size);
```
Craigacp: Does this call check that the size is larger than the required size for the shape & dtype?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Constant.java
```diff
@@ -515,22 +502,15 @@
...
-   * method.
-   *
    * @param scope is a scope used to add the underlying operation.
    * @param shape the tensor shape.
    * @param data a buffer containing the tensor data.
    * @return a double constant
    * @throws IllegalArgumentException If the tensor shape is not compatible with the buffer
-   * @deprecated use {@link Ops#val(Tensor) Ops.constant(Tensor&lt;TFloat64>)} instead
    */
   @Endpoint
```
Craigacp: This one is missing the `@Endpoint(name = "val")`.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/AutoCloseableList.java
```diff
@@ -0,0 +1,27 @@
+package org.tensorflow;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+public final class AutoCloseableList<E extends AutoCloseable> extends ArrayList<E>
```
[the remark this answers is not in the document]
karllessard: This class is only used for unit tests and I just moved it from its previous location in [`TestUtil`](https://github.com/tensorflow/java/blob/1b6047804f4e6a4faf2606401502f64fc2437162/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/TestUtil.java#L27), as it was the only utility remaining

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: I think the native part validate it but I can add a unit test to confirm this

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/AutoCloseableList.java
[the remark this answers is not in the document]
Craigacp: Ah ok, I'd missed it was in the test code. We might want to consider doing something for the return value of session.run anyway as it is a bit irritating. Not that I'm too happy with how it turned out for onnxruntime, but it does work and it's safe.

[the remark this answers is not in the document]
karllessard: I agree, it is irritating. We should address that but let’s do it in another PR

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
karllessard: Ok, so after some investigation, I can see that the native call returns null if the buffer is **smaller** than the required size (a larger buffer is accepted though so we should preserve that behaviour) or if it fails to allocate the tensor for any other reason.

In TF 1.15, a `NullPointerException` was [explicitly thrown](https://github.com/tensorflow/tensorflow/blob/8ca7de134ca196eebc43108ba79c643c2f5ac223/tensorflow/java/src/main/native/tensor_jni.cc#L358) in such case, which was a bit odd. @saudet changed it to a `IllegalStateException` in the [JavaCPP version](https://github.com/tensorflow/java/blob/1b6047804f4e6a4faf2606401502f64fc2437162/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java#L1126), which is better, but still misleading in case the mistake comes from a user-provided parameter (there is actually no way to know it from the native call).

So I've added an explicit check on the minimum tensor size requirement and now throw an `IllegalArgumentException` if that check does not pass. This is only possible with fixed-length datatypes, as we don't know the size of the elements for variable-length datatypes (`TString` factories will take care of this).

PR #35, opened 2020-03-23
Removed unnecessary examples for shape attribute which is not there in tensor_scatter_nd_add

ashutosh1919:
Fixes [#37669](https://github.com/tensorflow/tensorflow/issues/37669).
Removed code corresponding to example doc string for `shape` parameter which is not present in `tensor_scatter_nd_add`.

@mihaimaruseac, please review.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/core/TensorScatterNdAdd.java
+++ b/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/core/TensorScatterNdAdd.java
@@ -37,18 +37,6 @@ import org.tensorflow.types.family.TType;
  * are added onto an existing tensor (as opposed to a variable). If the memory
  * for the existing tensor cannot be re-used, a copy is made and updated.
  * <p>
- * `indices` is an integer tensor containing indices into a new tensor of shape
- * `shape`.  The last dimension of `indices` can be at most the rank of `shape`:
- * <p>
- *     indices.shape[-1] <= shape.rank
- * <p>
- * The last dimension of `indices` corresponds to indices into elements
- * (if `indices.shape[-1] = shape.rank`) or slices
- * (if `indices.shape[-1] < shape.rank`) along dimension `indices.shape[-1]` of
- * `shape`.  `updates` is a tensor with shape
- * <p>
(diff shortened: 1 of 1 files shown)


zaccharieramzi:
@ashutosh1919 I think the docs should still mention the shape constraints. The problem with the current documentation is that it doesn't refer to the correct shape (`tensor.shape`) but to an undefined `shape` argument.

ashutosh1919:
@zaccharieramzi , please check. Is it okay now?

zaccharieramzi:
I guess it's fine.

karllessard:
Hi @ashutosh1919 , thanks for your PR,

All files found under `src/gen` of `tensorflow-core-api` are generated at compile time and any changes made directly to them will be discarded on the next build.

The Javadoc found in the generated operator classes are actually issued from the description found in the TF main repository, in [this file](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/api_def/base_api/api_def_TensorScatterAdd.pbtxt) for this specific case. While they should be generic, most of these descriptions are unfortunately Python-oriented and misleading when imported into Java classes.

So the right way of fixing this would be copy the [description](https://github.com/tensorflow/tensorflow/blob/cbc5ab117445ac0328311df8abea61be6256b92f/tensorflow/core/api_def/base_api/api_def_TensorScatterAdd.pbtxt#L28) field into the [Java overrides](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_TensorScatterAdd.pbtxt) for this operator and apply your fix in this file.

Craigacp:
The original issue #37669 is a problem with the python docs, you'll need to take that up with whoever generates them from the protobuf file. The best fix is to change the protobuf file Karl referenced in https://github.com/tensorflow/tensorflow.

ashutosh1919:
@karllessard and @Craigacp , Thank you so much for your comments. Today, I learned this new thing. I   will keep this thing in mind for future contributions.
I have raised new PR on tf main repo [#37832](https://github.com/tensorflow/tensorflow/pull/37832).  
Also, we can add descriptions for java too as per @karllessard commented. But since, I am not aware about the java part of tf, I am closing this PR.
Thanks for help!

PR #37, opened 2020-04-04
Add GitHub Actions job that redeploys snapshots to get consistent timestamps

saudet:
As required for most build tools other than Maven, that is Gradle, sbt, etc.

On GitHub Actions, the result looks like this at
https://github.com/saudet/tensorflow-java/runs/559706945
which ultimately deployed those files:

```
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-gpu.jar.md5 | Sat Apr 04 01:58:54 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-gpu.jar.sha1 | Sat Apr 04 01:58:54 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl-gpu.jar | Sat Apr 04 01:58:54 UTC 2020 | 152881432 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl-gpu.jar.md5 | Sat Apr 04 01:59:01 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl-gpu.jar.sha1 | Sat Apr 04 01:59:01 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl.jar | Sat Apr 04 01:59:01 UTC 2020 | 39131128 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl.jar.md5 | Sat Apr 04 01:59:02 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64-mkl.jar.sha1 | Sat Apr 04 01:59:02 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64.jar | Sat Apr 04 01:59:02 UTC 2020 | 36061068 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64.jar.md5 | Sat Apr 04 01:59:03 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-linux-x86_64.jar.sha1 | Sat Apr 04 01:59:03 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64-mkl.jar | Sat Apr 04 01:59:03 UTC 2020 | 46506587 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64-mkl.jar.md5 | Sat Apr 04 01:59:04 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64-mkl.jar.sha1 | Sat Apr 04 01:59:04 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64.jar | Sat Apr 04 01:59:05 UTC 2020 | 45371550 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64.jar.md5 | Sat Apr 04 01:59:06 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100-macosx-x86_64.jar.sha1 | Sat Apr 04 01:59:06 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100.jar | Sat Apr 04 01:58:37 UTC 2020 | 2267498 |  
tensorflow-core-api-0.1.0-20200404.015837-100.jar.md5 | Sat Apr 04 01:58:37 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100.jar.sha1 | Sat Apr 04 01:58:37 UTC 2020 | 40 |  
tensorflow-core-api-0.1.0-20200404.015837-100.pom | Sat Apr 04 01:58:37 UTC 2020 | 416 |  
tensorflow-core-api-0.1.0-20200404.015837-100.pom.md5 | Sat Apr 04 01:58:37 UTC 2020 | 32 |  
tensorflow-core-api-0.1.0-20200404.015837-100.pom.sha1 | Sat Apr 04 01:58:37 UTC 2020 | 40 |  
```
https://oss.sonatype.org/content/repositories/snapshots/org/tensorflow/tensorflow-core-api/0.1.0-SNAPSHOT/

Also remove from GPU platform artifacts unavailable dependencies for Mac

/cc @frankfliu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,3 +124,24 @@ jobs:
           call mvn clean %MAVEN_PHASE% -Possrh -B -U -e -Djavacpp.platform=windows-x86_64 -Djavacpp.platform.extension=${{ matrix.ext }}
           df -h
           wmic pagefile list /format:list
+  redeploy:
+    if: github.event_name == 'push'
+    needs: [linux-x86_64, macosx-x86_64]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+      - name: Redeploy snapshot artifacts
+        run: |
+          cd tensorflow-core
+          mvn dependency:resolve -Possrh -Predeploy -N -B -U -e -Djavacpp.platform.custom -Djavacpp.platform.linux-x86_64 -Djavacpp.platform.macosx-x86_64
+          cp $HOME/.m2/repository/org/tensorflow/tensorflow-core-api/*-SNAPSHOT/tensorflow-core-api-*-SNAPSHOT*.jar .
+          for f in *.jar; do
+            if [[ $f =~ tensorflow-core-api-.*SNAPSHOT-(.*).jar ]]; then
+              [[ -n $FILES ]] && FILES=$FILES,$f || FILES=$f
+              [[ -n $TYPES ]] && TYPES=$TYPES,jar || TYPES=jar
+              [[ -n $CLASSIFIERS ]] && CLASSIFIERS=$CLASSIFIERS,${BASH_REMATCH[1]} || CLASSIFIERS=${BASH_REMATCH[1]}
+            fi
+          done
+          echo "<settings><servers><server><id>ossrh</id><username>${{ secrets.CI_DEPLOY_USERNAME }}</username><password>${{ secrets.CI_DEPLOY_PASSWORD }}</password></server></servers></settings>" > $HOME/.m2/settings.xml
+          mvn clean deploy -Possrh -Predeploy -N -B -U -e -Dfiles=$FILES -Dtypes=$TYPES -Dclassifiers=$CLASSIFIERS
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,33 @@
     <skipAssembly>true</skipAssembly>
   </properties>
 
+  <repositories>
+    <repository>
+      <id>sonatype-nexus-snapshots</id>
+      <name>Sonatype Nexus Snapshots</name>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>sonatype-nexus-snapshots</id>
+      <name>Sonatype Nexus Snapshots</name>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </pluginRepository>
+  </pluginRepositories>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
--- a/tensorflow-core/pom.xml
+++ b/tensorflow-core/pom.xml
@@ -56,6 +56,60 @@
   </properties>
 
   <profiles>
+    <!-- Redeploys in one shot the snapshot artifacts of tensorflow-core-api
+         given to maven-deploy-plugin to get consistent timestamps among them -->
+    <profile>
+      <id>redeploy</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.tensorflow</groupId>
+          <artifactId>tensorflow-core-platform</artifactId>
+          <version>${project.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.tensorflow</groupId>
+          <artifactId>tensorflow-core-platform-mkl</artifactId>
+          <version>${project.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.tensorflow</groupId>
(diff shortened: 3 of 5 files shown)


karllessard:
Thanks @saudet , 

I'm not too sure this is the right approach to fix our snapshot issues. Unless I'm missing something, we'll continue to deploy snapshots artifacts individually and then we have an extra step to fetch back those deployed artifacts to a single machine and redeploy them at once.

The problems I'm foreseeing is that between the time the first artifact is deployed and the `redeploy` step is executed, our users might have trouble to download the snapshots (this period of time may vary between a few minutes or a few hours I'd guess, depending on the speed of the individual GitHub machines).

Also, the process of fetching back the artifacts by resolving Maven dependencies may fail if some snapshots artifacts have already been discarded by desynchronized timestamps from the individual deployments, the same way we observe right now.

Can we think of a solution that would imply artifact staging, like was proposing initially @frankfliu? At a first glance, the Sonatype Nexus maven plugin sounds promising with its server-side staging feature.

What do you think?

frankfliu:
I actually not even sure this re-download can work. The timestamp already
messed up when 2nd deploy happen.

Instead of deployment artifact in the first round of build, we can upload
artifacts to github, And download them back at final publish job.

Thanks,
Frank

karllessard:
> I actually not even sure this re-download can work. The timestamp already
messed up when 2nd deploy happen.

Yes, that's what I was pointing out as well in my second concern. 

> Instead of deployment artifact in the first round of build, we can upload
artifacts to github

I think that should work too, I'm sharing here the github actions to upload/download artifacts you have previously mentioned in the mailing thread:  
[actions/upload-artifact@v1](https://github.com/actions/upload-artifact), [actions/download-artifact@v1](https://github.com/actions/download-artifact)

saudet:
I'm 100% sure it works fine. I never had any issues at all with this for
over 5 years. If you feel differently, at least please provide proof!

saudet:
BTW, Nexus provides no synchronization when uploading snapshots, so if
Gradle isn't able to wait for all artifacts to be uploaded, it's going to
fail anyway until everything is uploaded.

In any case, I keep telling you guys. I'm fine with whatever gets the job
done. If you know how to fix this exactly like you want it, please do it. I
don't intend to spend more time fixing something that doesn't need fixing,
especially when it's Gradle who should be fixing their code, not us.

saudet:
And yes, I meant that the developers of Gradle do not want to fix their code, see https://github.com/gradle/gradle/issues/2882.

frankfliu:
Thanks for detailed explanation. Just for my understanding:
1. Existing publishing should work for maven since it keeps the
snapshotVersion history. But gradle will fail to download it.
2. The re-deploy is purely workaround for gradle

Am I right?

Thanks,
Frank

karllessard:
If this works with Maven, then I would like to understand why we missed macosx artifacts in Nexus a few days ago while GitHub was saying that they have been deployed successfully. If it is not because of timestamp desynchronization, could have they been deleted by mistake?

@saudet , I think you also have the credentials for deploying TF snapshots to the repository, are you aware of doing something that might have deleted those artifacts on your side?  

Here is the last job that was ran before the artifacts were missing: https://github.com/tensorflow/java/actions/runs/63013370

karllessard:
Ok so here is what I suggest then: We leave the deployments scripts as they are now and see if the problem occurs again. If it does not, then we classify this issue as an unresolved mystery and move on. If it does, then investigate more about it, maybe requesting some help from Sonatype. Wdyt?

@saudet do you still want to make adjustments to disable OS X GPU artifacts though?

saudet:
> Thanks for detailed explanation. Just for my understanding:
> 1. Existing publishing should work for maven since it keeps the
> snapshotVersion history. But gradle will fail to download it.
> 2. The re-deploy is purely workaround for gradle
>
> Am I right?

Yes, that's correct. Maven has always supported that, and yeah it's only a workaround for Gradle and others like sbt too, but it's simple and it works.

> If this works with Maven, then I would like to understand why we missed macosx artifacts in Nexus a few days ago while GitHub was saying that they have been deployed successfully. If it is not because of timestamp desynchronization, could have they been deleted by mistake?

It's not by mistake, it's just a setting of Nexus. Sonatype OSSRH configured their servers to leave only the 4 latest versions of snapshots, and anything else that is older than 2 days gets removed automatically (whenever the scheduled task actually runs though, so typically after 3 days):
https://issues.sonatype.org/browse/OSSRH-33316?focusedCommentId=422828&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-422828
Redeploying like that though prevents this from happening because Nexus considers them as 1, so it's also a workaround for that. 

BTW, GitHub Packages seems to be even worse in that regard: https://github.community/t5/GitHub-API-Development-and/GithHub-Packages-Older-artifact-downloaded-when-using-maven/m-p/43873

karllessard:
All right, thanks for the explanations @saudet, let's merge it then.

Issue #21, opened 2020-01-22
Protobuf support in TF Java


karllessard:
Right now, the TensorFlow Java client from the repository does not expose directly protobufs that are part of the contract of the C API _(Though [in this thread](https://github.com/tensorflow/community/pull/77#discussion_r265413948), it was mentioned that protobufs would be eventually removed from the API but I have a feeling this won't happen anytime soon)._

Instead, it compiles and distribute the protobufs Java bindings as [a separate artifact](https://mvnrepository.com/artifact/org.tensorflow/proto) and the client itself remains agnostic of the content of the messages and simply expose unmapped byte arrays, like [here](https://github.com/tensorflow/tensorflow/blob/b436796fb271612c10da67b4d88e8758f66c9935/tensorflow/java/src/main/java/org/tensorflow/SavedModelBundle.java#L115).

We need to decide how we want to handle those protobufs in the new distribution. Possible choices are:

* We keep distributing the compiled protobufs as an external artifact like now from the core repository build
* We keep distributing the compiled protobufs as an external artifact but from our new repository build
* We compile and distribute the Java protobuf bindings within the new TF Java client artifact so they get exposed directly by our API
* We compile the Java protobuf bindings with the new TF Java client but wrap them so they are not exposed directly by our API (in case they are removed in TF core, see previous thread link)
* Any other idea?

Note that if we decide to compile the protobufs from the new TF Java client, it will brings its load of additional dependencies, such as grpc.

CC: @sjamesr

saudet:
I think we should just add it pretty much as is as an additional module like `tensorflow-proto` under `tensorflow-java`. That would allow us to use it easily from other modules to create higher-level APIs using it, but still have it there for users until that happens. If/when the protocol buffers themselves are removed from the API, then we can think about deprecating it.

BTW, one major issue with protobuf on Java is that all versions use the same package name `com.google.protobuf`, but they are not compatible with each other, especially versions 2.x and 3.x, so we'll probably need to do some shading like this: https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml#L59-L111

karllessard:
[...]

PR #38, opened 2020-04-07
Attach protobuf generated classes to TF Core API

karllessard:
The main purpose of this PR is to generate Java classes from TensorFlow core protocol buffer descriptions and to distribute them directly from the `tensorflow-core-api` artifact. This way, the API can use concrete typed classes in the signature of some methods accepting TF protocol buffers instead of raw byte arrays.

The classes are generated from the Bazel build, using the same protobuf compiler used to build TensorFlow native library, which matches the version of the protobuf Maven dependency. The generated classes are then unarchived in the `/src/gen/java` folder of the `tensorflow-core-api` for building the Java library.

A few other changes were also applied as side-effects of this feature:

* Upgrade version of TensorFlow runtime library to 2.2.0-rc2
    * requires bazel = 2.0.0
    * this increase the size of this PR considerably but was required to catch up with new protos
* Patch protobuf definitions that were missing the `java_package` option
* Add a utility `java_api_import` to help select the right package/group for new operations
    * ok this is really just a little helper that gives hints on where maybe that operations has landed in the Python client, no rocket science here... but it is still very useful and for a moment I thought I lost it so I prefer to put it now in a safe place :) Don't spend too much time reviewing it.
* Classify new ops and move some debugging ops to their own group/package `debugging`
* Removed GRPC patch for Ubuntu (TF uses a new version of GRPC that shouldn't have this issue)
    * @dhruvrajan can you please validate this still compiles for you?

To see the impact of including the protobuf in the Java client, look at changes applied to the `src/main/java/org/tensorflow` classes (e.g. `Graph`, `Session`, `EagerSession`, `SavedModelBundle`, `Server`...)

Ref: #21


tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -14,12 +14,22 @@
...
   <dependencies>
     <dependency>
       <groupId>org.bytedeco</groupId>
       <artifactId>javacpp</artifactId>
       <version>${javacpp.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>${protobuf.version}</version>
```
saudet, line 31: Using the original protobuf artifact as is will cause issues with other libraries that need a different version of protobuf, so this should all be shaded. If the plan is to wait until users having this problem show up, is it going to be OK to break backward compatibility at that time?

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
```diff
@@ -0,0 +1,317 @@
+/* Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
saudet, line 1: Could this be translated to Java? And if not, why not?

Craigacp:
Do we want the protobufs in the core artifact? Might be nice if they were packaged separately as they are conceptually different from the core jar, plus they wander all over our package namespace. That said, packaging them separately would cause module system issues later.

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
karllessard: Like I mentioned in the description of this PR, this is just an old utility I've written awhile ago when classifying the 900+ initial operation wrappers and I just put it there so it is not lost, as I still find it quite useful.

Now could it be rewritten in Java... I guess so, but with a lot of work as it depends on TF internal classes. I'm not planning to spend time on this anytime soon so if we think it does not fit in our repo, then I'll simply remove it from this PR and continue to keep it in my fork.

karllessard:
> Do we want the protobufs in the core artifact? Might be nice if they were packaged separately as they are conceptually different from the core jar, plus they wander all over our package namespace. That said, packaging them separately would cause module system issues later.

The "problem" we have in our actual repository setup is that the TensorFlow runtime is built and is only accessible from the `tensorflow-core-api` artifact. So anything depending on TensorFlow must be compiled in this artifact. 

To generate protos as a separate module, we would need to move the JavaCPP build somewhere else (maybe in the `tensorflow-core` parent?) and have its childs cherry-picking what they need out of that build. I see other pieces of code that would make sense to be outside the core API too.

I don't know if it is possible to do this though and that might imply a lot of work. If we really want this, I suggest to open the discussion as a separate issue. cc: @saudet

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: So embedding it using Maven shade plugin? I can give it a try.

saudet: Not exactly, we need to use a patched version of protobuf that can generate code that depends on shaded classes, like this one: https://github.com/os72/protoc-jar-maven-plugin

But like I said it's not important to do that right away if we plan on breaking compatibility at some point in the future anyway.

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
saudet: Does this code actually gets used during the build? If so, I suppose we can leave it here...

saudet:
Well, let's think of it that way. Can anyone anywhere use those protobufs with something else than the C API? Or, from the perspective of an end user, they are essentially just part of the C API? If it's the latter, I don't see any reason why they should be placed in another artifact. Anyone using them will also want the C API. If it's the former, what are those use cases exactly?

tensorflow-core/tensorflow-core-api/pom.xml
karllessard: Oh, that might require more changes then because right now, I let [Bazel generating the Java protobuf classes](https://github.com/karllessard/tensorflow-java/blob/d7618c2b9be4b5f0d4942ef8adbd571e8aa6bc43/tensorflow-core/tensorflow-core-api/BUILD#L82) and just match the version used in Maven, as I did not find how to export only the `.proto` files out of TensorFlow.

I also found out [this](https://github.com/tensorflow/ecosystem/blob/1b3a0529bfa7e8eca6419cb5bef4d6f85b86c81e/spark/spark-tensorflow-connector/pom.xml#L109), for what it is worth. Looks like there is already some `org.tensorflow` artifacts that shades protobuf in a different way.

Craigacp:
It's more to lessen the load of anyone looking at the core API. That's where we expect people to look, it's what we'll document, and it's what the IDEs will show. If there's a bunch of protobufs in the namespace and we can't control what package names they have then it's going to mess with the structure and add a bunch of extra cognitive load when people look at it.

Though thinking about it, why can't we just change the package names in the Java protobufs? If the SIG is in charge of the Java API surely we should get to decide where they live.

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
karllessard: No, I only use it when I discover that new operations have been added after a TF upgrade, so I can classify them properly by configuring the [API definitions](https://github.com/karllessard/tensorflow-java/tree/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def) properly before rebuilding. It just gives me some hints in which package they should be located based on Python endpoints that has a similar name.

karllessard:
> It's more to lessen the load of anyone looking at the core API. That's where we expect people to look, it's what we'll document, and it's what the IDEs will show. 

Having them in a separate module will still pollute the package namespace and the IDE will show them the same way. For the source code, at least they reside in `/src/gen` and not in `/src/main`, so it is clear that they are not part of the maintained code.

> Though thinking about it, why can't we just change the package names in the Java protobufs? If the SIG is in charge of the Java API surely we should get to decide where they live.

The protobuf compiler `protoc` will [extract the package name](https://developers.google.com/protocol-buffers/docs/reference/java-generated#package) from the `.proto` files themselves. If we want to change these packages, then we need to create a PR for the TensorFlow main repo to update all of them, which will break backward compatibility with their actual 1.x proto artifact (though I'm not sure if they are ever planning to build it again)... or we can extend [this patch](https://github.com/karllessard/tensorflow-java/blob/proto/tensorflow-core/tensorflow-core-api/external/tensorflow-proto.patch) to override all of them on our side, which is quite a job and a bit ugly but still possible.

Craigacp:
I understand it involves changing files in the main TF repo, but that doesn't seem unreasonable if the change only affects Java.

And if I understand correctly it only breaks compatibility for people trying to use the 1.x artifacts with our new jars right? It'll still read the protos themselves just fine, but they'll come up with different names depending on what jar is loaded.

karllessard:
@Craigacp : if changes made to the existing protos in 1.15 are backward compatible, then yes, it should only be matter of finding them in a different location if we rename those packages.

So you would suggest to have them in something like this?

`org.tensorflow.proto.distruntime`
`org.tensorflow.proto.util`
`org.tensorflow.proto.framework`
`org.tensorflow.proto.example`
`org.tensorflow.proto.profiler`

I understand the point and I also thought of it before. At the same time, I'm not sure how useful it is for a user to know that a file has been generated out of a protobuf file or not. For example, in [this method](https://github.com/karllessard/tensorflow-java/blob/d7618c2b9be4b5f0d4942ef8adbd571e8aa6bc43/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java#L132), what would it provide to know that the `MetaGraphDef` is located in a `proto` package? BTW I'm not saying that we shouldn't do it but just challenging the idea.

tensorflow-core/tensorflow-core-api/pom.xml
saudet: Right, we can let users do it themselves when required...

saudet:
If the goal is to avoid confusion, then just a different package name is fine in my opinion. FWIW, I don't find these classes to be any more user friendly than the C API itself, that is I still see all that as part of the "core API" anyway.

BTW, if we can't change the package name with `protoc` on the command line, `sed` is more efficient option than modifying and maintaining manually a bunch of files. See how I'm translating JNI written by @Craigacp from C to C++ to make it work well with JavaCPP here:
https://github.com/bytedeco/javacpp-presets/blob/master/onnxruntime/cppbuild.sh#L57

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
saudet: Ah, ok, so let's add the step to compile it to the build.sh script or something while we're at it?

karllessard:
If we decide to rename the packages, I also think it make sense to apply the changes in the core repo since that is where the proto files reside but that won't probably be merged before TF2.3 and we can't wait for this release before having our protos available. 

So the temporary solution would be to apply this PR internally as a Bazel patch.

karllessard:
Now that being said, I kind of assume that Google will continue to add this `java_package` option in their new protos and classify them properly but that does not seemed to be enforced, as some protos were missing it.

Craigacp:
My reasoning for separating out the protos is because they aren't real code. We can't do anything about them, or change them in any way beyond repackaging, and having them sprinkled throughout the namespace disrupts what the packages mean. If a user knows that `MetaGraphDef` lives in a `org.tensorflow.proto.<something>` package then they immediately know it's part of the serialisation mechanism, and that it's fixed across Tensorflow languages. Otherwise they have to go and look at the class file.

hthu:
> My reasoning for separating out the protos is because they aren't real code. We can't do anything about them, or change them in any way beyond repackaging, and having them sprinkled throughout the namespace disrupts what the packages mean. If a user knows that `MetaGraphDef` lives in a `org.tensorflow.proto.<something>` package then they immediately know it's part of the serialisation mechanism, and that it's fixed across Tensorflow languages. Otherwise they have to go and look at the class file.

+1. I think having `org.tensorflow.*.proto` as the default prefix is going to make things clearer.
This can be done in two ways:
1. we can just put `.proto` suffix for existing packages.
2. Use `org.tensorflow.proto` as unique prefix for all generated protos.

tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc
karllessard: It's already there, see [here](https://github.com/karllessard/tensorflow-java/blob/d7618c2b9be4b5f0d4942ef8adbd571e8aa6bc43/tensorflow-core/tensorflow-core-api/build.sh#L37) and [here](https://github.com/karllessard/tensorflow-java/blob/d7618c2b9be4b5f0d4942ef8adbd571e8aa6bc43/tensorflow-core/tensorflow-core-api/BUILD#L62)

karllessard:
I guess we'll go with `org.tensorflow.proto.*`, it would look like the list I gave in [my previous post](https://github.com/tensorflow/java/pull/38#issuecomment-610737900). Ok, I'll check what is required to apply this change.

hthu:
> I guess we'll go with `org.tensorflow.proto.*`, it would look like the list I gave in [my previous post](https://github.com/tensorflow/java/pull/38#issuecomment-610737900). Ok, I'll check what is required to apply this change.

FYI, you can probably copy paste this under `tensorflow/core/framework` dir.
`grep -rl 'java_package = ' . | xargs sed -i 's#^\(.*\) java_package = \"org.tensorflow.\(.*\)\"#\1 java_package = \"org.tensorflow.proto.\2\"#g'`

karllessard:
@hthu , in your command, you are kind of assuming that the only protos we want are those under `/core/framework`. In this PR, I've included all of those that are part of [the protos_all rule](https://github.com/tensorflow/tensorflow/blob/c3e12bcd6d01f3bc92cbb99f3c91fc0bec3a8562/tensorflow/core/BUILD#L252).

I think it is ok (e.g. I'm pretty sure we want the protos under `/core/example` as well) but I just wanted to double check with you if there is any reason we shouldn't do this, thanks.

karllessard:
Ok, all protos are under the `org.tensorflow.proto` package now

hthu:
> @hthu , in your command, you are kind of assuming that the only protos we want are those under `/core/framework`. In this PR, I've included all of those that are part of [the protos_all rule](https://github.com/tensorflow/tensorflow/blob/c3e12bcd6d01f3bc92cbb99f3c91fc0bec3a8562/tensorflow/core/BUILD#L252).
> 
> I think it is ok (e.g. I'm pretty sure we want the protos under `/core/example` as well) but I just wanted to double check with you if there is any reason we shouldn't do this, thanks.

You're right :)
They should be all replaced.

BTW, some proto files don't have java_package option.

karllessard:
> BTW, some proto files don't have java_package option.

@hthu : for those I have [a patch](https://github.com/karllessard/tensorflow-java/blob/proto/tensorflow-core/tensorflow-core-api/external/tensorflow-proto.patch) that adds the missing options before the command that fixes the `.proto` package is executed (hence why this package is not present in the patch). I guessed in which subpackage the protos should be located.

Now I probably should create a PR for the main TF repo that applies all these changes but at least, we are not blocked for doing it on our side. (BTW, I see that you are part of Google Brain team, will that PR ends up in your hands?)

hthu:
> > BTW, some proto files don't have java_package option.
> 
> @hthu : for those I have [a patch](https://github.com/karllessard/tensorflow-java/blob/proto/tensorflow-core/tensorflow-core-api/external/tensorflow-proto.patch) that adds the missing options before the command that fixes the `.proto` package is executed (hence why this package is not present in the patch). I guessed in which subpackage the protos should be located.
> 
> Now I probably should create a PR for the main TF repo that applies all these changes but at least, we are not blocked for doing it on our side. (BTW, I see that you are part of Google Brain team, will that PR ends up in your hands?)

I don't know.
I'm following this as I'm interested :)

karllessard:
Ok, I think that’s ready to be merged, thanks everyone

PR #42, opened 2020-04-20
Replace EagerSession.NativeReference with JavaCPP Pointer

saudet:
With this, we're effectively managing the scope of `Tensor` with reference counting and `PointerScope`. There's still some manual calls to `retainReference()` and `releaseReference` that could be replaced with local `PointerScope`, but it's not too bad the way it is, so I'm opting for less refactoring, for now.

Also upgrade JavaCPP to 1.5.3 and TensorFlow to 2.2.0-rc3


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
```diff
@@ -52,7 +50,8 @@
     this.session = session;
     this.type = type;
     this.name = name;
-    this.nativeRef = new NativeReference(session, this, opNativeHandle, outputNativeHandles);
+    session.attach(this.opHandle = opNativeHandle);
```
Craigacp, line 53: Can we avoid side effect assignment here? I'd much prefer two lines, one for the assignment and one for the `session.attach` call.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -3734,6 +3735,21 @@ public Print print(Operand<TString> input, Print.Options... options) {
...
+   * @param axis
+   * @param inputMins
+   * @param inputMaxes
+   * @return a new instance of QuantizedConcatV2
+   */
+  public <T extends TType, U extends TNumber> QuantizedConcatV2<T> quantizedConcatV2(
+      Iterable<Operand<T>> values, Operand<U> axis, Iterable<Operand<TFloat32>> inputMins,
+      Iterable<Operand<TFloat32>> inputMaxes) {
+    return QuantizedConcatV2.create(scope, values, axis, inputMins, inputMaxes);
+  }
```
karllessard, line 3751: That's a new op that needs to be reclassified to the right package. In fact, it should replace the actual `QuantizedConcat` op that is found under `org.tensorflow.op.quantization`. 

Here's how to do it:
* Duplicate the file [`api_def_QuantizedConcat.pbtxt`](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_QuantizedConcat.pbtxt) to `api_def_QuantizedConcatV2.pbtxt` in the same folder.
* Replace the `endpoint` in the old file by `visibility: SKIP`, like in [this one](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_PriorityQueue.pbtxt)
* Replace the `graph_op_name` in the new file by `QuantizedConcatV2`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
```diff
@@ -60,23 +58,28 @@
     this.session = session;
     this.type = type;
     this.name = name;
-    this.nativeRef = new NativeReference(session, this, allocate(session.nativeHandle(), type));
+    this.opHandle = allocate(session.nativeHandle(), type);
+    session.attach(opHandle);
+    opHandle.releaseReference();
```
karllessard, line 63: If I recall correctly what @EronWright was envisioning for the reference counts is that it could be possible for the `session` to "grab" the ownership of the `opHandle` when it gets attached to it, instead of explicitly releasing it like in few places in this PR. 

I guess it is done like this because in JavaCPP, you need to explicitly retain a reference so its count is `1`, instead of having an initial value of `1`. Would you mind explaining again why it needs to be like this?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
```diff
@@ -60,23 +58,28 @@
...
   public EagerOperation build() {
-    TFE_TensorHandle[] tensorHandles = execute(nativeRef.opHandle);
+    TFE_TensorHandle[] tensorHandles = execute(opHandle);
     EagerOperation operation =
-        new EagerOperation(session, nativeRef.opHandle, tensorHandles, type, name);
+        new EagerOperation(session, opHandle, tensorHandles, type, name);
     // Release our reference to the native op handle now that we transferred its
     // ownership to the EagerOperation
-    nativeRef.clear();
+    session.detach(opHandle);
```
karllessard, line 73: In `EagerOperation` constructor, we attach the `opHandle` to the session, which was already attached to it in the constructor of this class, before we detach it here. So if I understand correctly, attaching more than once the same resource to an `EagerSession` (therefore to a `PointerScope`) requires to detach it the same amount of time?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
```diff
@@ -364,174 +353,38 @@ ResourceCleanupStrategy resourceCleanupStrategy() {
...
-              }
-            });
-      } catch (Exception e) {
-        cleanupInBackground = false;
-        throw e;
+  void detach(Pointer... resources) {
+    checkSession();
+    for (Pointer r : resources) {
+      if (resourceCleanupStrategy == ResourceCleanupStrategy.ON_SAFE_POINTS) {
+          nativeRefs.attach(r);
```
karllessard, line 367: What do we need both `nativeRefs` and `nativeResources`? If I understand correctly, `nativeRefs` are detached resources that are cleared only on safe points?

The purpose of safe points was mainly to allow releasing resources while a session is active but without having a separate thread for garbage-collection. Is the GC thread in JavaCPP is always running? If it cannot be disabled, then we don't need to support safe-points. If it can be disabled, can we link this option with the picked-up strategy then?

karllessard:
Also I've notice that the non-GPU Linux build failed, this seems to be a regression, can you please double-check?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
saudet: Could you push that to my branch? Or I'll just leave those changes out and you can do it before or after we merge this PR?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
saudet: When we're inside a scope (as in `PointerScope`), the count does get initialized to 1 by default. But when we're outside of any scope, it stays to 0, as a flag to indicate that reference counting is disabled. We need to enable it manually. If a `Pointer` would own itself automatically, then the count would start at `2` when initialized inside a scope, and so users would need to release them manually, even inside of a scope. It would make the use of scope pointless.

The idea is to do everything with `PointerScope`, then we don't need to call anything else, and everything is done behind the scenes, since the scopes own all the `Pointer` objects. It works exactly like RAII in C++. If you would like to understand better the motivation behind this, please read up on C++ and RAII!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
saudet: Yes, that's how reference counting works.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
saudet: We can remove those options. I would vote for this. I wasn't sure myself what was their purpose...

JavaCPP starts a thread for GC by default, yes, but it should be considered more or less like a safety net. It can be disabled with a system property, but it's pretty much for debug purposes only:
http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/Pointer.html#referenceQueue
I haven't encountered a user who wanted to have it disabled as a "feature", yet anyway.

saudet:
> Also I've notice that the non-GPU Linux build failed, this seems to be a regression, can you please double-check?

Follow my epic thread about this in the community forum: 
https://github.community/t5/GitHub-Actions/Diminishing-disk-space-on-Ubuntu-images/m-p/54376

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
Craigacp: Why do the queues in JavaCPP call `System.gc()`? It's not guaranteed that that method does anything, and what it does is probably GC algorithm dependent. Does this work across all the GCs in Hotspot?

saudet: Like I said, it's a safety net, precisely because Java does not provide any guarantees that anything gets done at all, it shouldn't be relied upon.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
karllessard: Yeah I can do that

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
karllessard: Then for the sake of simplicity, I opt to remove these options as well, the timing is still good for doing such changes. Anyway we can't disable the GC thread only at the eager session level anymore so they don't make much sense to have them there, if they have to be restored, it would be in a different place.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
karllessard: Then why not working to integrate a bit better the `PointerScope` in this PR? How would that work, the `EagerSession` needs to extend from it?

saudet: No, we just need to change a few things to be able to pass scopes around a bit more. Sure, let's do it now while we're at it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
saudet: "Safety net" isn't the right analogy. System.gc() is actually a last-ditch effort, but it usually helps avoid OutOfMemoryError, so that's why it's there.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
saudet: Ok, it's done!

karllessard:
@saudet Please don't forget to merge [my PR](https://github.com/saudet/tensorflow-java/pull/1) to your branch to handle properly the new ops and protos

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
```diff
@@ -277,22 +257,18 @@ private static void requireTensorHandle(TFE_TensorHandle handle) {
     }
   }
 
-  private static TFE_Op allocate(TFE_Context ctxHandle, String type) {
+  private static TFE_Op allocate(TFE_Context ctxHandle, String type, PointerScope outputScope) {
```
karllessard, line 260: Can you please elaborate more on how `PointerScope` is working? 

For instance, the scope define in this method is not passed anywhere so I assume that the pointer created by `TFE_Op.newOp` is automatically attached to it on creation? Then, since we are already in the scope of another `PointerScope` (the one passed as `outputScope`), would it be automatically attached to it as well if we were not declaring a new one here (and shouldn't that be enough)? 

If it is not required, I think it would make the code easier to follow not having so many scopes declared during the process.

saudet:
Ok, didn't see it because... GitHub? Anyway you can just push those kinds of changes directly to my branch.

karllessard:
Ok looks good, I just left [a comment](https://github.com/tensorflow/java/pull/42#discussion_r416304587) in there (for my own understanding I guess?). Other than that, we are ready to merge.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
karllessard: In other words, instead of :
```java
void create() {
    allocate(scope);
  }
}

private void allocate(PointerScope outputScope) {
    Pointer p = new Pointer();
    outputScope.attach(p);
  }
}
```
, can't we simply do this?
```java
void create() {
    allocate(scope);
  }
}

private void allocate(PointerScope outputScope) {
  return new Pointer();
}
```

saudet: > Can you please elaborate more on how `PointerScope` is working?

- http://bytedeco.org/news/2018/07/17/bytedeco-as-distribution/
- https://github.com/karllessard/tensorflow-java-sandbox/issues/3

> For instance, the scope define in this method is not passed anywhere so I assume that the pointer created by `TFE_Op.newOp` is automatically attached to it on creation? Then, since we are already in the scope of another `PointerScope` (the one passed as `outputScope`), would it be automatically attached to it as well if we were not declaring a new one here (and shouldn't that be enough)?

Yes, that's the general idea. We need the new scopes to clean up any local resources that we don't want lying around.

> If it is not required, I think it would make the code easier to follow not having so many scopes declared during the process.

Surely, but I don't see an easier way of doing this, short of doing reference counting manually. Would you like to go that way? :) FWIW, to make it prettier, this pattern should really be integrated into the Java language. There's still most likely some more that can be done at the library level. I'll be updating JavaCPP with any ideas that help, for sure, but I'm not having any more for now...

saudet: > can't we simply do this?

Yes, but this is not what we have. We've got at least the `status` object to deallocate here.

saudet: I'm refactoring this a bit to get rid of those temporary scopes by using the scopes already available in the session objects. I'll push the changes soon.

karllessard: But the status will be deallocated ultimately by the outer scope as well, right?

I suggest then that we opt for simplicity and that we let the outer scopes release resources allocated by their private helper methods, which will probably end up being inlined anyway if we can them simple enough.

saudet: Sure, we can refactor some more and start inlining those private methods. We don't need to keep them separate now that we're not using writing JNI anymore. Anyway, let me know if it's good enough with the latest changes.

karllessard: Yep, looks good!

Issue #45, opened 2020-04-29
Operator with more than 8 outputs


roywei:
Hi, it seems the java api can't create an output more than 8. This is a big limitation especially on split operators. 
Here is a reproducible code, I'm using `tensorflow-core-api` latest SNAPSHOT version.
```
package ai.djl.tensorflow.engine;

import org.tensorflow.EagerSession;
import org.tensorflow.op.Ops;
import org.tensorflow.types.TInt64;

public class Tftest {
    public static void main(String[] args) {
        EagerSession eagerSession = EagerSession.options().async(true).build();
        Ops tf = Ops.create(eagerSession);
        // creates a (20, 20) tensor with zeros and split to 10 tensors equally on axis 0,
        // should return 10 tensors each with shape(2, 20).
        tf.splitV(
                tf.zeros(tf.constant(new long[]{20,20}), TInt64.DTYPE),
                tf.constant(10),
                tf.constant(0),
                10L
        );
    }
}
```

error message:
```
Exception in thread "main" java.lang.IllegalArgumentException: Expecting 10 outputs, but *num_retvals is 8
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:72)
	at org.tensorflow.EagerOperationBuilder.execute(EagerOperationBuilder.java:302)
	at org.tensorflow.EagerOperationBuilder.build(EagerOperationBuilder.java:68)
	at org.tensorflow.EagerOperationBuilder.build(EagerOperationBuilder.java:57)
	at org.tensorflow.op.core.SplitV.create(SplitV.java:64)
	at org.tensorflow.op.Ops.splitV(Ops.java:5715)
	at ai.djl.tensorflow.engine.Tftest.main(Tftest.java:11)
```

roywei:
The only possible cause may be [here](https://github.com/tensorflow/tensorflow/pull/28212/files#diff-d48204d45bb1a0c42ec898235026aa44R92), but this code is in the main TF repo, not in this repo.  I think API in this repo switched to JavaCPP instead of JNI, so it should not be affected? 

Currently I can workaround this by doing split multiple times each time with a maximum of 8 outputs. But it will affect performance as this operator will be used heavily when processing datasets.

Please help take a look, thank you so much!

cc @karllessard @saudet

saudet:
[...]

PR #46, opened 2020-04-29
increase max num of outputs

roywei:
fix https://github.com/tensorflow/java/issues/45

@karllessard @saudet @frankfliu

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
@@ -226,7 +226,7 @@ final class EagerOperationBuilder implements OperationBuilder {
   private final String name;
 
   /** This value should be >= to the maximum number of outputs in any op */
-  private static final int MAX_OUTPUTS_PER_OP = 8;
+  private static final int MAX_OUTPUTS_PER_OP = 1000;
 
   private static void requireOp(TFE_Op handle) {
     if (handle == null || handle.isNull()) {


roywei:
@googlebot I signed it!

karllessard:
Thanks @roywei , can you explain what is the use case of having 1000 outputs in a single op? Is this a custom op?

I agree that 8 is very conservative and should be increased but we should try to limit this number to something reasonable since it has impact on the amount of memory allocated by TF to [return the tensor handles.](https://github.com/tensorflow/java/blob/78f84c08bff9126f90cf6e123283f80450513d75/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java#L270)

saudet:
> I agree that 8 is very conservative and should be increased but we should try to limit this number to something reasonable since it has impact on the amount of memory allocated by TF to [return the tensor handles.](https://github.com/tensorflow/java/blob/78f84c08bff9126f90cf6e123283f80450513d75/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java#L270)

That's just 8 bytes for each pointer, so 8000 bytes for 1000 of them, not exactly a lot of memory.

roywei:
In one of [DJL's](https://djl.ai) NLP applications, we'd like to process a batch of data, sometime it require us to unbatch the data (split by batch size), apply the process (padding for example), and then batch them back together. Here is a [unit test ](https://github.com/awslabs/djl/blob/4353cb34e4f9ba4c281327ad3b887eaa61197648/integration/src/main/java/ai/djl/integration/tests/translate/PaddingStackBatchifierTest.java#L26)with small batch size. For actual NLP use cases, sometimes the actual batch size could be quite big.

So split a batch data (256, 224, 224) into 256 slices of (1,224,224) will return an iterable of size 256 and currently is limited by 8.

karllessard:
@roywei oh I see, an iterable output, that's right. Ok, let's try it out, thanks!

PR #54, opened 2020-05-13
Fix Windows build on GitHub Actions by increasing page file size

saudet:
Works for CPU builds, but not CUDA builds, which take too much time:
https://github.com/saudet/tensorflow-java/runs/669792547
Well, it's already better than nothing for now :)


karllessard:
Great! There is an issue actually with `tensorflow-framework` test that make all builds fail, for which I'm testing a fix [in this PR](https://github.com/tensorflow/java/pull/55) (currently building).

But out of curiosity, the Windows build still report a successful build even if it failed running those tests and the maven build was aborted. Do you know why we always receive a green light for Windows build while they should have been red?

saudet:
It's something I left unfinished on purpose because the Windows builds were not passing anyway, so that it wouldn't fail the other builds. Now would be a good time to add that. I've pushed a commit with a check, and it should now be able to report failures.

PR #55, opened 2020-05-13
Include native artifacts for unit testing

karllessard:


karllessard:
Strange, that fix worked locally but CI still have the same issue: `no jnitensorflow in java.library.path`

@saudet , any tip how we can fix this? including `tensorflow-core-platform` in the `test` scope should allow us to run unit tests using tensorflow, no?

CC: @dhruvrajan

saudet:
Yeah, that gets a bit tricky. We need to tweak this a bit to make it work correctly.
I've added the proper changes to pull #54.

karllessard:
Changed were pushed in #54

PR #63, opened 2020-05-27
Recycle objects when iterating an element sequence

karllessard:
The main addition of this PR is the creation on a `DataBufferWindow` interface, that allows a user to view partially the content of a given buffer and to move (or "slide") it across the memory segment so that the same window instance can reused to present different chunk of data.

This is particularly useful when visiting all values (or elements of a given dimension) of a `NdArray` in a given sequence (e.g. `ndArray.elements(0)` or `ndArray.scalars()`) and performance gains are substantial for large tensors (in the order of 5x faster).

### DIFF
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
@@ -248,6 +248,38 @@ public interface DataBuffer<T> {
   DataBuffer<T> slice(long index, long size);
 
   /**
+   * Creates a {@link DataBufferWindow} that provides a partial view of this buffer.
+   *
+   * <p>The created window has a fixed size and can be {@link DataBufferWindow#slideTo(long) "slide"}
+   * across this buffer at different offsets to provide different views of the data without allocating
+   * a new buffer instance, like {@link #offset(long)} does. This improves overall performances when
+   * this operation is repeated frequently. For example:
+   *
+   * <pre>{@code
+   * IntDataBuffer bufferA = DataBuffers.ofInts(1024);
+   * // ... init buffer data
+   * IntDataBuffer bufferB = DataBuffers.ofInts(1, 2, 3, 4);
+   *
+   * // Return the index of the first occurrence of bufferB in bufferA using a sliding window
+   * DataBufferWindow<IntDataBuffer> windowA = bufferA.window(4);
+   * for (int i = 0; i < bufferA.size() - bufferB.size(); ++i) {
+   *     if (windowA.slideTo(i).buffer().equals(bufferB)) {
+   *         return i;
+   *     }
+   * }
+   * }</pre>
+   *
+   * <p>The returned object is stateful and is not thread-safe.
+   *
+   * @param size size of the window
+   * @return a new window that starts at the index 0 of this buffer,
+   *         or null if this type of buffer does not support buffer windows
+   */
+  default DataBufferWindow<? extends DataBuffer<T>> window(long size) {
+    return null;
+  }
+
+  /**
    * Visits the backing storage of this buffer.
    *
    * <p>The buffer implementation is responsible of passing back a reference to the actual data
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
@@ -250,10 +250,10 @@ public interface DataBuffer<T> {
   /**
    * Creates a {@link DataBufferWindow} that provides a partial view of this buffer.
    *
-   * <p>The created window has a fixed size and can be {@link DataBufferWindow#slideTo(long) "slide"}
-   * across this buffer at different offsets to provide different views of the data without allocating
-   * a new buffer instance, like {@link #offset(long)} does. This improves overall performances when
-   * this operation is repeated frequently. For example:
+   * <p>The created window has a fixed size and can {@link DataBufferWindow#slide(long) "slide"}
+   * along this buffer to provide different views of the data without allocating a new buffer
+   * instance, like {@link #offset(long)} does. This improves overall performance when this
+   * operation is repeated frequently. For example:
    *
    * <pre>{@code
    * IntDataBuffer bufferA = DataBuffers.ofInts(1024);
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
@@ -272,11 +272,11 @@ public interface DataBuffer<T> {
    * <p>The returned object is stateful and is not thread-safe.
    *
    * @param size size of the window
-   * @return a new window that starts at the index 0 of this buffer,
-   *         or null if this type of buffer does not support buffer windows
+   * @return a new window that starts at the index 0 of this buffer
+   * @throws UnsupportedOperationException if this type of buffer does not support buffer windows
    */
   default DataBufferWindow<? extends DataBuffer<T>> window(long size) {
-    return null;
+    throw new UnsupportedOperationException();
   }
 
   /**
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
@@ -1,7 +1,7 @@
 package org.tensorflow.tools.buffer;
 
 /**
- * A mutable container for view part of a {@link DataBuffer}.
+ * A mutable container for viewing part of a {@link DataBuffer}.
  *
  * <p>Data buffer windows have a fixed size and can {@link DataBufferWindow#slide(long) "slide"}
  * along a buffer to provide different views of the data without allocating a new buffer instance,
@@ -29,7 +29,7 @@ package org.tensorflow.tools.buffer;
 public interface DataBufferWindow<B extends DataBuffer<?>> {
 
   /**
-   * Returns the current offset of this window in the underlying buffer.
+   * Returns the current offset of this window in the original buffer.
    */
   long offset();
 
@@ -39,20 +39,21 @@ public interface DataBufferWindow<B extends DataBuffer<?>> {
   long size();
 
   /**
-   * Moves the window at the given position in the underlying buffer.
+   * Moves the window at the given position in the original buffer.
    *
    * <p>The size of the window remains the same and its offset is set to {@code index}, so that
    * accessing the value of {@link #buffer()} at index {@code x} will return the value at
-   * {@code index + x} in the underlying buffer.
+   * {@code index + x} in the original buffer.
    *
    * @param index new offset for this window
    * @return this instance
-   * @throws IndexOutOfBoundsException if the window cannot be slid because it goes beyond buffer limits
+   * @throws IndexOutOfBoundsException if the window cannot be slid because it goes beyond
+   *                                   the original buffer limits
    */
   DataBufferWindow<B> slideTo(long index);
 
   /**
-   * Moves the window of {@code step} elements in the underlying buffer.
+   * Moves the window of {@code step} elements in the original buffer.
    *
    * <p>The size of the window remains the same and its offset is set to {@code offset() + step}.
    * If {@code step} is positive, then the window will slide forward. If it is negative, it will
@@ -60,15 +61,17 @@ public interface DataBufferWindow<B extends DataBuffer<?>> {
    *
    * @param step value to add to the current offset of this window
    * @return this instance
-   * @throws IndexOutOfBoundsException if the window cannot be slid because it goes beyond buffer limits
+   * @throws IndexOutOfBoundsException if the window cannot be slid because it goes beyond
+   *                                   the original buffer limits
    */
   DataBufferWindow<B> slide(long step);
 
   /**
-   * Returns a buffer presenting the data currently viewed by this window.
+   * Returns the buffer backing this window.
    *
-   * <p>The same instance is always returned for a given window, even after sliding it at
-   * different offsets in the underlying buffer. For example:
+   * <p>Each window instance has it's own buffer providing a view onto the original
+   * {@link DataBuffer}. The buffers are mutated when the window slides to different offsets.
+   * For example:
    *
    * <pre>{@code
    * IntDataBuffer buffer = DataBuffers.of(0, 1, 2, 3);
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
@@ -1,0 +1,90 @@
+package org.tensorflow.tools.buffer;
+
+/**
+ * A mutable container for viewing partially a {@link DataBuffer}.
+ *
+ * <p>Data buffer windows have a fixed size and can be {@link DataBufferWindow#slideTo(long) "slide"}
+ * across a buffer at different offsets to provide different views of the data without allocating
+ * a new buffer instance, like {@link #offset(long)} does. This improves overall performances when
+ * this operation is repeated frequently. For example:
+ *
+ * <pre>{@code
+ * IntDataBuffer bufferA = DataBuffers.ofInts(1024);
+ * // ... init buffer data
+ * IntDataBuffer bufferB = DataBuffers.ofInts(1, 2, 3, 4);
+ *
+ * // Return the index of the first occurrence of bufferB in bufferA using a sliding window
+ * DataBufferWindow<IntDataBuffer> windowA = bufferA.window(4);
+ * for (int i = 0; i < bufferA.size() - bufferB.size(); ++i) {
+ *     if (windowA.slideTo(i).buffer().equals(bufferB)) {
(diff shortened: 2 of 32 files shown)


tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/CoordinatesIncrementor.java
```diff
@@ -0,0 +1,42 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *  =======================================================================
+ */
+
+package org.tensorflow.tools.ndarray.impl.sequence;
+
+import java.util.NoSuchElementException;
+
+import org.tensorflow.tools.ndarray.impl.dimension.DimensionalSpace;
```
Craigacp, line 22: Unnecessary imports?

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DoubleDataBuffer.java
```diff
@@ -157,4 +157,9 @@ default DoubleDataBuffer narrow(long size) {
 
   @Override
   DoubleDataBuffer slice(long index, long size);
+
+  @Override
+  default DataBufferWindow<DoubleDataBuffer> window(long size) {
+    return null;
```
Craigacp, line 163: What's with all the default methods returning null? I didn't see null checks elsewhere in the code to deal with it not being overridden?

karllessard: [here](https://github.com/tensorflow/java/blob/442d65d546db4704422e6f98c7b9c4172fecabef/tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/dense/AbstractDenseNdArray.java#L45)

I just implemented windows for raw buffers at the moment. At some point, they should have a window implementation but it is not mandatory.

Unfortunately in Java, when you override the signature of a method in a super interface, you don’t inherit anymore of its default implementation, so I had to return null explicitly in all subinterfaces.

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/CoordinatesIncrementor.java
karllessard: Yup, thanks!

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DoubleDataBuffer.java
Craigacp: Ok, in which case the javadoc should probably note that the window method is optional on `DataBuffer`, similar to the remove methods on `List`. And maybe it should throw `UnsupportedOperationException`? It would be kinda ugly to have the method that uses it catch a RuntimeException though so maybe returning null is fine.

karllessard: Yeah, I was hesitating between the two... but probably after this PR I'll add windows to all other buffer types as well, so it will really be an exception if it is not supported (e.g. if the user provides one of its own buffer implementation). So maybe the exception is better, I'll change it.

karllessard:
Updated, please let me know if there is anything else @Craigacp , thanks!

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/FastElementSequence.java
```diff
@@ -0,0 +1,75 @@
+/*
+ *  Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 2: Copyright year 2020?

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/SingleElementSequence.java
```diff
@@ -24,7 +24,11 @@
 import org.tensorflow.tools.ndarray.NdArraySequence;
 import org.tensorflow.tools.ndarray.impl.AbstractNdArray;
 
-class SingleElementSequence<T, U extends NdArray<T>> implements NdArraySequence<U> {
+public final class SingleElementSequence<T, U extends NdArray<T>> implements NdArraySequence<U> {
```
Craigacp, line 27: Class level javadoc, now it's public? Or is this something we'd exclude using modules if we were on Java 9?

tensorflow-tools/src/test/java/org/tensorflow/tools/ndarray/impl/sequence/ElementSequenceTest.java
```diff
@@ -37,8 +41,7 @@ public void iterateVectorsWithIndex() {
     IntNdArray array = NdArrays.ofInts(Shape.of(2, 3, 2));
 
     @SuppressWarnings("unchecked")
-    NdArraySequence<IntNdArray> sequence = ElementSequence
-        .create((AbstractNdArray<Integer, IntNdArray>)array, 1);
+    NdArraySequence<IntNdArray> sequence = new SlicingElementSequence((AbstractNdArray<Integer, IntNdArray>)array, 1);
```
Craigacp, line 44: Do we still need to suppress warnings here? It looks like it's all properly typed?

tensorflow-tools/src/test/java/org/tensorflow/tools/ndarray/impl/sequence/ElementSequenceTest.java
```diff
@@ -56,7 +59,7 @@ public void iterateScalarsWithIndex() {
     IntNdArray array = NdArrays.ofInts(Shape.of(2, 3, 2));
 
     @SuppressWarnings("unchecked")
-    NdArraySequence<IntNdArray> cursor = ElementSequence.create((AbstractNdArray<Integer, IntNdArray>)array, 2);
+    NdArraySequence<IntNdArray> cursor = new SlicingElementSequence((AbstractNdArray<Integer, IntNdArray>)array, 2);
```
Craigacp, line 62: SuppressWarnings?

tensorflow-tools/src/test/java/org/tensorflow/tools/ndarray/impl/sequence/ElementSequenceTest.java
```diff
@@ -74,4 +77,71 @@ public void iterateScalarsWithIndex() {
     assertArrayEquals(new long[] {1, 2, 0}, coords.get(10));
     assertArrayEquals(new long[] {1, 2, 1}, coords.get(11));
   }
+
+  @Test
+  public void slicingElementSequenceReturnsUniqueInstances() {
+    IntNdArray array = NdArrays.ofInts(Shape.of(2, 3, 2));
+    @SuppressWarnings("unchecked")
```
Craigacp, line 84: SuppressWarnings?

tensorflow-tools/src/test/java/org/tensorflow/tools/ndarray/impl/sequence/ElementSequenceTest.java
```diff
@@ -74,4 +77,71 @@ public void iterateScalarsWithIndex() {
...
+      });
+      elements.add(e);
+    });
+  }
+
+  @Test
+  public void fastElementSequenceReturnsSameInstance() {
+    IntNdArray array = NdArrays.ofInts(Shape.of(2, 3, 2));
+    IntNdArray element = array.get(0);
+    @SuppressWarnings("unchecked")
```
Craigacp, line 102: SuppressWarnings?

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/raw/UnsafeMemoryHandle.java
```diff
@@ -138,39 +142,47 @@ int arrayOffset(Class<?> arrayClass) {
...
 
   DoubleBuffer toArrayDoubleBuffer() {
-    return DoubleBuffer.wrap((double[])object, (int)((byteOffset - UnsafeReference.UNSAFE.arrayBaseOffset(double[].class)) / scale), (int)size());
+    return DoubleBuffer.wrap((double[])object, (int)((byteOffset - UnsafeReference.UNSAFE.arrayBaseOffset(double[].class)) / scale), (int)size);
   }
 
   final Object object;
-  final long byteOffset;
+  final long baseOffset;
+  long byteOffset;
```
Craigacp, line 170: This is the crucial change right? Making this class no longer completely immutable so it can walk down an allocated block of memory.

karllessard:
Many thanks @Craigacp for your review and your patience with my poor English :) Going through your comments...

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBuffer.java
```diff
@@ -247,6 +247,38 @@
    */
   DataBuffer<T> slice(long index, long size);
 
+  /**
+   * Creates a {@link DataBufferWindow} that provides a partial view of this buffer.
+   *
+   * <p>The created window has a fixed size and can be {@link DataBufferWindow#slideTo(long) "slide"}
```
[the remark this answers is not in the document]
karllessard: I've picked "and can slide along"

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * <p>The size of the window remains the same and its offset is reset to {@code offset() + step}.
+   * If {@code step} is positive, then the window will slide forward. If it is negative, it will
+   * slide backward.
+   *
+   * @param step value to add to the current offset of this window
+   * @return this instance
+   * @throws IndexOutOfBoundsException if the resulting offset goes beyond the underlying buffer limits
+   */
+  DataBufferWindow<B> slideOf(long step);
```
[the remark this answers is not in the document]
karllessard: ...or just `slide`

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Moves the window of {@code step} elements in the underlying buffer.
+   *
+   * <p>The size of the window remains the same and its offset is reset to {@code offset() + step}.
+   * If {@code step} is positive, then the window will slide forward. If it is negative, it will
+   * slide backward.
+   *
+   * @param step value to add to the current offset of this window
+   * @return this instance
+   * @throws IndexOutOfBoundsException if the resulting offset goes beyond the underlying buffer limits
```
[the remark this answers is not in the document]
karllessard: "@throws IndexOutOfBoundsException if the window cannot be slid because it goes beyond buffer limits"

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  DataBufferWindow<B> slideOf(long step);
+
+  /**
+   * Returns a buffer presenting the data currently viewed by this window.
```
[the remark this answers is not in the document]
karllessard: No, this is the "window buffer", which is mutable and just shows a portion of the "underlying buffer". The "underlying buffer" is the buffer on which `.window(size)` has been invoked.  

I'm not a fan of all this terminology neither, so if you have better suggestions...

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/dense/AbstractDenseNdArray.java
```diff
@@ -17,16 +17,40 @@
...
+    if (rank() == 0 && dimensionIdx < 0) {
+      return new SingleElementSequence<>(this);
+    }
+    DimensionalSpace elemDims = dimensions().from(dimensionIdx + 1);
+    try {
+      DataBufferWindow<? extends DataBuffer<T>> elemWindow = buffer().window(elemDims.physicalSize());
+      U element = instantiate(elemWindow.buffer(), elemDims);
+      return new FastElementSequence(this, dimensionIdx, element, elemWindow);
+    } catch (UnsupportedOperationException e) {
+      // If buffer windows are not supported, fallback to slicing (and slower) sequence
```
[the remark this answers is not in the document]
karllessard: I thought of it too but the funny part is that we don't have any log yet in TF Java! So adding this one comes up with picking the right strategy to output logs from our library, which can be a story in itself.

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/dense/AbstractDenseNdArray.java
```diff
@@ -112,21 +136,16 @@ protected AbstractDenseNdArray(DimensionalSpace dimensions) {
     super(dimensions);
   }
 
-  abstract protected DataBuffer<T> buffer();
+  abstract protected <P> DataBuffer<T> buffer();
```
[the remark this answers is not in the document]
karllessard: ouf, left over

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/FastElementSequence.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Iterator;
+import java.util.function.BiConsumer;
+
+import org.tensorflow.tools.buffer.DataBufferWindow;
+import org.tensorflow.tools.ndarray.NdArray;
+import org.tensorflow.tools.ndarray.NdArraySequence;
+import org.tensorflow.tools.ndarray.impl.AbstractNdArray;
+
+public final class FastElementSequence<T, U extends NdArray<T>> implements NdArraySequence<U> {
```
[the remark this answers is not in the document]
karllessard: This is an internal class (under `impl` package) and a lot of them are not documented. I can add something for this one now but probably it would be worth adding doc to all other internal classes as well.

tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/impl/sequence/SingleElementSequence.java
karllessard: yes that would be something we would exclude in Java 9, pretty much everything under `impl` in fact should be limited to internal usage only. Still, some javadoc wouldn't hurt.

tensorflow-tools/src/test/java/org/tensorflow/tools/ndarray/impl/sequence/ElementSequenceTest.java
karllessard: good catch!

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/raw/UnsafeMemoryHandle.java
karllessard: It is... but the whole point of this PR is to have somewhere a mutable object that can be recycled. Note that this class is package-scoped, so nothing outside this internal package can change the value of it other that the endpoints we expose (which is `window.slide()` in this case).

Still, do you think that might be worth having two version of this object, one mutable and one not?

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
```diff
@@ -1,12 +1,12 @@
 package org.tensorflow.tools.buffer;
 
 /**
- * A mutable container for viewing partially a {@link DataBuffer}.
+ * A mutable container for view part of a {@link DataBuffer}.
```
Craigacp, line 4: "for viewing"

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/impl/raw/UnsafeMemoryHandle.java
Craigacp: It's fine as is, I was just checking that I'd understood the PR properly. I don't think it will make too much difference performance wise having a mutable field, unless it's on a JVM with more aggressive escape analysis like Graal (where it might be able to elide the allocation, but it might be able to do that for the mutable version too).

tensorflow-tools/src/main/java/org/tensorflow/tools/buffer/DataBufferWindow.java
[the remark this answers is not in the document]
Craigacp: Maybe "Returns the buffer backing this window. Each window instance has it's own buffer providing a view onto the original `DataBuffer`. The buffers are mutated when the window slides to different offsets." Or something along those lines? I like the notion of "backing" as I feel that's a fairly common usage for this kind of operation.

Craigacp:
> Many thanks @Craigacp for your review and your patience with my poor English :) Going through your comments...

No worries, happy to help. It's all minor stuff, it's in good shape.

The reason I asked for a little javadoc on the impl classes is for our benefit in the future, plus as we start to collaborate on an ndarray library with other groups we'll all need to be able to understand each other's code.

PR #64, opened 2020-05-28
Suggestions for minor improvements to Shape.java

chenkelmann:
* Added some utility methods to query the shape for its properties.
* Allowed negative indexing for Shape.size(), that way it is easier and better readable to get the size of the last dimension or second-to-last dimension etc.
* Add missing javadoc and new javadoc for the new methods.
* Added unit tests for the new methods.

I hope this doesn't make the API too bloated, I just wished Shape classes had these methods in other frameworks I have used, maybe they are useful for other users as well.

### DIFF
--- a/tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
+++ b/tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
@@ -19,16 +19,29 @@ package org.tensorflow.tools;
 
 import java.util.Arrays;
 
+/**
+ * The shape of a Tensor or {@link org.tensorflow.tools.ndarray.NdArray}. Defines the size along
+ * its axis. May contain an unknown size for one of the axis or may be totally unknown, in which
+ * case not even the number of axis is known.
+ */
 public final class Shape {
 
+  /**
+   * Size reported for shapes for which the total size cannot be computed, either because their
+   * number of dimensions is unknown or one or more of their dimensions have an unknown size.
+   */
   public static long UNKNOWN_SIZE = -1L;
 
-  /** Create a Shape representing an unknown number of dimensions. */
+  /** Create a Shape representing an unknown number of dimensions.
+   * @return a Shape for which {@link Shape#isUnknown()} is true.
+   */
   public static Shape unknown() {
     return new Shape(null);
   }
 
-  /** Create a Shape representing a scalar value. */
+  /** Create a Shape representing a scalar value.
+   * @return a Shape without dimensions for which {@link Shape#isScalar()} is true.
+   */
   public static Shape scalar() {
     return new Shape(new long[0]);
   }
@@ -61,6 +74,14 @@ public final class Shape {
     return new Shape(dimensionSizes);
   }
 
+  /**
+   * Returns the total number of elements a Tensor with this shape would have. If this
+   * {@link Shape#isUnknown()} is true or {@link Shape#hasUnknownDimension()}  is true,
+   * {@link Shape#UNKNOWN_SIZE} is returned.
+   *
+   * @return the total number of elements a Tensor with this shape would have if it can be
+   * calculated, else {@link Shape#UNKNOWN_SIZE}.
+   */
   public long size() {
     if (size == null) {
       size = computeSize(dimensionSizes);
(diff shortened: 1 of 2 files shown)


chenkelmann:
@googlebot I signed it!

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -19,16 +19,29 @@
 
 import java.util.Arrays;
 
+/**
+ * The shape of a Tensor or {@link org.tensorflow.tools.ndarray.NdArray}. Defines the size along
+ * its axis. May contain an unknown size for one of the axis or may be totally unknown, in which
+ * case not even the number of axis is known.
+ */
```
karllessard, line 26: Funny, I never realized that `Shape` was missing that much documentation... thanks!

Can you please reformat it to follow [Google Java Style guide](https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs)? "Defines the.." should start on a separate line, preceded by a blank line plus a `<p>` tag, like [here](https://github.com/tensorflow/java/blob/ef0b4f26254cbbfd29b39e184962706c2d348e59/tensorflow-tools/src/main/java/org/tensorflow/tools/ndarray/NdArray.java#L26). This should be applied to all other javadoc blocks as well.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -19,16 +19,29 @@
...
+  /**
+   * Size reported for shapes for which the total size cannot be computed, either because their
+   * number of dimensions is unknown or one or more of their dimensions have an unknown size.
+   */
   public static long UNKNOWN_SIZE = -1L;
 
-  /** Create a Shape representing an unknown number of dimensions. */
+  /** Create a Shape representing an unknown number of dimensions.
+   * @return a Shape for which {@link Shape#isUnknown()} is true.
+   */
```
karllessard, line 37: Here, the [guide](https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line) suggest to enter a newline after `/**` (same thing, should be applied to all javadoc blocks added)

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -61,21 +74,57 @@ public static Shape of(long... dimensionSizes) {
...
+      return dimensionSizes[i];
+    } else {
+      return dimensionSizes[dimensionSizes.length + i];
+    }
   }
 
+  /**
+   * Returns the number of dimensions of this shape. -1 if unknown, 0 for a scalar, 1 for a vector,
+   * 2 for a matrix etc.
+   * @return the number of dimensions of this shape. -1 if unknown, 0 for a scalar, 1 for a vector,
```
karllessard, line 117: again according to the [guide](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment), you should omit the `@return` tag. Same for other block below.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -88,6 +137,38 @@ public boolean hasUnknownDimension() {
...
+
+  /**
+   * Returns whether the number of dimensions of this shape is unknown.
+   * @return whether the number of dimensions of this shape is unknown.
+   */
+  public boolean isUnknown() {
+    return dimensionSizes == null;
+  }
+
   public long[] asArray() {
```
karllessard, line 172: maybe add doc to this one as well? :) 

Might be worth mentioning that changing the values in the returned array will affect the actual shape (or even better, change the code to return a copy of the array, it might be less performant but it is safer).

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -97,6 +178,17 @@ public int hashCode() {
     return dimensionSizes != null ? Arrays.hashCode(dimensionSizes) : super.hashCode();
   }
 
+  /**
+   * Equals implementation for shapes. Two shapes are considered equal iff:
+   * <p><ul>
+   * <li>the number of dimensions is defined and equal for both
+   * <li>the size of each dimension is defined and eqaul for both
```
karllessard, line 185: typo in `eqaul`

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -97,6 +178,17 @@ public int hashCode() {
...
+   * </ul><p>
+   * If either shape has an unknown dimension (even if it is the same in both) or if either
+   * shape has an unknown number of dimensions (even if both return {@code true} for
+   * {@link Shape#isUnknown()}), they are not considered equal. However, a shape will always
+   * equal itself, event it is unknown or contains unknown dimensions.
```
karllessard, line 190: `event it is...` you meant `even if it is`?

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
```diff
@@ -148,6 +278,51 @@ public Shape prepend(long firstDimension) {
...
+   */
+  public Shape append(long lastDimension) {
+    long[] newDimensions = new long[dimensionSizes.length + 1];
+    newDimensions[newDimensions.length - 1] = lastDimension;
+    System.arraycopy(dimensionSizes, 0, newDimensions, 0, dimensionSizes.length);
+
+    return Shape.of(newDimensions);
+  }
+
+  /** Returns a new shape, with another shapes' dimensions prepended.
```
karllessard, line 294: `shape's`? same thing for `append(Shape)`

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: Sure! Sorry for not checking the styleguide first.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: Done!

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: I prefer that anyway, should have removed all of them now.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: When I had a look at this, I thought about changing it to return a defensive copy, as the way it was seem like an oversight. But it was used in many places, so I did not feel I confident I knew enough about the inner workings to make such a change. This is why I didn't comment it: I wasn't sure about the intention. As per your suggestion, I changed the method to return a defensive copy and added the appropriate comment & test.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: Fixed

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: Yup, that's what I meant. Fixed it.

tensorflow-tools/src/main/java/org/tensorflow/tools/Shape.java
chenkelmann: I Really should have proof-read before committing...

PR #71, opened 2020-06-14
Add ShapeOps Enhancement to work with org.tensorflow.op.core.Shape

JimClarke5:
Add ShapeOps class as a tool for working with org.tensorflow.op.core.Shape objects directly during runtime as opposed to working with static shapes via org.tensorflow.tools.Shape. This is needed in Java TensorFlow frameworks, such as Keras,  to manipulate shapes when the actual shape is not known until the graph is run.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
@@ -1,8 +1,17 @@
-/*
- * To change this license header, choose License Headers in Project Properties.
- * To change this template file, choose Tools | Templates
- * and open the template in the editor.
- */
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
 package org.tensorflow.op.core;
 
 import java.util.Arrays;
@@ -38,7 +47,7 @@ import org.tensorflow.types.family.TType;
 public abstract class ShapeOps {
 
     /**
-     * flatten the shape to 1 dimension
+     * Flatten the operand to 1 dimension
      *
      * @param <T> the type of operand
      * @param scope current scope
@@ -51,7 +60,7 @@ public abstract class ShapeOps {
     }
 
     /**
-     * flatten the shape to 1 dimension
+     * Flatten the operand to 1 dimension
      *
      * @param <T> the type of operand
      * @param <U> the shape datatype.
@@ -67,12 +76,11 @@ public abstract class ShapeOps {
     }
 
     /**
-     * flatten the shape to 1 dimension
+     * Flatten the shape to 1 dimension
      *
      * @param scope current scope
      * @param shape the TensorFlow shape
      * @return the flattened shape
-     * @see reduceDims
      */
     @Endpoint(name = "flatten")
     public static Operand<TInt32> flatten(Scope scope, Shape<TInt32> shape) {
@@ -80,14 +88,13 @@ public abstract class ShapeOps {
     }
 
     /**
-     * flatten the shape to 1 dimension
+     * Flatten the shape to 1 dimension
      *
      * @param <U> the shape datatype.
      * @param scope current scope
      * @param shape the TensorFlow shape
      * @param dType the shape datatype.
      * @return the flattened shape
-     * @see reduceDims
      */
     @Endpoint(name = "flatten")
     public static <U extends TNumber> Operand<U> flatten(Scope scope, Shape<U> shape, DataType<U> dType) {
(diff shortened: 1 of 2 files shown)


JimClarke5:
@googlebot I signed it!

JimClarke5:
@googlebot I fixed it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -0,0 +1,217 @@
+/*
```
Craigacp, line 1: Please add the right license header (i.e. the non-Oracle one that references Tensorflow authors).

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/op/core/ShapeOpsTest.java
```diff
@@ -0,0 +1,305 @@
...
+            
+        }
+    }
+
+    /**
+     * Test of reduceDims method, of class ShapeOps.
+     */
+    @Test
+    public void testReduceDims_Operand_Operand() {
+        System.out.println("reduceDims");
```
Craigacp, line 240: What's this test trying to do?

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/op/core/ShapeOpsTest.java
```diff
@@ -0,0 +1,305 @@
...
+            assertEquals(TInt32.DTYPE, stf.datatype());
+        }
+    }
+
+    /**
+     * Test of create method, of class ShapeOps.
+     */
+    @Test
+    public void testCreate_Scope_DataType() {
+        System.out.println("create DataType");
```
Craigacp, line 68: Given the number of tests that are in TF at the moment, having it print out the name of each one is a little much.

JimClarke5: Ok I will remove the println's

Craigacp:
Might be worth running some of the tests in eager mode, just to make sure. And I'm not clear if you want this to be bound into the same ops instance that everything else uses, in which case you probably should annotate the methods with `@Endpoint` and the class with `@Operator` similar to this class https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Helpers.java, then you'll need to refactor the methods to accept the scope as their first argument. Once you've done that they'll appear in the main tf object and it'll feel a lot more like the python API.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/op/core/ShapeOpsTest.java
JimClarke5: My mistake, It was intended to be the same as testReduceDims_Shape_Operand, but based on passing the operand rather than the shape directly. It may be a redundant, unless you think we need to test it both ways.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: OK

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -0,0 +1,329 @@
...
+import org.tensorflow.types.family.TType;
+
+/**
+ * An operator working with org.tensorflow.op.core.Shape tensors
+ */
+@Operator
+public abstract class ShapeOps {
+
+    /**
+     * flatten the shape to 1 dimension
```
karllessard, line 41: For consistency with other classes, please start the doc summary with a capital letter.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -0,0 +1,329 @@
...
+     * @param <U> the shape datatype. the type of the shape
+     * @param scope current scope
+     * @param shape the TensorFlow shape
+     * @param dType the shape datatype.
+     * @return the size
+     */
+    @Endpoint(name = "size")
+    public static <U extends TNumber> Operand<U> size(Scope scope, Shape<U> shape, DataType<U> dType) {
+        Slice<U> dims = Slice.create(scope, shape,
+                Cast.create(scope, Constant.arrayOf(scope, (new int[]{0})), dType),
```
karllessard, line 123: Parentheses not required around the array instantiation

Craigacp:
I think these shape utils are pretty similar in kind to the variable with init we've already got in core. It's very low level unlike the optimizers or the datasets in the `tensorflow-framework` which are much higher level.

karllessard:
> I think these shape utils are pretty similar in kind to the variable with init we've already got in core. It's very low level unlike the optimizers or the datasets in the `tensorflow-framework` which are much higher level.

Yeah, that's fair. So please @JimClarke5 just readjust the few things that are left in this PR so I can merge it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: Ok

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: OK

PR #72, opened 2020-06-18
Remove dependency on javacpp-platform to avoid pulling unused code

saudet:
Also disable MKL and CUDA builds for Windows until we get machines for them

But add plain Windows snapshots to the redeploy job on GitHub Actions


karllessard:
First time I see this, thanks @saudet 
image

PR #73, opened 2020-06-18
Update UnitTest cases to JUnit5

JimClarke5:
Requires changing import packages, change @Before to @BeforeEach annotations, remove @RunTest annotation. Change Junit dependency in pom.xml's.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
+++ b/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
@@ -177,6 +177,7 @@ import org.tensorflow.op.core.Select;
 import org.tensorflow.op.core.SetDiff1d;
 import org.tensorflow.op.core.SetSize;
 import org.tensorflow.op.core.ShapeN;
+import org.tensorflow.op.core.ShapeOps;
 import org.tensorflow.op.core.Size;
 import org.tensorflow.op.core.Skipgram;
 import org.tensorflow.op.core.Slice;
@@ -336,9 +337,9 @@ public final class Ops {
 
   public final SignalOps signal;
 
-  public final TrainOps train;
-
   public final QuantizationOps quantization;
+
+  public final TrainOps train;
 
   private final Scope scope;
 
@@ -359,8 +360,8 @@ public final class Ops {
     math = new MathOps(scope);
     audio = new AudioOps(scope);
     signal = new SignalOps(scope);
-    train = new TrainOps(scope);
     quantization = new QuantizationOps(scope);
+    train = new TrainOps(scope);
   }
 
   /**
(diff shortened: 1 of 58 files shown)


tensorflow-framework/src/test/java/org/tensorflow/framework/data/MapDatasetTest.java
```diff
@@ -15,8 +15,9 @@
  */
 package org.tensorflow.framework.data;
 
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
Craigacp, line 19: Could you clean up the static imports so they are adjacent?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -336,10 +337,10 @@
 
   public final SignalOps signal;
 
-  public final TrainOps train;
-
   public final QuantizationOps quantization;
 
+  public final TrainOps train;
```
Craigacp, line 342: Looks like this got updated in after the shape change was integrated. @karllessard do you think it's worth making a separate PR (or just you push the ops update separately) so we don't have commits that make unintentional changes?

tensorflow-framework/src/test/java/org/tensorflow/framework/data/MapDatasetTest.java
JimClarke5: Actually there were several cases of unused imports for JUnit. I used a sed script to make the JUnit changes, so these were there before, but I'll be glad to clean them up.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
karllessard: I think it is up to you @JimClarke5 , normally this file should have been committed in your previous PR adding shape operators. Ideally we would commit it as a separate PR to keep this history clean, but I won't insist neither if you prefer we keep things simpler and just merge this PR as is.

JimClarke5:
I am not sure what happened. It should have all been committed with the ShapeOPs PR before I started working on JUNIT. 

jim

On Jun 21, 2020, at 8:54 AM, Karl Lessard <notifications@github.com> wrote:


@karllessard commented on this pull request.

In tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java <https://github.com/tensorflow/java/pull/73#discussion_r443216932>:

>    public final QuantizationOps quantization;
 
+  public final TrainOps train;
I think it is up to you @JimClarke5 <https://github.com/JimClarke5> , normally this file should have been committed in your previous PR adding shape operators. Ideally we would commit it as a separate PR to keep this history clean, but I won't insist neither if you prefer we keep things simpler and just merge this PR as is.

karllessard:
> I am not sure what happened. It should have all been committed with the ShapeOPs PR before I started working on JUNIT. 

So are you planning to make a separate PR for it or we just go ahead and merge it as part of the current PR?

JimClarke5:
Whatever is easiest. 

Jim

PR #74, opened 2020-06-19
Also extract and redeploy pom.xml file from tensorflow-core-api with GitHub Actions

saudet:
Fixes https://github.com/tensorflow/java-models/issues/4


karllessard:
If it is required for Gradle to have a pom that has the same timestamp as the other artifact (can you confirm this?), should we do the same for all artifacts? 

I'm thinking more specifically about the `source` jar, which will still have an older timestamp. Will users have any problems to download the sources of a given version from their IDE when working with Gradle?

saudet:
Correct, Gradle won't see those `javadoc` and `sources` either, but I don't think anyone is going to complain about that. Anyway, I've added the dependencies to the pom.xml file in the last commit, so they will get redeployed as well now.

PR #78, opened 2020-06-23
Add Maven profile for quick development build

karllessard:
This PR adds a new Maven profile to facilitate the environment setup for developers by skipping the native library build completely and fetching pre-compiled native artifacts from Nexus instead.

This profile is also applied by default for all incoming pull requests to `master`, which in most case do not require a complete build status check to be approved and will release some of our limited resources on GitHub Action. It is still possible though to build the native libraries on a given pull request by adding the `CI build` label to it.

I've picked the name `dev` for this profile as I wanted something very brief since, while not being applied by default, this profile might be used very often by developers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,21 @@ on:
   pull_request:
     branches:
       - master
-
+    types: [opened, reopened, synchronize, labeled, unlabeled]
 jobs:
+  quick-build:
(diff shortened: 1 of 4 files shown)


karllessard:
You can actually see the quick results of the CI jobs for this PR [here](https://github.com/tensorflow/java/actions/runs/144333355), which took 1m43s vs ~6h with a complete build.

karllessard:
I think it is pretty safe to merge this one now.

PR #80, opened 2020-06-30
Added methods head, take, tail, takeLast, prepend and append …

JimClarke5:
to mirror org.tensorflow.tools.Shape

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
@@ -15,28 +15,21 @@ limitations under the License.
 package org.tensorflow.op.core;
 
 import java.util.Arrays;
+
 import org.tensorflow.DataType;
 import org.tensorflow.Operand;
 import org.tensorflow.op.Scope;
 import org.tensorflow.op.annotation.Endpoint;
 import org.tensorflow.op.annotation.Operator;
-import org.tensorflow.op.core.Concat;
-import org.tensorflow.op.core.ExpandDims;
-import org.tensorflow.op.core.Reshape;
-import org.tensorflow.op.core.Shape;
-import org.tensorflow.op.core.Size;
-import org.tensorflow.op.core.Slice;
+
 import org.tensorflow.op.math.FloorMod;
-import org.tensorflow.op.core.Constant;
-import org.tensorflow.op.core.Gather;
-import org.tensorflow.op.core.OnesLike;
-import org.tensorflow.op.core.ReduceProd;
-import org.tensorflow.op.core.Where;
+
 import org.tensorflow.op.dtypes.Cast;
 import org.tensorflow.op.math.NotEqual;
 import org.tensorflow.op.math.Sub;
 import org.tensorflow.types.TBool;
 import org.tensorflow.types.TInt32;
+import org.tensorflow.types.TInt64;
 import org.tensorflow.types.family.TNumber;
 import org.tensorflow.types.family.TType;
 
@@ -49,28 +42,30 @@ public abstract class ShapeOps {
     /**
      * Flatten the operand to 1 dimension
      *
-     * @param <T> the type of operand
-     * @param scope current scope
+     * @param <T>     the type of operand
+     * @param scope   current scope
      * @param operand the operand to flatten
      * @return the reshaped operand
      */
     @Endpoint(name = "flatten")
-    public static <T extends TType> Operand<T> flatten(Scope scope, Operand<T> operand) {
+    public static <T extends TType> Operand<T> flatten(
+            Scope scope, Operand<T> operand) {
         return flatten(scope, operand, TInt32.DTYPE);
     }
 
     /**
      * Flatten the operand to 1 dimension
      *
-     * @param <T> the type of operand
-     * @param <U> the shape datatype.
-     * @param scope current scope
+     * @param <T>     the type of operand
+     * @param <U>     the shape datatype.
+     * @param scope   current scope
      * @param operand the operand to flatten
-     * @param dType the shape datatype.
+     * @param dType   the shape datatype.
      * @return the reshaped operand
      */
     @Endpoint(name = "flatten")
-    public static <T extends TType, U extends TNumber> Operand<T> flatten(Scope scope, Operand<T> operand, DataType<U> dType) {
+    public static <T extends TType, U extends TNumber> Operand<T> flatten(
+            Scope scope, Operand<T> operand, DataType<U> dType) {
         Operand<U> flatShape = flatten(scope, Shape.create(scope, operand, dType), dType);
         return Reshape.create(scope, operand, flatShape);
     }
@@ -90,14 +85,15 @@ public abstract class ShapeOps {
     /**
      * Flatten the shape to 1 dimension
      *
-     * @param <U> the shape datatype.
+     * @param <U>   the shape datatype.
      * @param scope current scope
      * @param shape the TensorFlow shape
      * @param dType the shape datatype.
      * @return the flattened shape
      */
     @Endpoint(name = "flatten")
-    public static <U extends TNumber> Operand<U> flatten(Scope scope, Shape<U> shape, DataType<U> dType) {
+    public static <U extends TNumber> Operand<U> flatten(
+            Scope scope, Shape<U> shape, DataType<U> dType) {
         return ExpandDims.create(scope,
                 size(scope, shape, dType),
                 Cast.create(scope, Constant.scalarOf(scope, -1), TInt32.DTYPE));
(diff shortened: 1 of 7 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -337,4 +336,252 @@
...
+     * Returns a new Shape, with another Shape's dimensions prepended.
+     *
+     * @param scope          current scope
+     * @param shape          the TensorFlow shape
+     * @param otherShape     the other shape to prepend
+     * @return a new shape with another Shape's dimensions prepended.
+     */
+    @Endpoint(name = "append")
+    public static <T extends TNumber> Operand<T> append(
+            Scope scope, Operand<T> shape, Operand<T> otherShape) {
```
karllessard, line 577: Idem, should the `Operand` here be `Shape` instead?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -337,4 +336,252 @@
...
+     * Returns a new Shape, with another Shape's dimensions prepended.
+     *
+     * @param scope          current scope
+     * @param shape          the TensorFlow shape
+     * @param otherShape     the other shape to prepend
+     * @return a new shape with another Shape's dimensions prepended.
+     */
+    @Endpoint(name = "prepend")
+    public static <T extends TNumber> Operand<T> prepend(
+            Scope scope, Operand<T> shape, Operand<T> otherShape) {
```
karllessard, line 523: should the `Operand` here be `Shape` instead? I think we need to use strong typing here otherwise a user might easily think that `tf.append` can be used for any kind of tensors.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
```diff
@@ -49,28 +42,30 @@
...
-     * @param scope current scope
+     * @param <T>     the type of operand
+     * @param scope   current scope
      * @param operand the operand to flatten
      * @return the reshaped operand
      */
     @Endpoint(name = "flatten")
-    public static <T extends TType> Operand<T> flatten(Scope scope, Operand<T> operand) {
+    public static <T extends TType> Operand<T> flatten(
+            Scope scope, Operand<T> operand) {
```
karllessard, line 52: Just double-checking with you @JimClarke5 that you are using Google Java Formatter in your IDE, as there are a lot of reformatting changes in the PR (quickly like that, it looks like the previous version of the code was closer to what we find in the other source files).

JimClarke5: I am using IntelliJ which seemed to have its own mind about formatting. I will see if I can add the Google Java Formatter to it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: I assume you are referring to org.tensorflow.op.core.Shape.
This class is referring to "dynamic shapes" as detailed in the python code, as opposed to "static shapes" represented
by org.tensorflow.tools.Shape.

org.tensorflow.op.core.Shape is a specific kind of operand, 
but any Operand that translates into a set of numbers works.
For example, tf.constant(new int[] { 1,2,3}), tf.slice(....), tf.concat(), tf.stack(), tf.shape(operand) or tf.constant(org.tensorflow.tools.Shape) all work equally. As far as I can tell there is no way to create a org.tensorflow.op.core.Shape from such arrays of numbers. tf.shape() only takes the shape from an Operand.
The primary use is in creating and manipulating shapes within the Graph operations and outside for Java code per se.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: I assume you are referring to `org.tensorflow.op.core.Shape`.
These new operations manipulate  "dynamic shapes" as detailed in the python code, as opposed to "static shapes" represented by `org.tensorflow.tools.Shape`. The primary use of this change is in creating and manipulating shapes within the TF operations, outside of Java code. Dynamic shape manipulations usually results because the static Shape has unknown dimensions or the shape needs to be manipulated within the graph to be compatible with another Graph Operation.

`org.tensorflow.op.core.Shape` is a specific kind of operand that contains dimensions, 
but any Operand that translates into a set of numbers (dimensions) works.
For example,` tf.constant(new int[] { 1,2,3}), tf.slice(....), tf.concat(), tf.stack(), tf.shape(operand) or tf.constant(org.tensorflow.tools.Shape)` all work equally. Also, the output of these all work with tf.reshape(),
defined as `reshape(Operand<T> tensor, Operand<U> shape)`.

As far as I can tell, there is no way to create a `org.tensorflow.op.core.Shape` object from Operands containing dimensions. `tf.shape()` only takes the shape associated with  the Operand, itself. Perhaps we need an operand that creates a `org.tensorflow.op.core.Shape` object from dimensions,  but using an Operand of dimensions seems to suffice, at least at the level that I need to  use these functions.

This PR creates three signatures for append, all for manipulating the dimensions of a "shape".

> Operand<T> append(Operand<T> shape, Operand<T> otherShape)
> Operand<TInt64> append(Shape<TInt64> shape,  long lastDimension)
> Operand<TInt32> append(Shape<TInt32> shape,  int lastDimension)

The Shape<TIntxx> versions os these methods require a Shape object but they are really a convenience in passing
the lastDimension as a Java int/long. Calling `append(shapeObj, tf.constant(new int/long[]i) )`
produces the same result, but is not as straight forward.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: Ok, if found the google-java-format plugin for intellij, so I can push up the reformatted code, pending  resolution to the first 2 questions.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
karllessard: Ok I understand your point, it is in case the given `shape` is any kind of tensor and not necessarily the result of the `Shape` op, make sense. 

Maybe we should update the documentation then because it stated that all the endpoint returns "a Shape" while it simply returns any integer rank-1 tensor that could be used as a shape or not (it looks like nothing would prevent a user to use this endpoint for any other reason).

If we want to make it clear that it should be used mainly for shapes, maybe we should add a prefix to all this endpoints so that they are accessible via `tf.shape.prepend` instead of just `tf.prepend`? That would be possible by simply adding a group to the operator annotation, i.e. `@Operator(group="shape")`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
karllessard: closing this thread as it is identical to the one above

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/ShapeOps.java
JimClarke5: I added the `@Operator(group="shape")`, I was thinking of doing this anyway as it was growing bigger that I had originally thought and now it makes it clear what it is used for.
I changed the JavaDoc and some other parameters to  clarify that Operand<> of dimensions are used rather than Shape objects. I reformatted with google-java-format formatter on both ShapeOps and ShapeOpsTest. Also cleaned up ShapeOps based on InteliJ Code Analysis. Changes are pushed.

karllessard: Thanks @JimClarke5 , looks good. The only thing though is that I'm pretty sure that when you ran `mvn install` on your branch, a few generated classes ended up being modified or added under `src/gen/`. These need to be committed as part of this PR too. 

Also, probably that one of the file created is called `ShapeOps` as well and found under `src/gen/annotations/org/tensorflow/op`. It won't conflict with yours but it might be confusing to have two classes with the same name in the same jar. So I suggest then that you rename your operator class just `Shapes` (and `ShapesTest`). 

Note: at some point we will probably also need to rename the `core` package found in `src/main/java/org/tensorflow/op` as it conflicts with the same package found under `src/gen/java` but we can deal with this another time.

JimClarke5: Ok, I have renamed `ShapeOps` ==> `Shapes` and `ShapeOpsTest` ==> `ShapesTest` under main.
`ShapeOps` under gen is generated from "Shapes.java", so I assumed (wrongly) that it wasn't tracked in Git,
so I have added it and Ops.java back into my local repository.

FYI, from my experience with Java modules, post Java 8, is that it gets very finicky about duplicate package names, especially across separate jars. 

Having duplicate class names under different packages is possible, but gets nasty when using the  identically named classes within  the same code of another class, as I am finding out with the two Shape classes. Many times in the Keras code, it tries to first resolve the shape statically using `org.tensorflow.tools.Shape`, but if it can't, it then resolves the shape dynamically using `org.tensorflow.op.core.Shape`. The resolution is to fully qualify the class with the package name when using both identically named classes.

karllessard: > gets nasty when using the identically named classes within the same code of another class, as I am finding out with the two Shape classes.

I totally agree with you, I have the same surprise every time I'm working in the `org.tensorflow.op.core` package and I realize that my `Shape` instances are not of the expected type. I think it makes more sense that the static version keeps the `Shape` name and that we should override the name for the generated one (maybe `TensorShape`?). 

Let's do this another time though, this PR looks good to be merged now, thanks!

karllessard:
Hey @JimClarke5, it looks like the build failed after merging your PR, can you please look at it? I won’t have access to my computer for some days: https://github.com/tensorflow/java/runs/837060226?check_suite_focus=true

PR #83, opened 2020-07-12
Add a simple DeviceSpec class.

hthu:
This is a straightforward port from device_spec.py.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/device_spec.py


Craigacp:
Do you know if the Auto value stuff will be upgraded to use Records (https://openjdk.java.net/jeps/384) when it comes out of preview? As otherwise it might be preferable to use plain Java classes that we can convert into records when support comes (as I think it's supposed to be a binary compatible change).

hthu:
> Do you know if the Auto value stuff will be upgraded to use Records (https://openjdk.java.net/jeps/384) when it comes out of preview? As otherwise it might be preferable to use plain Java classes that we can convert into records when support comes (as I think it's supposed to be a binary compatible change).

I don't :)
Though I'd imagine it to be a fairly easy port over?

Also taking a look at the record, it doesn't seem supporting a customer builder or stuff like that? In our case, the values are optional - the default null value builder seems cleaner ATM.

Craigacp:
Yeah, you have to specify the builder by hand, those aren't part of the language support for records. Though you can have multiple constructors for a record provided they call into the canonical constructor.

karllessard:
Thanks for the PR @hthu , 

I would suggest having a builder that does not depend on any external library. `tensorflow-core-api` is a low-level library that should bring a limited number of transitive dependencies as it increases the risk of incompatibilities if the user also depends on these other libraries. 

We had unfortunately no choice with `protobuf-java` but in your case, while auto-value provides an elegant way to build such builder, it can be easily replaced by a simple custom builder.

karllessard:
Hi @hthu , do you intend to make changes on this PR or leave it as is? Do you have any feedback to provide based on the suggestions @Craigacp and I have made?

hthu:
Oops, totally forgot about this.
Will drop autovalue deps and deal with it over the weekend!

hthu:
Updated to a version without auto.
PTAL.

hthu:
Updated to a version without using auto.
PTAL.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/device/DeviceSpec.java
```diff
@@ -0,0 +1,166 @@
+package org.tensorflow.device;
+
+import com.google.common.base.Joiner;
```
karllessard, line 3: Sorry @hthu , you should not use Guava neither :)

Guava is present in the classpath at compile time only as a result of depending on `tensorflow-core-generator` for running the annotation processor. As you can [see in the pom](https://github.com/tensorflow/java/blob/81a2b605da90689fbdaed975106fb2bd5f85ca6c/tensorflow-core/tensorflow-core-api/pom.xml#L41), this dependency is optional, meaning that it will not be automatically present at runtime nor its transitive dependencies. I'm even surprised they were included when running your unit tests.

Again, sorry for the trouble but I think we should try to avoid adding a new runtime dependency in the core module if there is an easy way to accomplish the same task without it.

hthu: NP!
Removed as well.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/device/DeviceSpec.java
```diff
@@ -0,0 +1,162 @@
+package org.tensorflow.device;
+
+import javax.annotation.Nullable;
```
Craigacp, line 3: Where do we get the Nullable annotation from?

karllessard: I was wondering as well, I assumed (maybe a bit too fast) that it was part of the JDK... ok, I'll remove them manually and merge back the file without it

karllessard: Yeah, it was coming again from our dependency to the `tensorflow-core-generator`, which got it from `com.google.code.findbugs`, thanks for raising this @Craigacp

hthu: Woah I always thought @Nullable was part of the JDK somewhere...
My bad for taking that for granted.
Are you folks already removing that?

Craigacp: No worries. Can we remove the generator dependency somehow? It pulls in a lot of things.

Craigacp: Personally I'm a bit confused as to why jsr305 wasn't integrated into the JDK, but there was a talk about nullness annotations from Google at the last JVMLS discussing all the issues with integrating them into things, and it seemed to be pretty complex. Unfortunately that talk wasn't recorded.

karllessard: > Can we remove the generator dependency somehow?

I was thinking of doing that as well but there is probably a good reason why it is there, probably maven was failing somehow to run the annotation processor without it. Will need to double check

So @hthu , it's all good now, I forced-pushed a new version with `@Nullable`. I also took the opportunity to move the `DeviceSpec` out of the `device` package. Since there was only one file in that package, I don't think we needed it.

hthu: Sounds good to me!
Thank you both!

Craigacp: > > Can we remove the generator dependency somehow?
> 
> I was thinking of doing that as well but there is probably a good reason why it is there, probably maven was failing somehow to run the annotation processor without it. Will need to double check

I guess without the explicit dependency it's up to maven which one of the generator and the core api it compiles first.  Maybe we could make it a test time dependency? That's even less clear in terms of the semantics, but it should make it run first.

Issue #85, opened 2020-07-17
Unable to execute tfhub model: getting TFInvalidArgumentException


samikrc:
<em>Please make sure that this is a bug. As per our [GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md), we only address code/doc bugs, performance issues, feature requests and build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Debian Buster
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device: No
- TensorFlow installed from (source or binary): binary
- TensorFlow version (use command below): 0.2.0-SNAPSHOT (from sonatype)
- Python version: -
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version:
- GPU model and memory:

You can collect some of this information using our environment capture [script](https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh)
You can also obtain the TensorFlow version with
python -c "import tensorflow as tf; print(tf.GIT_VERSION, tf.VERSION)"

**Describe the current behavior**

I am trying to invoke a TFHUB model (Universal Sentence Encoder v4) using the new java API (using Scala). However, I am getting stuck at the error below.
```
An exception or error caused a run to abort: Malformed TF_STRING tensor; too short to hold number of elements 
org.tensorflow.exceptions.TFInvalidArgumentException: Malformed TF_STRING tensor; too short to hold number of elements
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:87)
	at org.tensorflow.Session.run(Session.java:595)
	at org.tensorflow.Session.access$100(Session.java:70)
	at org.tensorflow.Session$Runner.runHelper(Session.java:335)
	at org.tensorflow.Session$Runner.run(Session.java:285)
	at org.samik.EmbeddingModelServer.USEEmbeddingServerTest.<init>(USEEmbeddingServerTest.scala:85)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[...]

PR #86, opened 2020-07-21
Copy NdArrays to standard arrays of n-dimensions

karllessard:
This PR is in linked with what was observed and discussed in https://github.com/tensorflow/java/issues/85. While it is possible to transform a standard array of n-dimension to a `NdArray`, the other way around is not possible, or at least not simple enough.

I'm adding here two new signatures for converting back `NdArray` instances to standard arrays:
`StdArrays.array*dCopyOf(NdArray<T>)` and `StdArrays.copyFrom(T[], NdArray<T>)`, for all their primitive types and when `*` is replaced by the number of dimensions. For example:

```java
IntNdArray matrix = StdArrays.ndCopyOf(new int[][] { { 1, 2, 3 }, { 4, 5, 6 } });
int[][] array = StdArrays.array2dCopyOf(matrix);
assertEquals(2, array.length);
assertEquals(3, array[0].length);
assertArrayEquals(new int[]{ 1, 2, 3 }, array[0]);
```

Note that while `array*dCopyOf` returns an array with the exact same dimensions as the original `NdArray`, `copyFrom` is more permissive and allow larger arrays to be used, as long as the number of dimensions matches. For example:

```java
IntNdArray matrix = StdArrays.ndCopyOf(new int[][] { { 1, 2, 3 }, { 4, 5, 6 } });
int[][] array = new int[3][4];
StdArrays.copyFrom(matrix, array);
assertArrayEquals(new int[]{ 1, 2, 3, 0 }, array[0]);
assertArrayEquals(new int[]{ 4, 5, 6, 0 }, array[1]);
assertArrayEquals(new int[]{ 0, 0, 0, 0 }, array[2]);
```

### DIFF
--- a/ndarray/src/main/java/org/tensorflow/ndarray/StdArrays.java
+++ b/ndarray/src/main/java/org/tensorflow/ndarray/StdArrays.java
@@ -2,6 +2,9 @@ package org.tensorflow.ndarray;
 
 import static org.tensorflow.ndarray.NdArrays.vectorOf;
 
+import java.lang.reflect.Array;
+import org.tensorflow.ndarray.buffer.DataBuffers;
+
 /**
  * Utility class for working with {@link NdArray} instances mixed with standard Java arrays.
  */
@@ -15,7 +18,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -27,7 +30,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[][] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -39,7 +42,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[][][] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -51,7 +54,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[][][][] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -63,7 +66,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[][][][][] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -75,7 +78,7 @@ public final class StdArrays {
    */
   public static IntNdArray ndCopyOf(int[][][][][][] array) {
     IntNdArray ndArray = NdArrays.ofInts(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -87,7 +90,7 @@ public final class StdArrays {
    */
   public static LongNdArray ndCopyOf(long[] array) {
     LongNdArray ndArray = NdArrays.ofLongs(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -99,7 +102,7 @@ public final class StdArrays {
    */
   public static LongNdArray ndCopyOf(long[][] array) {
     LongNdArray ndArray = NdArrays.ofLongs(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -111,7 +114,7 @@ public final class StdArrays {
    */
   public static LongNdArray ndCopyOf(long[][][] array) {
     LongNdArray ndArray = NdArrays.ofLongs(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -123,7 +126,7 @@ public final class StdArrays {
    */
   public static LongNdArray ndCopyOf(long[][][][] array) {
     LongNdArray ndArray = NdArrays.ofLongs(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
@@ -135,7 +138,7 @@ public final class StdArrays {
    */
   public static LongNdArray ndCopyOf(long[][][][][] array) {
     LongNdArray ndArray = NdArrays.ofLongs(shapeOf(array));
-    copyTo(ndArray, array);
+    copyTo(array, ndArray);
     return ndArray;
   }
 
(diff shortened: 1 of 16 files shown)


karllessard:
Hey @Craigacp , I totally understand your concerns as they were mine a few hours before. While I don't disagree, let me explain why I ended up with these names.

About `array2dCopyOf(...)`: we already have `ndCopyOf(...)`, which takes in parameter a standard array and convert it to a `NdArray`. The new method is basically the opposite of this one, so I tried to find a name showing that they are related (in an ideal world, I could have simply named them `1dCopyOf`, `2dCopyOf`, etc.). 

About `copyFrom(...)`: again, it was to show that this method do the opposite of `copyTo`, which copies a standard array to a `NdArray`. Since there is no object in context, I thought that the `StdArrays` namespace implicitly indicates that the verb is relative to the standard array. I also swapped at last minute the order of the parameters to match the `dst/src` pair in `copyTo` but I wasn't too sure about this.

karllessard:
Maybe both `copyFrom` and `copyTo` should only be named `copy`, where in one the source is an `NdArray` and in the other it is a standard array?

Craigacp:
I hadn't noticed it was flipped in other places, but I think in general it's better to have `src` then `dest` everywhere. Then it follows System.arraycopy.

I'm ok with the name mirroring, I noticed that once I'd posted the comment and went back to this initial PR page where your example is, so that lessens my objections (but still we might want to reconsider all these names as the ndarray API builds out, they might start to be uncomfortable).

In reference to `copyFrom` & `copyTo` I think more descriptive names are better, as the only difference between the two `copy` methods would be argument order right? And I think that would be confusing as they won't really behave like other overloads.

karllessard:
Ok, I can revert `copyTo` arguments as well then, I think that the few users we are having now probably use more `ndCopyOf` anyway

karllessard:
FYI, I've forced push the requested changes.

ndarray/src/main/java/org/tensorflow/ndarray/StdArrays.java
```diff
@@ -585,663 +588,2216 @@ public static BooleanNdArray ndCopyOf(boolean[][][][][][] array) {
...
-  public static void copyTo(LongNdArray dst, long[][][] array) {
+  public static void copyTo(long[][][] src, LongNdArray dst) {
     dst.elements(1).forEachIndexed((idx, e) ->
-        NdArrays.vectorOf(array[(int)idx[0]][(int)idx[1]]).copyTo(e)
+        NdArrays.vectorOf(src[(int)idx[0]][(int)idx[1]]).copyTo(e)
     );
   }
 
   /**
    * Copy a 4-dimensions array of longs into the {@code dst} {@link NdArray}
```
Craigacp, line 1446: `dimensions` should probably be `dimensional`.

samikrc:
@karllessard @Craigacp  Sorry, looks like I am late in the discussion. Based on my understanding of the discussion here, it seems that we got the new API's in the NDArray class. I think an additional helpful API would be to add a copyTo method directly to the Tensor class, like we have in the 1.15.x, and like the very helpful scalarOf/vectorOf methods that already exist on the Tensor class. Obviously the expectation is that this method will internally call the corresponding method of the NDArray class. Any thoughts on this? Can log a new issue if this requires more discussion.

Thanks again on the first turnaround on this. Quick question: If I want to pull this and other similar bleeding-edge changes, I cannot do this from sonatype, right? Currently I have this in my pom:
```
        <dependency>
            <artifactId>tensorflow-core-api</artifactId>
            <version>0.2.0-SNAPSHOT</version>
        </dependency>
        <dependency>
            <artifactId>tensorflow-core-api</artifactId>
            <version>0.2.0-SNAPSHOT</version>
            <classifier>linux-x86_64</classifier>
        </dependency>
```
alongwith, the sonatype repo listed in the <repositories> block.

karllessard:
Hi @samikrc , no worries you are not too late for the party ;)

I prefer to avoid polluting the `Tensor` or the `NdArray` interface with a huge amount of new signatures to support all possible primitive types and dimensions. Plus, while sometimes in cannot be avoided like in your case, it is recommended to stick as much as possible to the `NdArray` interface and avoid using multi-dimensional Java arrays, which tends to be slow and cannot map continuous memory like a tensor. There is an ongoing initiative to standardize such new API in Java and improve interoperability between different libraries (btw, `org.tensorflow:ndarray` does not depend on TensorFlow itself and can be used outside of it). For all these reasons, these signatures are listed in a separate class called `StdArrays`. 

The previous implementation in 1.x was not type safe neither, as you could request to copy a tensor of float to an array of integers, which will only fail at runtime. The new API prevent this at compile time.

Just to let you know to that I'm actually working on a proposal to implement both the `TType` interfaces (e.g. `TFloat32`) and the `Tensor` one in the same object, which will reduce the number of conversions as required by `tensor` and `tensor.data()` and should simplify a little bit the API as well.

> If I want to pull this and other similar bleeding-edge changes, I cannot do this from sonatype, right?

Yes you can, just make sure to force an update of your snapshots when you build (`mvn install -U`) as we do not always increase the version. Note also that after a PR is merged to the master branch, it can take up to 6 hours before the snapshots get refreshed in Sonatype.

karllessard:
@saudet, the windows build is failing for no apparent reasons, any idea? I really doubt that changes in this small PR caused the issue: https://github.com/tensorflow/java/runs/903746593?check_suite_focus=true

saudet:
No idea. `bash build.sh` outputs garbage, and that's an error code I've never seen before:
> Process exited with an error: -1

Sounds like a problem with the VM itself...

karllessard:
It looks like 'bash' is not installed anymore on the GitHub windows machine:
```
[INFO] Building for platform "windows-x86_64"
[INFO] bash D:\a\java\java\tensorflow-core\tensorflow-core-api/build.sh 
Windows Subsystem for Linux has no installed distributions.


Distributions can be installed by visiting the Microsoft Store:


https://aka.ms/wslstore
```
We need to install it manually

karllessard:
According to this post, just adding Git binaries to the path did the job: https://github.community/t/how-do-i-install-bash-on-my-windows-local-runner/17439/4, trying it: https://github.com/tensorflow/java/runs/907717855?check_suite_focus=true

karllessard:
Still didn't work (anyway we wouldn't have been able to checkout the repository if the binaries under Git/bin were not available...)

karllessard:
I've posted this to the community support of GitHub in regard of this new issue: https://github.community/t/wsl-not-available-for-hosted-windows-machine/124389

karllessard:
or is it possible that we install something on the Windows machine that now gets in conflict with WSL?

saudet:
I'm guessing the version of bash getting used changed, but changed from what? I didn't think of adding a `bash --version` line...

Craigacp:
Maybe the Windows version got upgraded? WSL got pretty aggressive at pretending to be things in one of the more recent releases. That could cause it to override bash with something higher up the path saying you need to install WSL to make this work. How is the path set on the VM?

JWenBin:
I still don't know how to copy tensor data to standard arrays of n-dimensions o(╯□╰)o

samikrc:
> 
> 
> I still don't know how to copy tensor data which returned by session().runner().run.get(0) to standard arrays of n-dimensions o(╯□╰)o

Does the code in this file help? 
https://github.com/samikrc/embedding-model-server/blob/tf2.2/src/test/scala/org/samik/EmbeddingModelServer/USEEmbeddingServerTest.scala

karllessard:
There are many methods in the StdArrays utility class that allows you to do this: https://github.com/tensorflow/java/blob/cc5c86ceaaa176cfd8948aa549ccc24f2a75783c/ndarray/src/main/java/org/tensorflow/ndarray/StdArrays.java#L609

JWenBin:
> 
> There are many methods in the StdArrays utility class that allows you to do this: https://github.com/tensorflow/java/blob/cc5c86ceaaa176cfd8948aa549ccc24f2a75783c/ndarray/src/main/java/org/tensorflow/ndarray/StdArrays.java#L609
> […](#)
> On Apr 2, 2021, at 03:54, JWenBin ***@***.***> wrote: ﻿ I still don't know how to copy tensor data to standard arrays of n-dimensions o(╯□╰)o — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

Thank you both you two! I understand now.

PR #89, opened 2020-07-27
Draft: Java API to use tf.function available on SavedModel.

Shajan:
Python models that contain tf.function is inconvenient to be consumed by Java clients.
This proposal provides an API to
(a) Invoke a tf.function, given the signature name
(b) Retrieve the node name in the graph corresponding to a tf.function

TBD: Unit tests before this change is ready to merge.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
@@ -20,6 +20,9 @@ import static org.tensorflow.internal.c_api.global.tensorflow.TF_NewGraph;
 import static org.tensorflow.internal.c_api.global.tensorflow.TF_SetConfig;
 
 import com.google.protobuf.InvalidProtocolBufferException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
 import org.bytedeco.javacpp.BytePointer;
 import org.bytedeco.javacpp.PointerPointer;
 import org.bytedeco.javacpp.PointerScope;
@@ -32,6 +35,7 @@ import org.tensorflow.internal.c_api.TF_Status;
 import org.tensorflow.proto.framework.ConfigProto;
 import org.tensorflow.proto.framework.MetaGraphDef;
 import org.tensorflow.proto.framework.RunOptions;
+import org.tensorflow.proto.framework.SignatureDef;
 
 /**
  * SavedModelBundle represents a model loaded from storage.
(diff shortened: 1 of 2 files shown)


karllessard:
Hi @Shajan , can you please rebase this PR to go the shared branch created for this story we will working on together? https://github.com/tensorflow/java/tree/shared-saved-model

I think we should merge it back to master only when the whole solution is complete

Shajan:
@karllessard i tried to rebase and then push the change to tensorflow:shared-saved-model i may not have permission to do so. Instead i made a pull request to the shared-saved-model branch.

karllessard:
Thanks @Shajan , it's ok I've merged it and sent you an invite to collaborate to this repository, I think after you accept it, I will be able to give you write access to the shared branch

Shajan:
Thanks @karllessard got the invite. Will wait for you to merge your changes and then add tests.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -94,6 +98,101 @@ private Loader(String exportDir) {
     private RunOptions runOptions = null;
   }
 
+  /**
+   * SignatureToNodeName finds the node names in the {@link Graph} corresponding to the
+   * input / output parameters of a <a
+   * href="https://www.tensorflow.org/api_docs/python/tf/function">tf.function</a>
+   */
+  public static final class SignatureToNodeName {
```
yzhuang, line 106: Read through this change, and I believe the following API design would be both cleaner and more intuitive to use.  Wdyt?

1. Update Loader and add `withSignature(String... signatures)`.
2. Update Loader.load() to also construct TfFunctions corresponding to the specified signatures.
3. SignatureToNodeName can become private and not exposed to end user, preferrably living inside TfFunction as private.
4. TfFunction should not have a reference to the session. Let SavedModelBundle manage the session (which it already does).  TfFunction can become a pure data class.
5. Add `SavedModelBundle.call(String signature, Map<String, Tensor<?>> inputs, Map<String, Tensor<?>> outputs)` to SavedModelBundle.   TfFunction can become private and not exposed to end user as well. 

What do you think about the above proposed API design?

Shajan: Not exposing SignatureToNodeName (3) sounds good
TtFunction not having reference to session (4) is good as well

Regarding (5), it is desirable to have a way to do repeated call to the same function. Having a TfFuction class allows for this. I am also thinking removing runtime check currently done with each call in the Tensor<?> call(Tensor<?>) method and do that once.

karllessard: Thanks for your proposal @yzhuang , some thoughts on it:

> SignatureToNodeName can become private and not exposed to end user, preferrably living inside TfFunction as private.

Looking at it, `getSignatureToNodeName()` can already be private in `SavedModelBundle` and `SignatureToNodeName` could be restricted at the default-package level.  I don't think it should be in `TfFunction` though as it maps all signatures while an instance of `TfFunction` is only mapped to one of them.

> TfFunction should not have a reference to the session. Let SavedModelBundle manage the session (which it already does). TfFunction can become a pure data class.

That would prevent though to just invoke `function.call(tensor)` to run the graph, which I personally like. While I understand that data classes have their lot of advantages, I'm not sure there is real gains for having `TfFunction` as one of them, especially that `Session` instances are thread-safe.

> Add SavedModelBundle.call(String signature, Map<String, Tensor<?>> inputs, Map<String, Tensor<?>> outputs) to SavedModelBundle. TfFunction can become private and not exposed to end user as well.

Again, I personally like the OO approach of letting users manipulating callable entities instead of having `SavedModelBundle` acting as a "service". Is the intention here is just to hide the `TfFunction` at the user level? Note that I was planning to reuse the same object to add new signatures to a model when exporting it (but I could also do it just with `SavedModelBundle` if we decide to take that direction).

karllessard:
@yzhuang , this SIG is having a community video chat where we will discuss about this topic, among other things. Are you available and interested to join?

Shajan:
@karllessard missed the meeting today. I see the decision to be 

> Confirmation: lets model the API with functions while still loading/saving session-centric graphs

I take it that we go with the current branch (after addressing open comments) and adding unit tests.

For unit tests: Here is a proposal
* Checkin savedmodel (created using python) as a resource
* [optionally] Add the python model + instructions (but not have the test run python code)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
yzhuang: Hi Karl and Shajan,

I have been busy with traveling, and didn't have time to follow up on this thread. Please feel free to take whatever you find helpful from my suggestions, and discard the rest. Thank you for taking time to go through my comments!

yzhuang: Hi @karllessard and @Shajan ,

Sorry for the late reply on your comment—just landed in Hong Kong :)

My suggestion to let SavedModel manage the session is not about thread safety, but about resource ownership. SavedModel currently “owns” the session and is responsible for closing the resource to avoid memory leaks.  If we create TFFunction objects holding references to the session, it complicate the resource ownership mental model.  For example, we will need to do things such as the below:

1. Perform reference counting on the session, and have the last “user” of the session close it.  This is user friendly, but introduces complexity and room for bugs.  OR 
2. Continue to let SavedModel own the session, but TFFunction objects needs to anticipate that it’s underlying resource (the session) can become defunct at anytime.   This can be counterintuitive to users.

> Is the intention here is just to hide the TfFunction at the user level? 

My intention was to not have to think about the resource ownership problem, and keep the resource ownership unchanged. The API of having TFFunctions be callable and exposed to users sounds great to me, and we need to think about the above 1 & 2 though. 

Thank you Karl and Shajan!

PR #103, opened 2020-08-24
Save models as functions

karllessard:
Ok I've modified a little this solution compared to the previous one, now `FunctionGraph` are auto-closeable and own the session/graph only if the function has been created before them. See examples provided in the `FunctionGraph` documentation for more details.

There are still a few things I need to look at before making this PR more official but I would like to get some feedbacks before continuing in that direction. 

In summary:

* `FunctionGraph` is the new (disposable) name for `TfFunction` as submitted previously by @Shajan 
* Users can create their graph in a "function style" (while we still don't support real function graphs but this API should be ready to accommodate it when we will come to it).
* Users can still define a graph and then convert it to a function (i.e. adding a signature to it). In this case, the function does not own the graph and releasing it will have no effect.
* Saving a model requires that one or more functions to be provided. The graph to be saved and the session to use to save the variables will be extracted from these functions (right now, since we can only export session-centric models, all functions saved should share the same session)

I'll provide more and better examples later, thanks!


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/FunctionGraph.java
```diff
@@ -0,0 +1,267 @@
...
+ * <p>A function can also invoke a
+ * <a href="https://www.tensorflow.org/api_docs/python/tf/function">tf.function</a>
+ * defined in a {@link SavedModelBundle}.
+ *
+ * <pre>{@code
+ * FunctionGraph myFunction = savedModelBundle.function("myFunctionSignatureName");
+ * Map<String, Tensor<?>> outputTensorMap = myFunction.call(inputTensorMap);
+ * }</pre>
+ */
+public class FunctionGraph implements AutoCloseable {
```
Craigacp, line 40: Maybe `FunctionSession` as it's a specialised session? That way it conceptually lives next to `EagerSession` and  `Session`, rather than next to `Graph`. I feel like this is much closer to a session than a graph.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/FunctionGraph.java
```diff
@@ -0,0 +1,267 @@
...
+    if (ownership != Ownership.NONE) {
+      session.close();
+      if (ownership == Ownership.GRAPH) {
+        graph.close();
+      }
+    }
+  }
+
+  private enum Ownership {
+    GRAPH, SESSION, NONE;
```
Shajan, line 253: Suggestion: Rename GRAPH to GRAPH_AND_SESSION

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -287,6 +301,17 @@ public synchronized void addInitializer(Op initializer) {
     return addGradients(null, new Output<?>[] {y}, x, null);
   }
 
+  public SaverDef saverDef() {
+    if (saverDef == null) {
+      synchronized (this) {
+        if (saverDef == null) {
+          saverDef = addVariableSaver(this);
```
Craigacp, line 308: `saverDef` needs to be declared volatile for this to work - http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

Craigacp: Or we make the whole method synchronized, which also seems fine as it's unlikely to be heavily contended.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -224,6 +240,18 @@ public static Loader loader(String exportDir) {
...
 
+  /**
+   * Export a saved model.
+   *
+   * <p/>Returns a <code>Exporter</code> object for setting configuration options before actually
+   * saving the model.
+   *
+   * @param exportDir the directory path containing a saved model.
+   */
+  public static Exporter exporter(String exportDir) {
```
Craigacp, line 251: Should we add a helper that takes an exportDir, a FunctionGraph and then performs all the necessary operations? Seems like there are a bunch of hoops to jump through if you just want to save a single `FunctionGraph`.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/SavedModelBundleTest.java
```diff
@@ -15,20 +15,38 @@
...
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.io.IOException;
 import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
+import jdk.nashorn.internal.codegen.FunctionSignature;
```
Craigacp, line 28: Wrong import.

karllessard:
> The change looks good.

Thanks @Shajan , I'll clean this up and merge it to our shared branch. It would be great if you can then validate that this merged solution still successfully cover your initial use case. If so, we will then make an official PR to the master branch out of it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/FunctionGraph.java
karllessard: Right now, it is true that all functions loaded from a saved model share the same graph and are just used to execute it with a given signature (therefore they are acting more as a session). Nonetheless, I still have a preference to preserve the `FunctionGraph` naming. 

If we ignore the saving part (which is limited due to the actual state of the C API), conceptually these functions also allow you to build your graphs, replacing the need of allocating explicitly a `Graph` instance, e.g.
```java
public class MyModel {

  // Function graph #1
  private static Signature addTwo(Ops tf) {
    Placeholder<TFloat32> input = tf.placeholder(TFloat32.DTYPE);
    Add<TFloat32> output = tf.math.add(input, tf.constant(2.0f));
    return Signature.builder("addTwo").input("x", input).output("y", output).build();
  }

  // Function graph #2
  private static Signature minusTwo(Ops tf) {
    Placeholder<TFloat32> input = tf.placeholder(TFloat32.DTYPE);
    Add<TFloat32> output = tf.math.sub(input, tf.constant(2.0f));
    return Signature.builder("subTwo").input("x", input).output("y", output).build();
  }

  public static void main(String args[]) {
    try (FunctionGraph f1 = FunctionGraph.create(MyModel::addTwo);
         FunctionGraph f2 = FunctionGraph.create(MyModel::minusTwo)) {
        ... // execute functions (saving both of them in the same model won't work for now)
    }
  }
}
```
Each function has its own graph and therefore if very coupled with that concept (where `EagerSession` has no graph at all). They will also appear as separate graphs when exporting them into a saved model the same way Python does it (in fact, Python saves each function as one or more "objects", which are then linked to their distinct graph in the function library).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
karllessard: I guess we could, what about making that endpoint available at the function level? Something like `function.save(exportDir)`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/FunctionGraph.java
karllessard: Another idea would be to call this class a `ConcreteFunction`. Concrete functions and functions are two very similar but distinct concepts in [TF Python](https://www.tensorflow.org/guide/intro_to_graphs). The former is a typed realization of a function graph while the latter is its polymorphic version, only acting as a facade (e.g. a function is eventually composed of multiple concrete functions, one for each type of operands the function has been called with).

In our scenario here, the function graphs are strongly typed and refer to a single graph, therefore more behave like a concrete function in the Python paradigm and it would make sense to name them after it.

We can probably support polymorphic functions too in the future, where we pass the type of the input tensors in parameter to the function builder. So we will still need to find a proper name for this concept, which will probably result in class encapsulating one or more `ConcreteFunction`. Maybe simply `PolymorphicFunction`?

Craigacp: My preference for a non-graph name is that `Graph`s aren't executable, and are missing necessary state because the `Variable`s live in the `Session`. So `FunctionGraph` is more like a `Session` because it contains all the necessary bits to execute. However I'm also fine with the `ConcreteFunction` name as that's just a new concept that people have to learn.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
Craigacp: I like that.

karllessard:
@googlebot I consent.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -0,0 +1,286 @@
...
+   * @return a new function
+   */
+  public static ConcreteFunction create(Signature signature, Session session) {
+    return new ConcreteFunction(signature, session.graph(), session, Ownership.NONE);
+  }
+
+  /**
+   * Returns the signature of this function
+   */
+  public Signature signature() {
```
yzhuang, line 152: Minor feature request: adding a toString() function to this ConcreteFunction class that prints a readable text representation of the signature would be really useful.

PR #105, opened 2020-08-24
add more cuda compute capabilities

### LINKED ISSUE #104: TF Java 2.3 failed on Tesla V100 GPU
Hi,

`tensorflow-core-api:0.2.0-SNAPSHOT` is failing on AWS EC2 instances with Tesla V100 GPUs.
 all the core-api tests failed with error code:
```
CUDA runtime implicit initialization on GPU:0 failed. Status: device kernel image is invalid
```

I have root caused the issue to be CUDA compute 7.0 is not enabled during compilation, running the following command and build from source again fixed the issue.
`export TF_CUDA_COMPUTE_CAPABILITIES=7.0`

Somehow in the release build of tensorflow-core-api, compute 7.0 is not enabled, but compute 3.5 and 7.0 should be the default capability of TF 2.3 according to [here](https://github.com/tensorflow/tensorflow/blob/master/configure.py#L39). The python packages and main repo built from source works fine without any modification. 

References:
https://github.com/tensorflow/tensorflow/issues/41132
https://github.com/tensorflow/tensorflow/commit/cf1b6b3dfe9ba82e805fddf7f4462b2d92fe550a

roywei:
should fix #104 

I'm not sure if this is the proper way to fix. It should be configured through [configure.py](https://github.com/tensorflow/tensorflow/blob/master/configure.py#L1003) during built from source.

I added these versions to match what `tensorflow-gpu` pip packages support, but will increase build time.

cc @karllessard @saudet

### DIFF
--- a/tensorflow-core/tensorflow-core-api/build.sh
+++ b/tensorflow-core/tensorflow-core-api/build.sh
@@ -24,7 +24,7 @@ fi
 
 if [[ "${EXTENSION:-}" == *gpu* ]]; then
     export BUILD_FLAGS="$BUILD_FLAGS --config=cuda"
-    export TF_CUDA_COMPUTE_CAPABILITIES=3.5,sm_52,sm_60,sm_61,7.0,7.5
+    export TF_CUDA_COMPUTE_CAPABILITIES=sm_35,sm_37,sm_52,sm_60,sm_61,7.0
     if [[ -z ${TF_CUDA_PATHS:-} ]] && [[ -d ${CUDA_PATH:-} ]]; then
         # Work around some issue with Bazel preventing it from detecting CUDA on Windows
         export TF_CUDA_PATHS="$CUDA_PATH"
--- a/tensorflow-core/tensorflow-core-api/build.sh
+++ b/tensorflow-core/tensorflow-core-api/build.sh
@@ -24,6 +24,7 @@ fi
 
 if [[ "${EXTENSION:-}" == *gpu* ]]; then
     export BUILD_FLAGS="$BUILD_FLAGS --config=cuda"
+    export TF_CUDA_COMPUTE_CAPABILITIES=3.5,sm_52,sm_60,sm_61,7.0,7.5
     if [[ -z ${TF_CUDA_PATHS:-} ]] && [[ -d ${CUDA_PATH:-} ]]; then
         # Work around some issue with Bazel preventing it from detecting CUDA on Windows
         export TF_CUDA_PATHS="$CUDA_PATH"


tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -24,6 +24,7 @@ fi
 
 if [[ "${EXTENSION:-}" == *gpu* ]]; then
     export BUILD_FLAGS="$BUILD_FLAGS --config=cuda"
+    export TF_CUDA_COMPUTE_CAPABILITIES=3.5,sm_52,sm_60,sm_61,7.0,7.5
```
Craigacp, line 27: Why the mix of `3.5` and `sm_52` formats? Why not `3.5` style or `sm_35` style for everything?

roywei: Trying to reduce the binary size according to https://github.com/tensorflow/tensorflow/blob/master/configure.py#L1003

>Each capability can be specified as "x.y" or "compute_xy" to include both virtual and binary GPU code, or as "sm_xy" to only include the binary code.\n Please note that each additional compute capability significantly increases your build time and binary size

Craigacp: Ok, so that means it'll fallback to 3.5 level compute if it hits a non 5.2 Maxwell or a non 6.0 Pascal. Not sure why it's got both Turing and Volta levels though. From the docs it suggests it uses `sm_35,sm_37,sm_52,sm_60,sm_61,7.0` - https://www.tensorflow.org/install/gpu as it says it's only got PTX for the latest supported architecture. Where did you get the list you used from?

roywei: From one of the committer comments when I research about the issue and I can't seem to find it now.
Updated to the same as official website doc. Thanks!

karllessard:
Thanks @roywei ! Triggering a CI build from your branch, let see if we are still able to build in time the GPU binaries for Linux on GitHub action with these new settings.

tensorflow-core/tensorflow-core-api/build.sh
saudet: The latest version of the official binaries for Python don't include PTX code. It's probably not a good idea to follow what they are doing there. They'll probably revert that change once they get flooded with enough duplicate issues about that.

roywei:
Right now this is what we need for [Deep Java Library](https://djl.ai) to work on GPU. We can reduce the versions to 3.5 and 7.0 if it brings too much burden to CI, DJL can build from source using our machines.

I agree that TF might revert that change if they see more of the same issues. We can keep the issue open and revert this change if TF also reverted theirs.

tensorflow-core/tensorflow-core-api/build.sh
Craigacp: We should poke someone to update the docs page to note they've removed the PTX from the binaries (though I agree it'll be back for compatibility reasons).

roywei:
Seems CI taking too long, let me know what arch I should change it to. If you guys can report this to someone at tensorflow, we can also leave it for now.

karllessard:
So after two attempts, the GitHub build was failing with this new set of capabilities (it was a bit expected since we were already so closed to the 6 hours timeout with the previous/more conservative settings). 

I'm not sure what we can do to improve the situation right now but I had a discussion with Joana (@google) today and it seems that our inquiry for GCP machines is still moving forward, we can then retest it. We'll keep the PR opened but I've added a label "On hold" until there is more progress on this side.

saudet:
@roywei If 3.5 and 7.0 with PTX works for you, I'd be OK with this:
```
export TF_CUDA_COMPUTE_CAPABILITIES="3.5,7.0"
```

roywei:
Yes, thanks! I see it's been updated, closing.

PR #106, opened 2020-09-01
Learning rate

JimClarke5:
This PR requires PR "Initial checkin of Keras Optimzers and helper classes" to be merged first.

Added changeable learning rate to Optimizers. This was done by adding a Placeholder for the learning rate, a Tensor to track the actual learning rate, and adding a Map to map the Placeholder to the Tensor that can be used to "feed" the runner.
Test Sessions were modified to accept a "FeedDict" Map to popullate the feed() of the runner.


deansher (changes requested):
A couple of quick observations as I start reading this code.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DataType.java
```diff
@@ -43,27 +52,96 @@
...
+        return TBool.DTYPE;
+      case TString.NAME:
+        return TString.DTYPE;
+      default:
+        throw new IllegalArgumentException(String.format("%s is an unknown DataType", name));
+    }
+  }
+
+  /** Returns true if this data type represents a floating point type */
+  public boolean isFloating() {
```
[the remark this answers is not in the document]
deansher: Never mind, in this context -- I see in my local diff that this delta is unrelated to this PR. I'll raise this as an issue.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/AdaDelta.java
```diff
@@ -15,88 +15,169 @@
...
 
+  /**
+   * Creates an AdaDelta Optimizer
+   *
+   * @param graph the TensorFlow Graph
+   * @param name the name for this Optimizer (defaults to 'Adadelta')
+   * @param learningRate the learning rate
+   */
   public AdaDelta(Graph graph, String name, float learningRate) {
     this(graph, name, learningRate, 0.95f, 1e-8f);
```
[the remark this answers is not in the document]
deansher: Never mind, in this context.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -56,70 +63,68 @@ protected Optimizer(Graph graph) {
...
    * Builds an optimizer for the supplied graph.
    *
+   * <p>Uses the name from {@link Optimizer#getOptimizerName()} to name the operations.
+   *
    * @param graph The graph to optimize.
-   * @param name The base name for the operations.
+   * @param learningRate the learning rate.
    */
-  protected Optimizer(Graph graph, String name) {
+  protected Optimizer(Graph graph, float learningRate) {
```
Craigacp, line 77: Can we have both of these constructors call into the `Optimizer(Graph,float,String)` one?

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -56,70 +63,68 @@ protected Optimizer(Graph graph) {
...
+    this.slots = new HashMap<>();
+    this.globals = new ArrayList<>();
+    setLearningRate(learningRate);
+  }
+
   public static String createName(Output<? extends TType> variable, String slotName) {
     return variable.op().name() + "-" + slotName;
   }
 
-  /**
```
Craigacp, line 93: Why'd the Javadoc go away?

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -305,41 +350,20 @@ private Options() {}
     }
   }
 
-  /**
```
Craigacp, line 308: Where'd the javadoc go?

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -280,12 +271,68 @@ protected Op finish(List<Op> updateOperations, String name) {
...
    */
   public abstract String getOptimizerName();
 
+  /**
+   * Sets the learning rate
+   *
+   * @param learningRate the learning rate
+   */
+  public final void setLearningRate(float learningRate) {
+    if (this.learningRatePlaceholder == null) {
```
Craigacp, line 286: Everything seems to have grown a `this` reference. I don't think that's particularly necessary in these methods, as the argument could be `newLearningRate` rather than `learningRate` and then there is no aliasing.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: Being that Optimizer is abstract, we really only need one constructor, `protected Optimizer(Graph graph, String name, float learningRate)` . Of course, we would have to handle a `null` name, with something like:
`this.tf = Ops.create(graph).withName(name == null? getOptimizerName() : name);`

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: I am not sure what happened. I had a local copy that I saved and it was there, so will add it back in.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: I do this out of habit. I can easily change it as you suggest.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: Update pushed

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: I have added it back in. Update pushed

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: changed `setLearningRate` to `setLearningRate(float newLearningRate)`, removed spurious` this.`.

Update pushed

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -280,20 +321,20 @@ protected Op finish(List<Op> updateOperations, String name) {
   /**
    * Sets the learning rate
    *
-   * @param learningRate the learning rate
+   * @param newLearningRate the new earning rate
```
Craigacp, line 324: typo - "earning"

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -323,13 +364,15 @@ public float getLearningRate() {
...
   }
 
+  /** {@inheritDoc}  */
   public void close() {
     // close the learningRate Tensor if it exists.
-    if (this.feedMap != null) {
-      this.feedMap.get(this.learningRatePlaceholder).close();
+    if (learningRateTensor != null) {
+      learningRateTensor.close();
+      learningRateTensor = null;
```
Craigacp, line 375: This still leaves a valid reference in the `feedMap` though the tensor it refers to is closed.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: CTORS have been changed

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: OK

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
JimClarke5: OK, but I am starting to question in my mind, whether we should use `PlaceHolder` within the `Optimizer` itself, but rather store the `learningRate` as an `Operand`. This allows the user of the `Optimizer` to define the learning rate as needed, For example,  a `Constant`, `Placeholder`, `Variable`, or an `Operand` such as what may be defined in a scheduler.  If we did this, it would be up to the `Optimizer` user, to make the change to the learning rate, via one of the mechanisms just outlined as an `Operand`.  Also, if we did this general approach, we could get rid of `feedMap`.  

In reality, `feedMap` is just a pair tying a `Placeholder` to a `Tensor`.  When you use `feed` in `Session.Runner`, each member of the pair is passed as two distinct parameters.

Craigacp: So I think that having it be an operand is probably reasonable, as then the learning rate scheduler can own it, but I think the current approach is the most flexible (as it means users don't have to do the scheduler calculation in TensorFlow if they don't want to). If it's an operand then we'd need to implement a variety of learning rate schedulers, and I'm not sure if that's in the set of things you've already done or not.

JimClarke5: OK, I'll leave it as is with the Placeholder. 

I do have scheduler classes that build their own graphs for calculating the new learning rate based on the current training step. This is usually done be adding a `Variable` into that generated graph, and updating it for each training step. However,  I was going to add those in another PR later. For that feature,  I would need to change `Optimizer`  to add a `setLearningRateOperand` method. I already have a `getLearningRateOperand` method that currently returns the Placeholder for the `Optimizer` subclasses to use when building their graphs. My plan was to have `getLearningRateOperand`  return either the internal `Placeholder` or the new `Operand` based on how the `Optimizer` was constructed. Should I add that method now or wait?

Craigacp: May as well add it now if you've already got it written. It'll change the logic in here anyway and we may as well see all of it in one go.

JimClarke5: You want me to add the schedules?. They are currently in package `org.tensorflow.framework.optimizers.schedules` and include `ExponentialDecay`, `InverseTimeDecay`, `PiecewiseConstantDecay` and `PolynomialDecay`.

Craigacp: Erm, probably not, lets try to keep the PRs small-ish.

JimClarke5: I have added `Operand<TFloat32> learningRateOperand` to `Optimizer` and kept `Placeholder<TFloat32> learningRatePlaceholder`. I also added a constructor to set the `Operand` and kept the one that uses a  `float` for learning rate. 
`getLearningRateOperand` now returns the `learningRatePlaceholder ` if it exists, otherwise it returns the `learningRateOperand`.

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/RMSPropTest.java
```diff
@@ -197,6 +215,319 @@ public void testDense() {
...
+      float momentum = (float) testParamValue[2];
+      float epsilon = (float) testParamValue[3];
+      boolean centered = (boolean) testParamValue[4];
+
+      try (TestSession session = TestSession.createTestSession(tfMode);
+          RMSProp instance =
+              new RMSProp(session.getGraph(), learningRate, decay, momentum, epsilon, centered)) {
+        Ops tf = session.getTF();
+        session.setEpsilon(1e-2f);
+        float[] var0_init = {1.0F, 2.0F};
```
Craigacp, line 391: Please switch the python style variable names to camelCase.

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/AdamTest.java
```diff
@@ -190,25 +185,293 @@ public void testBasic() {
...
     FloatNdArray mul1 = ND.mul(v, beta);
     FloatNdArray squareG = ND.square(gT);
     FloatNdArray mul2 = ND.mul((1 - beta), squareG);
     return ND.add(mul1, mul2);
   }
 
   private FloatNdArray calculateParam(
-      FloatNdArray param, float lrT, FloatNdArray m, FloatNdArray v, float epsilon) {
-    //  param - lrT * mT / (np.sqrt(vT) + epsilon)
+      FloatNdArray param, float lr_t, FloatNdArray m, FloatNdArray v, float epsilon) {
```
Craigacp, line 471: Switch python style name to camelCase.

Craigacp (changes requested):
Just a few snake_case variable names in the tests that need converting to camelCase, and then I'll merge this in.

JimClarke5:
This PR requires some rework due to #174, `Type Refactor`,  Should we close this PR and open a new one later? Also, this PR is closely related to `callbacks`, so I suggest we revisit this one after `callbacks` is done. My current plan is #180, Metrics Phase1, followed by Metrics Phase2, then Model/Layers Phase 1, then `callbacks`, then this PR can be revisited.

Craigacp:
> This PR requires some rework due to #174, `Type Refactor`, Should we close this PR and open a new one later? Also, this PR is closely related to `callbacks`, so I suggest we revisit this one after `callbacks` is done. My current plan is #180, Metrics Phase1, followed by Metrics Phase2, then Model/Layers Phase 1, then `callbacks`, then this PR can be revisited.

Whatever you think is easiest is fine by me.

JimClarke5:
I am closing for now, and will reopen after we get further along on Model.

PR #108, opened 2020-09-09
Updated README.md to describe our API layers.

deansher:
Updated the list of API layers in the README to include the new Keras layer. Spelled out the purpose and audience for each layer, both to explain our current direction and to prepare for a discussion of whether the Keras and framework layers should in fact be separate.


deansher:
@googlebot I signed it!

karllessard (approved):

PR #110, opened 2020-09-10
Upgrade JavaCPP to just released version 1.5.4

saudet:


karllessard (approved):

karllessard:
@saudet , any idea how we can fix the build? There is a recurrent Cuda package error now on the GPUs platforms.

saudet:
Looks like the CUDA Docker images from NVIDIA are broken? We might want to switch to some other way of installing CUDA...

karllessard:
Just out of curiosity, I'll try to build on top of Cuda 10.2 docker image to see if that works...

karllessard:
Mmh, worked on Cuda 10.2... https://github.com/tensorflow/java/runs/1101759245?check_suite_focus=true

 I don't know if we can have issues at runtime, but according to @frankfliu, it will fallback anyway looking for Cuda 10.1 so this should be impact-less to our users. Should I commit?

saudet:
No, it's probably not going to work with either CUDA 10.1 or 10.2...

PR #111, opened 2020-09-11
Try to build with Cuda 10.2 on Linux

karllessard:
DO NOT MERGE, this is just a test for now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
   linux-x86_64:
     if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'CI build')
     runs-on: ubuntu-latest
-    container: nvidia/cuda:10.1-cudnn7-devel-centos7
+    container: nvidia/cuda:10.2-cudnn7-devel-centos7
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


roywei:
Hi, I was also looking into cuda 10.2 support. Even if I installed cuda 10.2 on a local machine, and successfully built from source. TF2.3 (dso_loader.cc) is still trying to load cuda 10.1 as it's hard coded. Need to symlink cuda10.2 as 10.1.

karllessard:
Hi @roywei , 

As we have discussed with @frankfliu during our last meeting, we try to keep the build of the native libraries as close as what is being delivered in the Python wheels and try to avoid tampering with the source code of the main TF repository, since it would be hard for us to guarantee any stability by doing so.

We also have discussed starting to have additional "experimental" builds, in which case no stability will be guaranteed and users would be well aware of it. Cuda 10.2-enabled can be one of them. But right now, we already have a lot of issues with our build platform (GitHub actions) and adding more stress on it is probably not something that we will look at in the near future. So I would say that having custom builds (where you can patch the code as you want) is probably still the best approach until TensorFlow supports Cuda 10.2 officially.

There is already this issue reported in TensorFlow core where you can share your ideas as well: https://github.com/tensorflow/tensorflow/issues/38194

PR #118, opened 2020-09-25
Prepare scripts for release deployment

karllessard:
This PR is best reviewed by commits. There are two commits:

[First one](https://github.com/tensorflow/java/commit/0f9ff089c92ba7d20d1084edac8e158f511acd34) is only fixing a few javadoc issues that doclint was complaining about.

[Last one](https://github.com/tensorflow/java/commit/fa362c5e6ba266f005dceaccdd43a39a173cabbf) adds the scripts for releases and adjust the CI workflow. So basically the solution is the same as in TF1.x, where the native artifacts gets compiled by the build servers and uploaded on an external storage (AWS S3 in this case, as it was super easy for me to setup). For platforms that we cannot build on GitHub actions, I'll do it manually and upload their native artifacts in the same S3 bucket.

Then locally, we fetch all these artifacts and package, sign and deploy them all on Sonatype using a Docker image for a more hermetic build, by calling `release.sh`. No need to deploy on Bintray, as it should take care automatically to proxy Maven Central and expose the artifacts on JCenter after.

I'm sure this procedure is not final and will evolve in time but at least we will have something that works for our first release that I plan to deploy next week.

[Note: This PR is still a draft as I'm waiting to complete the whole roundtrip to validate that the solution works as expected]


saudet:
We don't need AWS S3 or anything else than Nexus. Simply staging everything in Nexus works fine the way it is right now.

The problem remains that, at the moment, we're not able to compile everything with GitHub Actions!

karllessard:
@saudet: do you need to sign your artifacts before pushing them on OSSRH staging?

saudet:
No, we don't, just when we close and release it.

saudet:
We need to add this plugin for that though with auto close and release
disabled, which we need to do when deploying from multiple machines anyway:
https://central.sonatype.org/pages/apache-maven.html#nexus-staging-maven-plugin-for-deployment-and-release

karllessard:
Sounds good, I'll look at it. I do prefer also only dealing with only one external repository but I was under the wrong impression that signing would have been required in our builders.

karllessard:
Closing for now, will reopen once the new version has been tested

karllessard:
@saudet , we can't download artifacts from a staging repository unless it has been closed but we can't close a staging repository unless all its artifacts have been signed. So it is not possible to deploy the native libraries first on Nexus then sign and redeploy them from a local process, like I was doing when using S3.

saudet:
Yes, it's possible to use a staging repository to download artifacts. Make sure you have something like this in `~/.m2/settings.xml`:
```xml
<settings>
  <profiles>
    <profile>
      <id>staging</id>
      <repositories>
        <repository>
          <id>sonatype-nexus-staging</id>
          <name>Sonatype Nexus Staging</name>
          <url>https://oss.sonatype.org/service/local/staging/deployByRepositoryId/orgtensorflow-1087/</url>
          <releases>
            <enabled>true</enabled>
          </releases>
          <snapshots>
            <enabled>false</enabled>
          </snapshots>
        </repository>
      </repositories>
    </profile>
  </profiles>
  <activeProfiles>
    <activeProfile>staging</activeProfile>
  </activeProfiles>
  <servers>
 ...
  </servers>
</settings>
```
We can also put that stuff in the pom.xml file with properties and things for convenience like this:
https://github.com/bytedeco/javacpp/blob/master/platform/pom.xml#L332

PR #121, opened 2020-09-29
Add Activation classes in framework

JimClarke5:
Added the activations under `framework/activations`. Each activation class extends the base class `Activation`. 
This PR is not dependent on any other PRs.


tensorflow-framework/src/test/java/org/tensorflow/framework/activations/TanhTest.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+          });
+  }
+
+  /** Test of Tanh call method. */
+  @Test
+  public void testCall__Float() {
+    float[] input = {1, 2, 3, 4, 5, 6, 7, 8};
+    float[] expected = {
+      0.7615942F, 0.9640276F, 0.9950547F, 0.9993292F, 0.99990916F, 0.99998784F, 0.99999833F, 1.0F
```
Craigacp, line 67: This should probably test the negative range too, to make sure we get negative values out (though I'm sure we will).

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,165 @@
...
+ * @param <T> the data type of the result
+ */
+public class ReLU<T extends TNumber> extends Activation<T> {
+
+  private static final float ALPHA_DEFAULT = 0.0f;
+  private static final Float MAX_VALUE_DEFAULT = null;
+  private static final float THRESHOLD_DEFAULT = 0.0f;
+
+  private final float alpha;
+  private final Float maxValue;
```
Craigacp, line 69: Can we use `Double.NaN` as a sentinel here rather than a boxed value? Or a boolean?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /** Creates the abstract class for an Activation */
+  public Activation() {}
+
+  /**
+   * Creates the abstract class for an Activation
+   *
+   * @param tf the TensorFlow Ops
+   */
+  public Activation(Ops tf) {
```
Craigacp, line 40: What are the situations where we'd want to construct an Activation without an ops and then set it in manually? Seems like we'd always need one, and it could be set to a final field in the constructor.

Craigacp (changes requested):
Looks like the initializers code made it's way in here too, I didn't review that. There is a bug in HardSigmoid, and a few more minor comments to discuss.

PR #123, opened 2020-10-01
Add Activations

JimClarke5:
This is a redone PR to make sure the `Activations` branch did not rely on the `Initializers` branch.

I have removed all the no-arg CTORs. The original motivation was that I found previously in looking at Keras layers is that  objects like “activations”, “loss” and “metric"  most likely would  need to be constructed
before the TensorFlow Ops is available within the model, hence I make constructors without Ops and used` setTF()` so the `Model` could set it later. The model itself hides `Ops tf` from the user. We have decided to address this requirement later when we start looking at `Model`.

I fixed `HardSigmoid`, as I forgot to apply a `clipByValue` operation on the final result. Also fixed the Unit test.

I have verified all the Unit tests against their TF Python corresponding methods.

For `Softmax`, we eliminated the `IllegalArgmentException` on 1D input that was in the Keras version.

In `Swish`, there is a comment on not handling the `grad` argument from the corresponding Python class. This is a Python optimization and will not be implemented in Java, at least at this time.


tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,143 @@
...
+ *
+ *     // With maxValue = 5
+ *     relu = new ReLU&lt;&gt;(tf, ReLU.ALPHA_DEFAULT, 5f, ReLU.THRESHOLD_DEFAULT);
+ *     result = relu.call(input);
+ *     // result is [0.f, 0.f, 0.f, 5.f, 5.f]
+ *
+ *     // With threshold = 5
+ *     relu = new ReLU&lt;&gt;(tf, ReLU.ALPHA_DEFAULT, ReLU.MAX_VALUE_DEFAULT, 5f);
+ *     result = relu.call(input);
+ *     // result is [-0.f, -0.f,  0.f,  0.f, 10.f]
```
Craigacp, line 57: How does this emit negative zero? Could we catch that and emit a positive zero?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
```diff
@@ -0,0 +1,92 @@
...
+    Shape shape = input.asOutput().shape();
+    int numDimensions = shape.numDimensions();
+    if (numDimensions == 2) {
+      return tf.nn.softmax(input);
+    } else {
+      Operand<T> e =
+          tf.math.exp(
+              tf.math.sub(input, tf.reduceMax(input, tf.constant(axis), ReduceMax.keepDims(true))));
+      Operand<T> s = tf.reduceSum(input, tf.constant(axis), ReduceSum.keepDims(true));
+      return tf.math.div(e, s);
```
Craigacp, line 89: Does this work on a vector now the check has been removed?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Linear.java
```diff
@@ -0,0 +1,50 @@
...
+=======================================================================*/
+package org.tensorflow.framework.activations;
+
+import org.tensorflow.Operand;
+import org.tensorflow.op.Ops;
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Linear activation function.
+ *
```
Craigacp, line 23: This should probably note that the linear activation function is the identity function.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftplusTest.java
```diff
@@ -0,0 +1,96 @@
...
+        session.evaluate(expected, result);
+      }
+  }
+
+  /** Test of Softplus call method */
+  @Test
+  public void testCall__Double() {
+    double[] input = {1, 2, 3, 4, 5, 6, 7, 8};
+    double[] expected = {
+      1.3132616875182228, 2.1269280110429727, 3.048587351573742,
```
Craigacp, line 84: This should test that a negative or near zero value behaves appropriately rather than just the linear region.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftmaxTest.java
```diff
@@ -0,0 +1,112 @@
...
+        Softmax<TFloat32> instance = new Softmax<>(tf);
+        Operand<TFloat32> result = instance.call(tf.constant(input));
+        session.evaluate(tf.constant(expected), result);
+      }
+  }
+
+  /** Test of Softmax method, of class Activations. */
+  @Test
+  public void testSoftmax_Ops_Operand_Double() {
+    double[][] input = {{1, 2, 3, 4}, {5, 6, 7, 8}};
```
Craigacp, line 82: It would be nice to have a test for the 3-tensor and a vector input as they have different code paths.

Craigacp (changes requested):
Few small things, plus it would be nice to rename the test methods to get rid of the underscores and use standard Java naming conventions.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: I am not sure who is the culprit here. It seems to be the result of the multiply operation of a negative number * zero always produces `-0`, maybe this is a Pyhtonism.
Here is the same operation in just `Numpy`:
```
a = np.array([-10., -5., 0.0, 5., 10.], dtype=np.float32)
c = np.greater(a, 5.)
print(c);
[False False False False  True]

c = np.cast['float32'](c)
print(c)
[0. 0. 0. 0. 1.]

d = a * c
print(d)
[-0. -0.  0.  0. 10.]

z = -1. * 0.
print(z)
-0.0

```
TensorFlow Python and TensorFlow Java produce the same `-0 `result.

If you don't want this behavior, we could add zero to the final result, because `-0 + 0 = 0`.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
JimClarke5: Yes it works, I have added tests for 1D and 3D to make both path are covered.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftplusTest.java
JimClarke5: I have modified the input to use a combination of positive and negative values.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftmaxTest.java
JimClarke5: I have added a test for 3D input

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/ReLUTest.java
```diff
@@ -82,7 +82,7 @@ public void testCall__Long() {
 
   /** Test of ReLU call method */
   @Test
-  public void testCall__Float16() {
+  public void testCall_Float16() {
```
Craigacp, line 85: Still snake_case, looks like the find replace didn't get all of them.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftmaxTest.java
```diff
@@ -109,4 +111,38 @@ public void testSoftmax_Ops_Operand_Double_Negative() {
         session.evaluate(tf.constant(expected), result);
       }
   }
+
+  /** Test of Softmax method, of class Activations. */
+  @Test
+  public void testSoftmax_1D() {
```
Craigacp, line 117: Swap the snake_case for camelCase.

Craigacp (changes requested):
Tiny things to do with casing.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/ReLUTest.java
JimClarke5: Fixed

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SoftmaxTest.java
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,143 @@
...
+    DataType<T> dataType = input.asOutput().dataType();
+    Operand<T> negativePart = null;
+    boolean clipMax = !Float.isNaN(maxValue);
+    if (alpha != 0) {
+      if (Float.isNaN(maxValue) && threshold == 0) {
+        // TODO LeakyRelu is not in tf.nn ????
+        return LeakyRelu.create(tf.scope(), input, LeakyRelu.alpha(alpha));
+      }
+      if (threshold != 0) {
+        negativePart =
```
deansher, line 111: What would you think about moving both the declaration and the initialization of `negativePart` down to immediately before it is used? I realize that would require a change to keep `input` as the original input instead of reusing it for intermediate calculations, but personally I'd see that as a feature not a bug.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Rectified Linear Unit(ReLU) activation.
+ *
+ * <p>With default values, this returns the standard ReLU activation: <code>max(x, 0)</code>, the
+ * element-wise maximum of 0 and the input tensor.
+ *
+ * <p>Modifying default parameters allows you to use non-zero thresholds, change the max value of
+ * the activation, and to use a non-zero multiple of the input for values below the threshold.
+ *
```
deansher, line 32: Could we adapt this part of the Keras documentation?

**alpha**: A float that governs the slope for values lower than the threshold.

**max_value**: A float that sets the saturation threshold (the largest value the function will return).

**threshold**: A float giving the threshold value of the activation function below which values will be damped or set to zero.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
deansher: Perhaps rename `negativePart` -> `amountBelowThreshold`? For one thing, `negativePart` is vague to me. For another, it's always positive.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,143 @@
...
+    if (clipMax) {
+      Operand<T> lmaxValue = tf.dtypes.cast(tf.constant(maxValue), dataType);
+      Operand<T> zero = tf.dtypes.cast(tf.constant(0), dataType);
+      input = tf.clipByValue(input, zero, lmaxValue);
+    }
+
+    if (alpha != 0.) {
+      input =
+          tf.math.sub(
+              input, tf.math.mul(tf.dtypes.cast(tf.constant(alpha), dataType), negativePart));
```
deansher, line 139: I'm finding it extremely difficult to think through the interactions of all these options. E.g. if we calculate `negativePart` before applying `clipMax`, and then we clip the input, and then we subtract a fraction of `negativePart` from the input, did we do the right thing? How about if there was also a `threshold`?

Perhaps either there's some reasoning we could document up front, or somehow there's a way to order the computation and name the intermediate results so it's clearly correct?

Perhaps the problem is that there are neither documentation, nor examples, nor tests that show how these parameters are supposed to interact.

Or if all of this accurately mimics the Python implementation, we leave it for now while opening a ticket for some newbie like me to come back and tidy it up?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,143 @@
...
+      // if no threshold, then can use nn.relu6 native TF op for performance
+      input = tf.nn.relu6(input);
+      clipMax = false;
+    } else {
+      input = tf.nn.relu(input);
+    }
+    if (clipMax) {
+      Operand<T> lmaxValue = tf.dtypes.cast(tf.constant(maxValue), dataType);
+      Operand<T> zero = tf.dtypes.cast(tf.constant(0), dataType);
+      input = tf.clipByValue(input, zero, lmaxValue);
```
deansher, line 133: Is zero still the right lower clip if there's a threshold?

deansher (changes requested):

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Swish.java
```diff
@@ -0,0 +1,68 @@
...
+ *                                        {-20, -1.0, 0.0, 1.0, 20});
+ *     Swish&lt;TFloat32&gt; swish = new Swish&lt;&gt;(tf);
+ *     Operand&lt;TFloat32&gt; result = swish.call(input);
+ *     // result = [-4.1223075e-08f, -2.6894143e-01f,  0.0000000e+00f,
+ *     //          7.3105860e-01f,  2.0000000e+01f ]
+ *
+ * </pre>
+ *
+ * @param <T> the data type of the activation
+ * @see <a href="https://arxiv.org/abs/1710.05941">Ramachandran et al., 2017</a>.
```
deansher, line 41: trailing period freaks out Maven's javadoc generation

deansher (changes requested):

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>ELUs have negative values which pushes the mean of the activations closer to zero. Mean
+ * activations that are closer to zero enable faster learning as they bring the gradient closer to
+ * the natural gradient. ELUs saturate to a negative value when the argument gets smaller.
+ * Saturation means a small derivative which decreases the variation and the information that is
+ * propagated to the next layer.
+ *
+ * <p>Example Usage:
+ *
+ * <pre>
+ *     ELU&lt;TFloat32&gt; input = ...;
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: I am not sure if this is used here as a math term, but I found [Wolfram MathWorld - Negative Part](https://mathworld.wolfram.com/NegativePart.html). Perhaps @Craigacp  and @karllessard  might want to comment.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: I am not sure there is an advantage in Java to specify a _"A float"_ in the JavaDoc comment, as it is already defined as a `float` in the method signature. I could see in Python why this may be necessary.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: This was ported mainly line by line from the Python ReLU method in Keras backend. Java does not have operator overload, so it is a little more cumbersome code than the original python. For example, Python 

`x -= alpha * negative_part`

translates to TF Java as:

```
input =
          tf.math.sub(
              input, tf.math.mul(tf.dtypes.cast(tf.constant(alpha), dataType), negativePart));
```

There may be some benefit to making this in more individual statements, but TF Java in  Graph mode is very difficult to actually debug.  When I have to debug it, I use special code to print the values using a try(Tensor...) logic.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/SELU.java
```diff
@@ -0,0 +1,74 @@
...
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Scaled Exponential Linear Unit (SELU).
+ *
+ * <p>The Scaled Exponential Linear Unit (SELU) activation function is defined as:
+ *
+ * <ul>
+ *   <li><code>if x > 0: return scale * x</code>
+ *   <li><code>if x < 0: return scale * alpha * (exp(x) - 1)</code>
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/SELU.java
```diff
@@ -0,0 +1,74 @@
...
+ * </ul>
+ *
+ * <p>where <code>alpha</code> and <code>scale</code> are pre-defined constants (<code>
+ * alpha=1.67326324</code> and <code>scale=1.05070098</code>).
+ *
+ * <p>Basically, the SELU activation function multiplies <code>scale</code> (> 1) with the output of
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Sigmoid.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.tensorflow.Operand;
+import org.tensorflow.op.Ops;
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Sigmoid activation. <code>sigmoid(x) = 1 / (1 + exp(-x))</code>.
+ *
+ * <p>Applies the sigmoid activation function. For small values (<-5), <code>sigmoid</code> returns
+ * a value close to zero, and for large values (>5) the result of the function gets close to 1.
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -0,0 +1,103 @@
...
+ *     ELU&lt;TFloat32&gt;elu = new ELU&lt;&gt;(tf, 2.0f);
+ *     Operand&lt;TFloat32&gt; result = elu.call(input);
+ * </pre>
+ *
+ * @param <T> the data type of the activation
+ * @see <a href="https://arxiv.org/abs/1511.07289">Clevert et al, 2016, Fast and Accurate Deep
+ *     Network Learning by Exponential Linear Units (ELUs)</a>.
```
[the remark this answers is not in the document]
JimClarke5: I changed them to &#46;.  IntelliJ's javadoc formatter handled them

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/HardSigmoid.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * <p>A faster approximation of the sigmoid activation.
+ *
+ * <p>Defined as:
+ *
+ * <ul>
+ *   <li><code>if x < -2.5: return 0</code>
+ *   <li><code>if x > 2.5: return 1</code>
+ *   <li><code>if -2.5 <= x <= 2.5: return 0.2 * x + 0.5</code>
+ * </ul>
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/SELU.java
```diff
@@ -0,0 +1,74 @@
...
+ * <p>The values of <code>alpha</code> and <code>scale</code> are chosen so that the mean and
+ * variance of the inputs are preserved between two consecutive layers as long as the weights are
+ * initialized correctly (see {@link org.tensorflow.framework.initializers.LeCun} with Normal
+ * Distribution) and the number of input units is "large enough"
+ *
+ * <p><b>Notes: </b> To be used together with the {@link
+ * org.tensorflow.framework.initializers.LeCun} initializer with Normal Distribution.
+ *
+ * @param <T> the data type of the activation
+ * @see <a href="https://arxiv.org/abs/1706.02515">Klambauer et al., 2017</a>.
```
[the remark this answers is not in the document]
JimClarke5: Removed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
```diff
@@ -0,0 +1,92 @@
...
+import org.tensorflow.op.core.ReduceMax;
+import org.tensorflow.op.core.ReduceSum;
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Softmax converts a real vector to a vector of categorical probabilities.
+ *
+ * <p>The elements of the output vector are in range (0, 1) and sum to 1.
+ *
+ * <p>Each vector is handled independently. The <code>axis</code>>argument sets which axis of the
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
```diff
@@ -0,0 +1,92 @@
...
+   * @param input the input tensor
+   * @return The operand for the activation
+   */
+  @Override
+  public Operand<T> call(Operand<T> input) {
+    if (!input.asOutput().dataType().isFloating()) {
+      throw new IllegalArgumentException(
+          "Must be a Floating Point DataType: " + input.asOutput().dataType());
+    }
+    Shape shape = input.asOutput().shape();
```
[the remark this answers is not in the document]
JimClarke5: There are 2 shapes, which one, `asOutput()` or `asTensor()`?   They may be different at different times.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Swish.java
JimClarke5: Removed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
deansher: Good point, that Wolfram definition does show why `negativePart` was used in the first place. That said, I'd still advocate that in the presence of `threshold`, it's a confusing name.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
deansher: For sure! But the text like "the threshold value of the activation function below which values will be damped or set to zero" may be helpful.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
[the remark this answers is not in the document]
deansher: :-) oh!

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
deansher: My difficulty with the code isn't with this sort of tactic. It's with the overall flow of the computation and the variable names used. Given that it's transliterated from the Python and that the same tests pass, seems good to leave it for another day.

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/ELUTest.java
```diff
@@ -0,0 +1,103 @@
...
+              Operand<TInt32> input = tf.constant(new int[] {1, 2, 3, 4, 5});
+              ELU<TInt32> instance = new ELU<>(tf);
+              Operand<TInt32> result = instance.call(input);
+            }
+          });
+  }
+
+  /** Test of ELU call method */
+  @Test
+  public void testCallFloat() {
```
deansher, line 64: Perhaps also consolidate this sort of test logic into a helper class? E.g.
~~~
assertFloat32Output(tf -> new ELU<Float32>(tf),
            {1f, -0.86466473f, 3f, -0.9816844f, -0.63212055f, 2f, -0.95021296f, 4f},
            {1, -2, 3, -4, -1, 2, -3, 4})

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /** The TensorFlow Ops */
+  protected Ops tf;
+
+  /**
+   * Creates the abstract class for an Activation
+   *
+   * @param tf the TensorFlow Ops
+   */
+  protected Activation(Ops tf) {
```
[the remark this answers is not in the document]
karllessard: We've already discussed about adding the `@Nullable` annotation but this is not part of the standard JDK and we need to bring additional dependencies, which we try to avoid as much as possible to avoid conflict with other version of that same dependency that might have been imported by the user. 

That's being said, when we discussed about it, it was related to the Core API. I'm OK though to bring more dependencies in the framework if we need to.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * Gets the TensorFlow Ops
+   *
+   * @return the TensorFlow Ops
+   */
+  protected Ops getTF() {
+    return this.tf;
+  }
```
karllessard, line 59: Probably a way to avoid these `setTF/getTF` would be to push created `Ops` instances on a stack somewhere... but this is out of scope of this PR for sure

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Gets the calculation operation for the activation.
+   *
+   * @param input the input tensor
+   * @return The operand for the activation
+   */
+  public abstract Operand<T> call(Operand<T> input);
```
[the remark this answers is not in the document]
karllessard: I agree with annotating `@Nullable` fields in the framework, especially that other JVM languages like Kotlin handle nullable values in a very specific way. Should we do it in this PR or go through all our classes later?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public Operand<T> call(Operand<T> input) {
+    if (!input.asOutput().dataType().isFloating()) {
+      throw new IllegalArgumentException(
+          "Must be a Floating Point DataType: " + input.asOutput().dataType());
+    }
+    Operand<T> result = tf.nn.elu(input);
+    if (alpha == 1.0) {
+      return result;
+    } else {
```
karllessard, line 96: Nit: no need of a `else` block if you return unconditionally in the `if` one

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Calculates the Exponential activation.
+   *
+   * @param input the input tensor
+   * @return an Operand for the exponential activation: <code>exp(x)</code>.
+   * @throws IllegalArgumentException if the input is not a floating type
+   */
+  @Override
+  public Operand<T> call(Operand<T> input) {
+    if (!input.asOutput().dataType().isFloating()) {
```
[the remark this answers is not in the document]
karllessard: Ultimately, we would have a `TFloating` family that can the be bound to the `T` parameter of that method (this family is present in PR #92 ).

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,144 @@
...
+  @Override
+  public Operand<T> call(Operand<T> input) {
+
+    DataType<T> dataType = input.asOutput().dataType();
+
+    boolean clipMax = !Float.isNaN(maxValue);
+    Operand<T> negativePart = null;
+    if (alpha != 0) {
+      if (Float.isNaN(maxValue) && threshold == 0) {
+        // TODO LeakyRelu is not in tf.nn ????
```
karllessard, line 108: `LeakyRelu` is not add to `tf.nn` because it is set as [not visible](https://github.com/tensorflow/tensorflow/blob/465eab955687d78dc7431fcc52e62d7694d2c6da/tensorflow/core/api_def/base_api/api_def_LeakyRelu.pbtxt#L3) in the main API def proto. To add it to `tf.nn`, you can override its visibility by setting it to `VISIBLE` in [our proto](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_LeakyRelu.pbtxt)

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -0,0 +1,144 @@
...
+                tf.math.add(tf.math.neg(input), tf.dtypes.cast(tf.constant(threshold), dataType)));
+      } else {
+        negativePart = tf.nn.relu(tf.math.neg(input));
+      }
+    }
+
+    if (threshold != 0) {
+      // computes input for input > threshold else 0
+      Greater greater = tf.math.greater(input, tf.dtypes.cast(tf.constant(threshold), dataType));
+      input = tf.math.mul(input, tf.dtypes.cast(greater, dataType));
```
karllessard, line 123: I would try to avoid changing the value of a method parameter, better use a new variable.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
deansher: My leaning would be to do it as a separate complete pass that results in a new invariant in the build, like tool X with configuration Y gives no warnings. Also, that would let us agree on conventions and then bootstrap following them throughout the framework API.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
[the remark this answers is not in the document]
deansher: Good point, so this is fine for now.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Tanh.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Creates a Hyperbolic tangent activation.
+   *
+   * @param tf the TensorFlow Ops
+   */
+  public Tanh(Ops tf) {
+    super(tf);
+  }
+
+  /** {@inheritDoc} */
```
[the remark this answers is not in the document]
karllessard: Similar to my [previous comment](https://github.com/tensorflow/java/pull/123/files#r505123680), when we will bind `T` to `TFloating`, no doc will be required as no exception will be thrown

karllessard (commented):
@JimClarke5 , it looks like this PR can almost be merged, can you please go through all remaining comments and resolve them, either by a fix or by an answer? Thanks

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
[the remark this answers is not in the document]
karllessard: I would even say that if you feel adding `TFloating` right now @JimClarke5 , go ahead, just extend `TNumber` from it and implement it in all floating point tensor types we have (`TFloat32`, `TFloat16`, ...). Then you can remove all these type checks at runtime by binding `T` to `TFloating`, i.e. `Exponential<T extends TFloating> { ... } `

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SwishTest.java
```diff
@@ -0,0 +1,106 @@
...
+              Operand<TInt32> result = instance.call(tf.constant(input));
+            }
+          });
+  }
+
+  /** Test of Swish call method */
+  @Test
+  public void testCallFloat() {
+    float[] input = {1, -2, 3, -4, -5, 6, -7, 8};
+    float[] expected = {
```
KartikChugh, line 65: Where were the expected values computed? I would've thought we'd be using the examples provided in the Python API docs: https://www.tensorflow.org/api_docs/python/tf/keras/activations/swish

KartikChugh (approved):

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
JimClarke5: The present `@Nullable` is in a private Java api and will break after Java 8.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
JimClarke5: OK

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -0,0 +1,103 @@
...
+import org.tensorflow.types.TBool;
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Exponential linear unit.
+ *
+ * <p>The exponential linear unit (ELU) with <code>alpha &gt; 0</code> is:
+ *
+ * <p><code>x</code> if <code>x &gt; 0</code> and <code>alpha * (exp(x) -
+ * 1)</code> if <code>x &gt; 0</code>.
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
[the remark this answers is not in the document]
JimClarke5: @karllessard  Ok, did we want to also add `TInteger` while we're at it?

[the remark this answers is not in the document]
JimClarke5: Added `TFloating extends TNumber` and modified `TBfloat16`, `TFloat16`, `TFloat32`, and `TFloat64` to extend `TFloating`. Activation classes have been modified for `TFloating` where appropriate.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: The numbers less than the `threshold` get converted to `zero` before the `clipMax` is applied.
So, there is no need to move the lower bound on the `clipByValue` operation.

```
if (threshold != 0) {
      // computes input for input > threshold else 0
      Greater greater = tf.math.greater(input, tf.dtypes.cast(tf.constant(threshold), dataType));
      input = tf.math.mul(input, tf.dtypes.cast(greater, dataType));
```

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/SELU.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Gets the calculation operation for the activation.
+   *
+   * @param input the input tensor
+   * @return The operand for the activation
+   * @throws IllegalArgumentException if the data type is not a floating data type.
+   */
+  @Override
+  public Operand<T> call(Operand<T> input) {
```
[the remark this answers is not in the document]
JimClarke5: This is no longer the case after implementing `TFloating`.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: OK

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/SwishTest.java
JimClarke5: For activations, the expected values where determined by running the same inputs against Python TF.

JimClarke5:
I have just pushed the changes requested, the biggest was adding `TFloating` in the core-api.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
```diff
@@ -0,0 +1,92 @@
...
+import org.tensorflow.op.core.ReduceMax;
+import org.tensorflow.op.core.ReduceSum;
+import org.tensorflow.types.family.TNumber;
+
+/**
+ * Softmax converts a real vector to a vector of categorical probabilities.
+ *
+ * <p>The elements of the output vector are in range (0, 1) and sum to 1.
+ *
+ * <p>Each vector is handled independently. The <code>axis</code>argument sets which axis of the
```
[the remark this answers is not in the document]
JimClarke5: Fixed

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: @karllessard should we do this or just leave it as is for internal use? In Python it is available as `tf.nn.leaky_relu`.
If we want to add this, it may be better as a separate PR because of the generated classes being changed.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
saudet: BTW, we can use Lombok for that kind of stuff. It needs a dependency that's used only by the compiler, it's not required at runtime.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
karllessard: Ok I suggest we keep the `@Nullable` introduction for the future, outside this PR

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TFloat64.java
```diff
@@ -29,10 +29,11 @@
 import org.tensorflow.ndarray.NdArray;
 import org.tensorflow.ndarray.StdArrays;
 import org.tensorflow.ndarray.impl.dense.DoubleDenseNdArray;
+import org.tensorflow.types.family.TFloating;
```
karllessard, line 32: Provably that the `TNumber` import is not required anymore?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
karllessard: I suggest we make it visible in this PR, even if the ops will be regenerated, only two files (ˋNnOps` and ˋLeakyRelu`) should end up being different

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TFloat64.java
JimClarke5: @karllessard I don’t fully understand your comment. Did you mean `TNumber` is not required for most Activations or not required at all? `TFloating` works well with those activations that only work with floating point. But there is still use for `TNumber` when any number type is accepted.

karllessard: No, I just meant that in these type classes `TFloat32, TFloat16, etc.`, the `TNumber` seems not to be used anymore since you replaced it by `TFloating` therefore there is no need to import it :)

karllessard: actually my comment should have been put on the line below... sorry, that is kind of things that happens when you review a PR on a iPad

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
karllessard: @Craigacp , do you have any other comment about this?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Softmax.java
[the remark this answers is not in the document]
karllessard: I think it could be reasonable still to have a `Operand.shape()` operation... let's keep that anyway for another PR

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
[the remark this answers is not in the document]
saudet: Hum, from what I understand that's what `Optional` is for and that's in Java SE 8. Wouldn't that do? We can make it such that anything that isn't `Optional` can't be null. Anyway, something for a future PR, yes.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
JimClarke5: Fixed

tensorflow-framework/src/test/java/org/tensorflow/framework/activations/ELUTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @BeforeEach
+  public void setUp() {}
+
+  @AfterEach
+  public void tearDown() {}
+
+  /** Test of ELU call method */
+  @Test
+  public void testCallInt() {
```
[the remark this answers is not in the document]
JimClarke5: This test is obsolete after implementing TFloating as it will no longer compile.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TFloat64.java
JimClarke5: Interesting the `google-java-format` did not remove the `TNumber` import, I had to remove them all by hand.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: I have made the changes to add `LeakyRelu` to NnOps and generate `tf.nn.leakyRelu()`

karllessard (approved):
@Craigacp @deansher , there are a few discussions you've started in this PR that is still open, for me it looks like minor changes but can you can please have a quick check and give us your blessing to merge it in its current state or not?

deansher (approved):
Ship it! This is a huge step forward.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
Craigacp: If TF python is doing it then I guess it's fine.

Craigacp (approved):

karllessard (approved):

karllessard:
Thanks @JimClarke5 !

PR #124, opened 2020-10-02
Add scripts for deployment and releases

karllessard:
This is the new version for https://github.com/tensorflow/java/pull/118

Native artifacts are first build uploaded on OSSRH, then fetch and redeploy by running the `deploy.sh` script. Publishing a snapshot or a release now share the same process, except that for releases we must execute the script locally for signing while it is done by the CI build for the snapshots.

Note that I'm still validating that solution in my personal fork, there might still be small changes coming up but you can already start to take a look at it.

Again, you don't need to go through all changes in the Java files from the first commit, they are just updates to the Javadoc to in order to pass the doclint checks.


RELEASE.md
```diff
@@ -0,0 +1,221 @@
...
+    **unchecked**, otherwise it will conflict with locally signed artifacts.
+-   A GPG signing key, required
+    [to sign the release artifacts](http://central.sonatype.org/pages/apache-maven.html#gpg-signed-components).
+
+### Preparing a release
+
+#### Major or minor release
+
+1.  Get a clean version of the source code by cloning the
+    [TensorFlow Java GitHub repository](https://github.com/tensorflow-java)
```
Craigacp, line 55: Incorrect link.

RELEASE.md
```diff
@@ -0,0 +1,221 @@
...
+    ```
+    git add .
+    git commit -m "Releasing 1.0.0"
+    git push --set-upstream upstream r1.0
+    ```
+
+#### Patch release
+
+1.  Get a clean version of the source code by cloning the
+    [TensorFlow Java GitHub repository](https://github.com/tensorflow-java)
```
Craigacp, line 77: Incorrect link.

RELEASE.md
```diff
@@ -0,0 +1,221 @@
...
+4.  Commit the changes and push the branch to the GitHub repository
+    ```
+    git add .
+    git commit -m "Releasing 1.0.1"
+    git push
+    ```
+
+### Building native artifacts
+
+Once a release branch (i.e. a branch prefixed by `r`) is pushed on the repository, a workflow will
```
Craigacp, line 101: Does any commit to a release branch trigger this?

RELEASE.md
```diff
@@ -0,0 +1,221 @@
...
+There is no user action required for this step other than watching the progress of the GitHub
+Actions workflow and making sure that all steps have been completed successfully.
+
+#### Build native artifacts manually
+
+Some platforms cannot be build successfully on GitHub Actions, due to some limits to their resources
+(e.g. max 6 hours for a job). For this reasons, we need to build manually some of our artifacts on
+private servers. 
+
+To do so, follow the same steps as the [CI build](https://github.com/tensorflow/java/.github/workflows/ci.yml)
```
Craigacp, line 114: This link should be something like this - https://github.com/tensorflow/java/blob/master/.github/workflows/ci.yml

deploy.sh
```diff
@@ -0,0 +1,152 @@
...
+#
+# This script is intended to be run inside a docker container to provide a
+# hermetic process. See release.sh for the expected invocation.
+
+set -e
+
+# By default we deploy to both ossrh and bintray. These two
+# environment variables can be set to skip either repository.
+DEPLOY_OSSRH="${DEPLOY_OSSRH:-true}"
+DEPLOY_BINTRAY="${DEPLOY_BINTRAY:-false}"
```
Craigacp, line 25: The comment says it deploys to both bintray and ossrh, but doesn't this line turn off bintray by default?

release.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 2: Fix the copyright year?

deploy.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 2: Fix the copyright year?

tensorflow-core/pom.xml
```diff
@@ -480,7 +454,7 @@
     </profile>
 
     <!-- Disable platform dependency when building the artifact for the dependency itself -->
-    <profile>
+    <!--profile>
```
Craigacp, line 457: Why did this profile get commented out?

tensorflow-framework/src/main/java/org/tensorflow/framework/data/Dataset.java
```diff
@@ -255,12 +255,12 @@ public DatasetIterator makeOneShotIterator() {
 
   /**
    * Creates an in-memory `Dataset` whose elements are slices of the given tensors. Each element of
-   * this dataset will be a List<Operand<?>>, representing slices (e.g. batches) of the provided
-   * tensors.
+   * this dataset will be a {@code ListOperand<?>>}, representing slices (e.g. batches) of the
```
Craigacp, line 258: This comment lost a `<`.

Craigacp (changes requested):
There are a few incorrect links in the doc, and other minor things, but otherwise it looks good. I've not checked the JavaCPP bits in detail.

karllessard:
I need to change the script so that native artifacts downloaded for OSS are signed as well on a release, `gpg:sign-and-deploy-file` can be used for that. I'll push a fix soon.

karllessard:
@Craigacp : yes looks like I've done a bad cleanup, I've started from [this file](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/java/maven/README.md) in the old repo. I'll go through it once again (and yes, I guess at this point we can copyright the version to 2020...)

RELEASE.md
karllessard: Yes, it is assumed that any commit to a release branch is for releasing a version. On a major/minor release, a new branch is created. On a patch release, we push changes on the existing branch for the release we are patching.

RELEASE.md
karllessard: Wow, I wrote that thing to fast, all broken links are new ones :-/

deploy.sh
karllessard: I'm thinking now about removing completely support for Bintray. The deployment process I'm putting in place uses exclusively OSSRH so keeping this support for Bintray just bring noise in the scripts.

tensorflow-core/pom.xml
karllessard: Despite the comment, I'm not sure to understand what this profile is used for but I know that the deployment script fails when it is present but everything works fine when it is not.

Craigacp: Maybe we should delete it then. We can always pull it out of the history.

deploy.sh
Craigacp: Maybe commit it with bintray then we can remove it later. That way it's in the repo history and we can get it if necessary.

RELEASE.md
Craigacp: Ok. Might be worth explicitly noting that then.

karllessard:
Just pushed new changes, where I totally remove support for Bintray (we'll restore it from history if we really need it one day).

karllessard:
image

saudet:
You've been able to get staging without closing working without nexus-staging-maven-plugin? That's kind of strange. Maybe Sonatype changed the settings of OSSRH so we don't need to use it anymore.

karllessard:
@saudet : yes, no matter if the repository is created by the maven-deploy-plugin or manually in the CI build, it remains open until I close it from the console. I can't tell if this is a change of behavior since it is the first time I doing something like this.

deploy.sh
karllessard: It is already there, I'll take care of not squashing the commits on merge and it will remain in history.

tensorflow-core/pom.xml
karllessard: Yes, that is what I ended up to do.

PR #133, opened 2020-10-14
Fix validateTags() bug

frankfliu:
empty string array is a valid tag. TF 1.x model does not have any tags


frankfliu:
@googlebot I signed it!

karllessard:
Thanks @frankfliu

karllessard (approved):

PR #138, opened 2020-10-30
Fix decoding of string length

brychcy:
When turning tensors with length > 127 into java Strings, they were truncated.

Patch contains an added test case with an example.


karllessard (approved):
Thanks @brychcy !

brychcy:
Can you pleeeease make a release with this soonish? This is the only issue that stops us from switching to these awesome new bindings.
 (I tried using the snapshots, but that doesn't work with our corporate setup.)

karllessard:
Hi @brychcy, I would honestly be pleased to unblock you that quickly but our release process is not currently up-to-speed and I personally don't have the bandwidth for doing another one right now (releasing projects with native artifacts like TF is not as straightforward as any standard Java/Maven projects)

I'm following up with Google right now to get more resources available so we can do quick releases like this. Hopefully, these will be unblocked soon, I'll let you know.

PR #150, opened 2020-11-16
Make Ops classes accessible from sub-Ops classes

### LINKED ISSUE #147: Make Ops class assessible from sub-ops classes (i.e. MathOps)
I'm writing some Kotlin extension functions, which if you aren't familiar with them, take `this` as an argument like python's `self` but hide it.  Essentially adding a function to `MathOps` or one of the other ops classes, like this:
```kotlin
fun <T: TType> MathOps.mean(x: Operand<T>, vararg axis: Int, keepDims: Boolean? = null) = mean(x, constant(axis), Mean.keepDims(keepDims))
```

However, if you look at the above snippet, there's a problem: there is no `constant` in scope, it's in `Ops` and not accessible from `MathOps`.  This happens whenever you want to use ops from other holder classes.  I would like to make Ops accessible from the sub-ops classes, probably just by passing it as a parameter to the constructor (which is package private and only called from Ops).  You could pass it instead of the Scope and get the scope from it if you make Ops' `scope` package private.  I most likely won't be the only person creating extension functions that want to use other ops.

I can make a PR if the idea is approved.

rnett:
I adjust codegen to make ops group classes take Ops instead of Scope, get their Scope from it, save it in a field (`ops`), and provide an accessor for that field (`ops()`).

Fixes #147.


rnett:
@googlebot I signed it!

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/AudioOps.java
```diff
@@ -144,4 +147,11 @@ public Mfcc mfcc(Operand<TFloat32> spectrogram, Operand<TInt32> sampleRate,
       Mfcc.Options... options) {
     return Mfcc.create(scope, spectrogram, sampleRate, options);
   }
+
+  /**
+   * Get the parent {@link Ops} object.
+   */
+  public final Ops ops() {
+    return ops;
+  }
```
karllessard, line 156: Do we need this to be public?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/LinalgOps.java
```diff
@@ -1606,4 +1609,11 @@ public LoadAndRemapMatrix loadAndRemapMatrix(Operand<TString> ckptPath,
       TriangularSolve.Options... options) {
     return TriangularSolve.create(scope, matrix, rhs, options);
   }
+
+  /**
+   * Get the parent {@link Ops} object.
+   */
+  public final Ops ops() {
+    return ops;
+  }
```
karllessard, line 1618: I was expecting to see [`LinalgSparseOp`](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/LinalgSparseOps.java) in your PR, which should be a child of `LinalgOps`, but now looking at the master code base, it looks that this child ops is not linked anymore to `LinalgOps`. Probably a mistake that has been introduced in the past by upgrading the TF core, unless all these `sparseMatrix*` endpoints have been removed from the native library as well... we can investigate that later but maybe right now we should simply delete it from our sources, as it brings confusion.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/AudioOps.java
rnett: Unless you'd rather expose the field.  The point, as I understood it, was to allow a user to get to the Ops instance from a subgroup like MathOps, which requires it being public.  I could just make the final field public but the codebase seems to prefer accessor functions as is typical for Java.

karllessard: Ah I forgot that when adding an extension method to `MathOps`, you are not in the scope of its package but in the public scope. So yes, this needs to be public then, all good, thanks

karllessard (approved):

karllessard:
Thanks for the contribution @rnett !

PR #151, opened 2020-11-16
Add default methods to Operand to get the shape and datatype from asOutput().

### LINKED ISSUE #146: Only having shape and dataType on Output is unintuitive and hard to find
I just spent an hour or so trying to figure out how to get the build-time shape of a graph operation.  Had an issue ready to go and everything.  It wasn't until I went poking around in the debugger that I realized it was on Output.  I realize that there are design reasons to separate the output from the op, but `asOutput()` already exists, so I think having `shape()` and `dataType()` default methods on Operation that use `asOutput` is worth it from a usage and discovery standpoint.

rnett:
Add default methods to Operand to get the shape and datatype from `asOutput()`.

Fixes #146.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Operand.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Operand.java
@@ -15,6 +15,7 @@ limitations under the License.
 
 package org.tensorflow;
 
+import org.tensorflow.ndarray.Shape;
 import org.tensorflow.op.Op;
 import org.tensorflow.types.family.TType;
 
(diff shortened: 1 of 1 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Operand.java
```diff
@@ -76,4 +77,18 @@
   default T data() {
     return asOutput().tensor().data();
   }
+
+  /**
+   * Returns the build-time shape of this operand.  May be only partially known.
+   */
+  default Shape shape(){
```
karllessard, line 84: I guess we want to mark the same methods in [`Output`](https://github.com/tensorflow/java/blob/91400f4e0f26f4324a8843599810baf346583395/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java#L40) as `@Override` now

karllessard (changes requested):
Thanks for the PR @rnett, just a small change and then we are good

karllessard:
@rnett , I've started to create PRs for the type-system refactoring (see #139) and the first one (#153) partially covers the changes in your PR, but wrt the new design. Plus, the addition of `dataType()` in `Operand` will come soon after, again with some alterations.

Therefore, I would like to see if you would agree to close your PR to avoid any conflict.

rnett:
I saw that, it's fine by me.

PR #153, opened 2020-11-19
[TypeRefact] Added `Shaped` interface shared across different containers

karllessard:
PR related to #139.

It extracts the shape-related operators of the `NdArray` interface into a separate interface (`Shaped`) that is shared across the client for each container with a given shape. 

I had to rename the output of a few operators as it was conflicting with the `size()` method of `Shaped`, which is now inherited by `Operand`.

CC\ @deansher

### DIFF
--- a/ndarray/src/main/java/org/tensorflow/ndarray/NdArray.java
+++ b/ndarray/src/main/java/org/tensorflow/ndarray/NdArray.java
@@ -55,29 +55,7 @@ import org.tensorflow.ndarray.index.Index;
  *
  * @param <T> the type of values to be mapped
  */
-public interface NdArray<T> {
-
-  /**
-   * @return the shape of this N-dimensional array
-   */
-  Shape shape();
-
-  /**
-   * @return the rank of this N-dimensional array
-   */
-  default int rank() {
-    return shape().numDimensions();
-  }
-
-  /**
-   * Computes and returns the total size of this N-dimensional array, in number of values.
-   *
-   * <p>For example, given a 3x3x2 matrix, the return value will be 18.
-   * @return total size of this nd array
-   */
-  default long size() {
-    return shape().size();
-  }
+public interface NdArray<T> extends Shaped {
 
   /**
    * Returns a sequence of all elements at a given dimension.
--- a/ndarray/src/main/java/org/tensorflow/ndarray/NdArrays.java
+++ b/ndarray/src/main/java/org/tensorflow/ndarray/NdArrays.java
@@ -65,7 +65,7 @@ public final class NdArrays {
     if (values == null) {
       throw new IllegalArgumentException("Values cannot be null");
     }
-    return wrap(DataBuffers.of(values, false, false), Shape.of(values.length));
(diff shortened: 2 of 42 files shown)


deansher (approved):
To me, this is a clear improvement. It does add a slight hand-wave to code like the following, but I lean toward believing that the previous, more explicit, version required just as much subtle knowledge as the new, more hand-wavy version.

Example of the old:
``` Java
  public Operand<T> call(Operand<T> input) {
    Shape shape = input.asOutput().shape();
```

This now becomes:
``` Java
  public Operand<T> call(Operand<T> input) {
    Shape shape = input.shape();
```

The old version explicitly showed it was talking about `input`'s shape as an `Output` rather than as an eagerly evaluated `Tensor`. Arguably, that was good. But as someone who only recently learned this distinction (and who still feels slightly insecure on this point), I can say that `.asOutput()` didn't clarify this for me before I understood, and that it seems unnecessary now that I do.

I did suggest retaining overriding-class-specific javadoc in a couple of places that at least hints at this distinction.

karllessard:
Thanks @deansher , I've updated the sources with your suggestions

Craigacp (approved):
LGTM.

However I did go and look at `org.tensorflow.ndarray.Shape` again. The computation of the size seems racy, and it stores it as a boxed `Long` rather than a primitive `long`. At the very least it should store it as a `long`. It might even be better to compute the shape size eagerly in the constructor, then the whole object becomes immutable.

karllessard:
>  It might even be better to compute the shape size eagerly in the constructor, then the whole object becomes immutable.

That was my initial implementation, as it's the most intuitive one as well, but I've rolled back to that lazy-initialization of the size because surprisingly its computation had a bigger impact on my [benchmark](https://github.com/tensorflow/java/blob/master/ndarray/src/test/java/org/tensorflow/ndarray/benchmark/NdArrayBenchmark.java) results that I was expecting. 

We can revisit this later if you want but just to let you know that it is not a design mistake but a deliberated choice.

Craigacp:
> > It might even be better to compute the shape size eagerly in the constructor, then the whole object becomes immutable.
> 
> That was my initial implementation, as it's the most intuitive one as well, but I've rolled back to that lazy-initialization of the size because surprisingly its computation had a bigger impact on my [benchmark](https://github.com/tensorflow/java/blob/master/ndarray/src/test/java/org/tensorflow/ndarray/benchmark/NdArrayBenchmark.java) results that I was expecting.
> 
> We can revisit this later if you want but just to let you know that it is not a design mistake but a deliberated choice.

Ok, do you have the output of that benchmark before and after somewhere or should I run it myself?

karllessard:
>  Ok, do you have the output of that benchmark before and after somewhere or should I run it myself?

Yes give it a try and let see what you observe on your side. If I remember correctly, pay more attention on [this test](https://github.com/tensorflow/java/blob/fc3d960d9e7837464ff60df19cfbd3cee794ccfd/ndarray/src/test/java/org/tensorflow/ndarray/benchmark/NdArrayBenchmark.java#L127)

PR #158, opened 2020-11-27
Change eager mode control inputs from exception to no-op, reflecting eager semantics.

### LINKED ISSUE #157: Gracefully handle control dependencies in Eager enviroments
Currently, calling `withControlDependencies` in Eager mode will cause an exception when the op is build.  Ideally, this wouldn't be called at all, as it does nothing: if you have a Operand to pass into it, it has already been calculated.  However, when writing code that should work in eager or graph mode (like `tensorflow-framework`) having to check for this every time you want to add a control dependency is annoying and currently doesn't seem to be done.  Semantically, using a control dependency in eager mode is a no-op.

I'd propose making `EagerOperationBuilder.addControlInput` a no-op instead of throwing an exception for the reasons above.  This could also be handled in the `Scope` methods if you don't want to change `EagerOperationBuilder` for some reason.  I can handle the PR if this is approved.

rnett:
Simple PR to fix #157.  Makes adding control inputs in eager mode a no-op instead of an exception, for the reasons described in the issue (since it's eager mode, any control deps being added will already have been computed).

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
@@ -90,8 +90,8 @@ final class EagerOperationBuilder implements OperationBuilder {
 
   @Override
   public OperationBuilder addControlInput(Operation control) {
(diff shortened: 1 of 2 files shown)


karllessard (approved):
Thanks @rnett !

PR #162, opened 2020-12-05
Add ones op

rnett:
Simple PR to add a ones op to match Python TF and numpy APIs.  Copies from Zeros and operates in the same way (`fill` + `constant`).

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
+++ b/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
@@ -128,6 +128,7 @@ import org.tensorflow.op.core.MutexLock;
 import org.tensorflow.op.core.NextIteration;
 import org.tensorflow.op.core.NoOp;
 import org.tensorflow.op.core.OneHot;
+import org.tensorflow.op.core.Ones;
 import org.tensorflow.op.core.OnesLike;
 import org.tensorflow.op.core.OrderedMapClear;
 import org.tensorflow.op.core.OrderedMapIncompleteSize;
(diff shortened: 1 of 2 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Ones.java
```diff
@@ -0,0 +1,77 @@
+/* Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
karllessard, line 1: Please adjust the year to 2020

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Ones.java
```diff
@@ -0,0 +1,77 @@
...
+import org.tensorflow.op.dtypes.Cast;
+import org.tensorflow.types.TString;
+import org.tensorflow.types.family.TNumber;
+import org.tensorflow.types.family.TType;
+
+/**
+ * An operator creating a constant initialized with ones of the shape given by `dims`.
+ * 
+ * <p>For example, the following expression
+ * <pre>{@code tf.ones(tf.constant(shape), TFloat32.DTYPE)</pre>
```
karllessard, line 34: The code block is not closed (same thing below)

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -7726,7 +7740,7 @@ public Ops withName(String opName) {
   }
 
   /**
-   * Returns an API that uses the provided DeviceSpec for an op.
+   * Returns an API that places the created operations on the device(s) matching the provided spec.
```
karllessard, line 7743: Ah, looks like something was out of sync with the last PR that was merged, good thing that we'll fix it here.

karllessard (changes requested):

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Ones.java
rnett: I think this is due to a cc generation issue, as this is copied from Zeroes and the same issues are present there.

karllessard (approved):

PR #166, opened 2020-12-07
Indexing API

### LINKED ISSUE #164: Indexing/slicing API
If you want to index a tensor currently your only choice is `strided_slice` or `slice`, neither of which have good APIs.

I'd like to create an indexing API that supports point indices, slices, strided slices, ellipses, all, and newaxis.  This isn't particularly hard to do by using `strided_slice`, I already have a similar Kotlin API I can port.  This will also port well to the Kotlin API, especially with the `rangeTo` operator.

I'd like the usage to look something like:
```java
tf.slice(x, Index.point(0), Index.all, Index.slice(0, 5, 2) // [0, :, 0:5::2]

tf.slice(x, Index.newaxis, Index.point(10, true), Index.elipses) // [tf.newaxis, 10:11, ...]
```

This would cover the assign variants, too.

rnett:
Fixes #164.

One question: Should I also make endpoints named `get` and `set`?  It's more discoverable to people coming from Python (who usually use `[]` rather than the stridedSlice op).

There's examples in the `Index` javadoc.


tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -5899,6 +5900,61 @@ public StageSize stageSize(List<DataType<?>> dtypes, StageSize.Options... option
...
+   *   The goal of this op is to produce a new tensor with a subset of
+   *   the elements from the `n` dimensional `input` tensor. The subset is chosen using
+   *   a sequence of `m` sparse range specifications encoded into the arguments
+   *   of this function. Note, in some cases
+   *   `m` could be equal to `n`, but this need not be the case. Each
+   *   range specification entry can be one of the following:
+   *   <p>
+   *   - An ellipsis (...) using {@link Index#ellipses()}. Ellipses are used to imply zero or more
+   *     dimensions of full-dimension selection and are produced using
+   *     `ellipsis_mask`. For example, `foo[...]` is the identity slice.
```
Craigacp, line 5915: We don't have this kind of overloading in Java, so does it make sense to describe it here in the Java Ops?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Indexing.java
```diff
@@ -0,0 +1,206 @@
...
+    final int[] begin;
+    final int[] end;
+    final int[] strides;
+    final long beginMask;
+    final long endMask;
+    final long ellipsisMask;
+    final long newAxisMask;
+    final long shrinkAxisMask;
+
+    public StridedSliceArgs(int[] begin, int[] end, int[] strides, long beginMask, long endMask, long ellipsisMask,
```
Craigacp, line 44: Does this need to be public? Can't it be package private?

Craigacp (commented):
More generally these indices don't seem to link up with the indexing operations already available in ndarray, which might prove more confusing as the ndarrays move closer to Tensors given Karl's current refactorings.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Indexing.java
rnett: Yeah, it should be.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
rnett: Agreed.  I copied most of this from the `stridedSlice` docs and tried to keep it as close as possible.  I talk enough about the Python and Java equivalents that I think it's understood to be `stridedSlice(foo, Index.ellipsis())`.  I'm kind of using it as a generic "slice notation" rather than specifically referencing the Python syntax.

rnett:
I looked at that, there's a few concepts that are impossible to map into `stridedSlice`, which as far as I know is the only Op we could use for this.  In particular, multiple points (`seq`) and `hyperslab`.

Craigacp:
I think extending the ndarray API to cover the gaps would be worthwhile, and then we can throw exceptions for things which fail to map from ndarray into TF ops.

rnett:
There's some odd things in the ndarray API like `At` not supporting negative indices, `Range` not having a stride or supporting negative indices, and having flip, even, and odd separate from `[::-1]`, `[::2]`, and `[1::2]`.  Do you think I should bring the ndarray indexing a bit closer to Numpy?  It would make the differences between the ndarray indexing and the tensor indexing less confusing.

karllessard:
> Do you think I should bring the ndarray indexing a bit closer to Numpy? It would make the differences between the ndarray indexing and the tensor indexing less confusing.

Yes, if something is lacking in the `NdArray` API, it should be added instead of trying to fill the gaps from TensorFlow Java core instead.

> If we do want to unify APIs, I think it would be easier to map this API onto the current ndarray one the visa versa. To map the ndarray indices onto Operands

That was the original intent, we are just not there yet. Again (sorry for that), the tensor type refactoring currently in progress might be helpful to achieve this. I also thing that `TType` should extend from `Operand` at some point, so we can pass a tensor directly to an op without converting it to a constant explicitly. I did something like that in a previous draft for the same refactoring [here](https://github.com/karllessard/tensorflow-java/blob/tensor-as-ndarrays/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TType.java).

Should we also put this PR on hold for now and continue that discussion in #164 ?

rnett:
> I also think that TType should extend from Operand at some point

That would be very nice, especially for eager mode.

I pushed what I have as far as adding to the ndarray API, it doesn't seem that bad, just newaxis, ellipsis, and strided slice.  My impression from the refactoring is that Operand indexing will still have to use StridedSlice?  If so most of this PR is still fine, I just need to move the builders to the NDArray builders and unify the APIs (which is mostly done as of the latest commit, just needs cleanup).  I'll add a comment to #164 about the API unification changes I would like to make.

karllessard:
Idem than https://github.com/tensorflow/java/pull/167#issuecomment-751778041

rnett:
@karllessard can you re-open this PR too?  I don't have the permissions to.

karllessard:
I'm really sorry, I don't know why but it does not allow me to do it neither, though I was able to do it with #167 ... 

image

karllessard:
Can you try this? https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c

rnett:
@karllessard I still have no re-open option, but I did the steps there.  Can you do it now?

rnett:
Ok, this is ready for review.  One last thing: I'd like to move the builders from `Indices` to `Index`, it reads better to me (kind of like if they were nested classes) and is shorter.   Thoughts?

ndarray/src/main/java/org/tensorflow/ndarray/index/Ellipsis.java
```diff
@@ -12,26 +12,43 @@
...
   @Override
   public long numElements(Dimension dim) {
-    return (dim.numElements() >> 1) + (dim.numElements() % 2);
+    throw new IllegalStateException();
   }
 
   @Override
   public long mapCoordinate(long coordinate, Dimension dim) {
-    return coordinate << 1;
+    throw new IllegalStateException();
```
karllessard, line 37: I don't think `IllegalStateException` is the right exception to be thrown here, since there is no state involve. What about `UnsupportedOperationException`?

ndarray/src/main/java/org/tensorflow/ndarray/index/Ellipsis.java
```diff
@@ -12,26 +12,43 @@
...
 
-  private Even() {
+  @Override
+  public boolean ellipsisMask() {
+    return true;
+  }
+
+  @Override
+  public String toString() {
+    return "Ellipsis()";
```
karllessard, line 52: in `Hyperslab`, you use the `getClass().getSimpleName()` to return that string, should we do it everywhere?

ndarray/src/main/java/org/tensorflow/ndarray/index/Indices.java
```diff
@@ -216,4 +255,129 @@ public static Index flip() {
...
+    return Ellipsis.INSTANCE;
+  }
+
+  /**
+   * An index that expands to fill all available source dimensions.
+   * Works the same as Python's {@code ...}.
+   *
+   * @return index
+   */
+  public static TensorIndex expand(){
```
karllessard, line 284: why having both `expand()` and `ellipsis()` if they are the same?

ndarray/src/main/java/org/tensorflow/ndarray/index/Indices.java
```diff
@@ -216,4 +255,129 @@ public static Index flip() {
...
+   * An index that returns elements between {@code start} and {@code end}.
+   * If {@code start} or {@code end} is {@code null}, starts or ends at the beginning or the end, respectively.
+   * <p>
+   * Analogous to Python's {@code :} slice syntax.
+   *
+   * @return index
+   */
+  public static TensorIndex slice(long start, long end){
+    return slice(start, end, 1);
+  }
```
karllessard, line 334: instead of having all those variants with nullable boxed type and primitive long, could we just rename the methods differently so they accept only parameters that matters?

e.g. 
`slice(Long, long)` -> `sliceTo(long)`
`slice(long, Long)` -> `sliceFrom(long)`
`slice(Long, Long)` -> do we really need this one, as it just returns `all()` ultimately?

ndarray/src/main/java/org/tensorflow/ndarray/index/Slice.java
```diff
@@ -0,0 +1,111 @@
...
+      if(stride > 0){
+        return 0;
+      } else {
+        return dim.numElements() - 1; // it's inclusive
+      }
+    } else if(start < 0){
+      return dim.numElements() + start;
+    } else {
+      return start;
+    }
```
karllessard, line 39: Nit: no need of `else` clauses when you return unconditionally in your previous `if` blocks, I find it improves readability

Also not required, most of the code in TF Java follow a reversed order of the elements of a class, where elements with a higher visibility (public) are at the top of the class while those at the bottom has the lower one.

ndarray/src/main/java/org/tensorflow/ndarray/index/Slice.java
```diff
@@ -0,0 +1,111 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================
+ */
+package org.tensorflow.ndarray.index;
+
+import java.util.StringJoiner;
+import org.tensorflow.ndarray.impl.dimension.Dimension;
+
+final class Slice implements TensorIndex {
```
karllessard, line 22: If we take the `sliceTo`/`sliceFrom` route, having distinct implementation of `Index` (i.e. `SliceTo` and `SliceFrom`) could reduce the complexity/cost of each case and avoid the boxed type conversion. There is not much gain having them all implemented by a single class.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/StridedSliceHelper.java
```diff
@@ -0,0 +1,214 @@
...
+        newAxisMask |= 1L << i;
+      }
+
+      if (idx.shrinkAxisMask()) {
+        shrinkAxisMask |= 1L << i;
+      }
+    }
+
+    return new StridedSliceArgs(begin, end, strides, beginMask, endMask, ellipsisMask, newAxisMask, shrinkAxisMask);
+  }
```
karllessard, line 109: Nit: add a space here

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/StridedSliceHelper.java
```diff
@@ -0,0 +1,214 @@
...
+   *    `begin` and `end` can be negative where `-1` is the last element, `-2` is
+   *    the second to last. `begin_mask` controls whether to replace the explicitly
+   *    given `begin` with an implicit effective value of `0` if `stride > 0` and
+   *    `-1` if `stride < 0`. `end_mask` is analogous but produces the number
+   *    required to create the largest open interval. For example, given a shape
+   *    `(3,)` tensor `foo[:]`, the effective `begin` and `end` are `0` and `3`. Do
+   *    not assume this is equivalent to `foo[0:-1]` which has an effective `begin`
+   *    and `end` of `0` and `2`. Another example is `foo[-2::-1]` which reverses the
+   *    first dimension of a tensor while dropping the last two (in the original
+   *    order elements). For example `foo = [1,2,3,4]; foo[-2::-1]` is `[4,3]`.
```
karllessard, line 142: Documentation in non-generated classes should be converted from Python to Java, meaning that `foo[-2::-1]` and other examples like that must be converted to their Java equivalent.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/StridedSliceHelper.java
```diff
@@ -0,0 +1,214 @@
...
+
+    for (int i = 0; i < indices.length; i++) {
+      TensorIndex idx = indices[i];
+      if (idx == null) {
+        idx = Indices.all();
+      }
+
+      begin[i] = (int) idx.begin();
+      if(begin[i] != idx.begin())
+        throw new IllegalArgumentException("Can't convert long begin value to int for index " + idx + ": Out of bounds");
```
karllessard, line 75: Please add braces around [single-line statements](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used)

karllessard (changes requested):

karllessard:
> Ok, this is ready for review. One last thing: I'd like to move the builders from `Indices` to `Index`, it reads better to me (kind of like if they were nested classes) and is shorter. Thoughts?

I never liked `Indices` neither but I did it to be consistent with other builders distributed in this library, i.e. `NdArrays/NdArray`, `DataBuffers/DataBuffer`, `DataLayouts/DataLayout`, etc. So while `Index` sounds better, it would break to consistency. Though, I'm not totally opposed to it, maybe somebody else can chime in?

ndarray/src/main/java/org/tensorflow/ndarray/index/Indices.java
rnett: Because it really should be called "expand" or similar, but anyone coming from Python will be looking for ellipsis.  I could go with just one.

ndarray/src/main/java/org/tensorflow/ndarray/index/TensorIndex.java
```diff
@@ -12,29 +12,39 @@
  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  See the License for the specific language governing permissions and
  limitations under the License.
- =======================================================================
+ ==============================================================================
  */
 package org.tensorflow.ndarray.index;
 
-import org.tensorflow.ndarray.impl.dimension.Dimension;
+public interface TensorIndex extends Index{
```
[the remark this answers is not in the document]
rnett: I like that better.  For now, I've implemented it as a `tensorSupport` method, since the other methods have useful defaults and I don't like "override a method if it doesn't work", it's too easy to miss when adding indices.  The javadocs of those methods mention tensors too.  Do you think it would be better to call it non-DimensionalSpace or non-NDArray usage, instead of Tensor usage?

ndarray/src/main/java/org/tensorflow/ndarray/index/Indices.java
rnett: We already have `to` and `from`, I'll change them to `sliceTo` and `sliceFrom` and add overloads w/ stride.

The reason I included the nullable versions there is for calling with maybe-null data.  Otherwise the caller would have to decide which one to use based on there parameters, so I think it's worth having a method that does that disambiguation.

rnett: Either way I can reduce the overloads though, they were for literal conversions which won't be an issue any more.

ndarray/src/main/java/org/tensorflow/ndarray/impl/dimension/DimensionalSpace.java
```diff
@@ -35,24 +35,42 @@ public static DimensionalSpace create(Shape shape) {
   }
 
   public RelativeDimensionalSpace mapTo(Index[] indices) {
```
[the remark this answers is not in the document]
rnett: I ran with `@Fork(value = 1)` since 0 gave warnings, should I push that?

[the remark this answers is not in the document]
rnett: There's a tiny bit of slowdown, but it's marginal:
Original:
```java
Benchmark                                                  Mode  Cnt   Score   Error  Units
NdArrayBenchmark.getElementAtIndex                         avgt    5   0.097 ± 0.017   s/op
NdArrayBenchmark.readingAllPixelsChannelsByIndex           avgt    5  10.035 ± 0.023   s/op
NdArrayBenchmark.readingAllPixelsChannelsBySequence        avgt    5   0.033 ± 0.002   s/op
NdArrayBenchmark.readingAllPixelsChannelsBySequenceSlices  avgt    5   0.166 ± 0.011   s/op
NdArrayBenchmark.slicing                                   avgt    5   0.222 ± 0.049   s/op
NdArrayBenchmark.writeAllBatchChannels                     avgt    5   0.335 ± 0.048   s/op
NdArrayBenchmark.writeAllPixelsByIndex                     avgt    5   5.870 ± 0.817   s/op
NdArrayBenchmark.writeAllPixelsBySlicing                   avgt    5  29.603 ± 1.304   s/op
NdArrayBenchmark.writeFirstBatchChannels                   avgt    5   0.349 ± 0.102   s/op
NdArrayBenchmark.writeOnePixelBySlicing                    avgt    5   0.705 ± 0.082   s/op
NdArrayBenchmark.writeOnePixelsByIndex                     avgt    5   0.131 ± 0.036   s/op
```
Mine:
```java
Benchmark                                                  Mode  Cnt   Score    Error  Units
NdArrayBenchmark.getElementAtIndex                         avgt    5   0.104 ±  0.020   s/op
NdArrayBenchmark.readingAllPixelsChannelsByIndex           avgt    5  10.037 ±  0.063   s/op
NdArrayBenchmark.readingAllPixelsChannelsBySequence        avgt    5   0.033 ±  0.001   s/op
NdArrayBenchmark.readingAllPixelsChannelsBySequenceSlices  avgt    5   0.174 ±  0.009   s/op
NdArrayBenchmark.slicing                                   avgt    5   0.208 ±  0.001   s/op
NdArrayBenchmark.writeAllBatchChannels                     avgt    5   0.316 ±  0.061   s/op
NdArrayBenchmark.writeAllPixelsByIndex                     avgt    5   5.555 ±  0.137   s/op
NdArrayBenchmark.writeAllPixelsBySlicing                   avgt    5  30.081 ±  3.360   s/op
NdArrayBenchmark.writeFirstBatchChannels                   avgt    5   0.351 ±  0.099   s/op
NdArrayBenchmark.writeOnePixelBySlicing                    avgt    5   0.777 ±  0.347   s/op
NdArrayBenchmark.writeOnePixelsByIndex                     avgt    5   0.122 ±  0.004   s/op
```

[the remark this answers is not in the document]
karllessard: So as we've discussed, the difference between both implementation is negligible.

ndarray/src/main/java/org/tensorflow/ndarray/index/Index.java
```diff
@@ -74,4 +70,61 @@ default Dimension apply(Dimension dim) {
...
+  /**
+   * Returns true if this index is an ellipsis, expanding to take as many dimensions as possible (and applying all() to
+   * them)
+   */
+  default boolean isEllipsis() {
+    return false;
+  }
+
+  /**
+   * Get whether the Index supports Tensor conversion.
```
karllessard, line 90: I don't think we need to mention `for Tensor conversion` in the methods below, as long as the documentation says what the method returns, it can be of any use after that. 

Now for `tensorSupport()`, what about something more generic like `hasBoundaries()`?

ndarray/src/main/java/org/tensorflow/ndarray/index/Indices.java
```diff
@@ -199,21 +265,109 @@ public static Index range(long start, long end) {
...
+    return Ellipsis.INSTANCE;
+  }
+
+  /**
+   * An index that expands to fill all available source dimensions. Works the same as Python's {@code ...}.
+   *
+   * @return index
+   */
+  public static Index expand() {
+    return ellipsis();
```
karllessard, line 310: Like we've [discussed](https://github.com/tensorflow/java/pull/166#discussion_r553098269), we probably just want to go with one name, I'm ok with any (we don't need to be too strict to follow what the names in Python so `expand` is fine if that's what you prefer).

ndarray/src/main/java/org/tensorflow/ndarray/index/Index.java
rnett: Agreed in theory, but I don't like `hasBoundaries`, because to me that sounds like it would exclude `all` and `sliceTo`, etc.  The only things that are excluded are `Hyperslab` and `Sequence`.  I went with `tensorSupport` because I couldn't think of anything better.

Maybe use "strided slice conversion" rather than "tensor conversion"?  In theory, it would apply to any library that uses strided slicing.  Or `specifiesBoundaries`.

karllessard: ok, so let's call it "strided slice conversion" and name this method `isStridedSlicingCompliant()`?

karllessard (approved):

rnett:
Before you merge, any more thoughts on `Indices` vs `Index`?

karllessard:
Yeah I was thinking about that... while I do prefer to move static creators to `Index`, it is less coherent with the rest of the API like I've explained earlier.

I always had in mind that `Indices.*` are good candidates to be imported statically. I know that most IDE might complain unless configuring it otherwise, but in Kotlin static imports like these are more frequent. Meaning that whether they are under `Index` or `Indices` is less important.

Why not doing it in a seperate PR? That one is fully backward compatible, which I like.

rnett:
Another PR is fine, although I don't know that there will be enough impetus to make one, it's a pretty tiny change.  I'd like to have infix or `..` style creators for most these in Kotlin, anyways.

PR #169, opened 2020-12-09
Nicer error messages for mode-forbidden ops

### LINKED ISSUE #154: Better error messages for ops forbidden in Eager mode
Currently, if you try to use `placeholder` in Eager mode you get a cryptic error message that says nothing about it being forbidden in eager in mode (I had to look it up on StackOverflow).  As far as I'm aware there's no ops that are eager only, but I think throwing a custom exception for mode-dependent ops is reasonable and helpful.

I'd probably do this in `EagerOperationBuilder.build` rather than in the ops classes to catch custom created ops.  If that's fine I can make a PR if a list of forbidden ops is provided (or just with `placeholder`).

### LINKED ISSUE #155: tf.variable with initial value does not work in eager mode
As the title says, `tf.variable` with an initial value does not work in eager mode.  I'm not sure if it is supposed to, as `tf.Variable` does in the python api but apparently didn't at first.  Interestingly, it's not the variable op that fails, it's the assign:
```java
cannot compute Assign as input #0(zero-based) was expected to be a int32_ref tensor but is a int32 tensor
org.tensorflow.exceptions.TFInvalidArgumentException: cannot compute Assign as input #0(zero-based) was expected to be a int32_ref tensor but is a int32 tensor
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:87)
	at org.tensorflow.EagerOperationBuilder.execute(EagerOperationBuilder.java:273)
	at org.tensorflow.EagerOperationBuilder.build(EagerOperationBuilder.java:66)
	at org.tensorflow.EagerOperationBuilder.build(EagerOperationBuilder.java:55)
	at org.tensorflow.op.core.Assign.create(Assign.java:97)
	at org.tensorflow.op.core.Helpers.createVariableWithInit(Helpers.java:51)
	at org.tensorflow.op.Ops.variable(Ops.java:7520)
```

rnett:
Adds a way to forbid ops in `ExecutionEnviroment`s, uses it to forbid Placeholder, Variable, and Assign in eager mode.  As far as I know there's no way to detect ahead of time whether an op is forbidden, so this list will have to be added to as people run into mode-specific ops.  These three are likely to be the most common though.

Fixes #155 and fixes #154.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
@@ -56,10 +56,6 @@ import org.tensorflow.proto.framework.DataType;
 final class EagerOperationBuilder implements OperationBuilder {
 
   EagerOperationBuilder(EagerSession session, String type, String name) {
-    if (!session.isOpEnabled(type)) {
-      throw new IllegalArgumentException("Op " + type + " is not valid in eager mode.");
-    }
-
     this.session = session;
     this.type = type;
     this.name = name;
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
@@ -56,8 +56,9 @@ import org.tensorflow.proto.framework.DataType;
 final class EagerOperationBuilder implements OperationBuilder {
 
   EagerOperationBuilder(EagerSession session, String type, String name) {
-    if(!session.isOpEnabled(type))
+    if (!session.isOpEnabled(type)) {
       throw new IllegalArgumentException("Op " + type + " is not valid in eager mode.");
+    }
 
     this.session = session;
     this.type = type;
@@ -78,7 +79,7 @@ final class EagerOperationBuilder implements OperationBuilder {
 
   @Override
   public EagerOperationBuilder addInput(Output<?> input) {
-    addInput(opHandle, (TFE_TensorHandle)input.getUnsafeNativeHandle());
+    addInput(opHandle, (TFE_TensorHandle) input.getUnsafeNativeHandle());
     return this;
   }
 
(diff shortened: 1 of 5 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperationBuilder.java
```diff
@@ -55,6 +55,9 @@
 final class EagerOperationBuilder implements OperationBuilder {
 
   EagerOperationBuilder(EagerSession session, String type, String name) {
+    if(!session.isOpEnabled(type))
+      throw new IllegalArgumentException("Op " + type + " is not valid in eager mode.");
```
karllessard, line 59: can you please add a space after `if` and also wrap the `throw` statement in a block `{...}`? This is more in line with the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used) that we follow in this project.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperationBuilder.java
```diff
@@ -57,6 +57,9 @@
 public final class GraphOperationBuilder implements OperationBuilder {
 
   GraphOperationBuilder(Graph graph, String type, String name) {
+    if(!graph.isOpEnabled(type))
+      throw new IllegalArgumentException("Op " + type + " is not valid in graph mode.");
+
```
karllessard, line 62: same comment as [here](https://github.com/tensorflow/java/pull/169/files#r550779534)

karllessard (changes requested):

karllessard:
What about moving the `isOpEnabled` check [here](https://github.com/tensorflow/java/blob/f85623ed366d903cfddb75177725dc276f359b15/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java#L149) and [here](https://github.com/tensorflow/java/blob/f85623ed366d903cfddb75177725dc276f359b15/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java#L274) to avoid instantiating an operation builder if the op is not supported?

rnett:
> What about moving the `isOpEnabled` check [here](https://github.com/tensorflow/java/blob/f85623ed366d903cfddb75177725dc276f359b15/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java#L149) and [here](https://github.com/tensorflow/java/blob/f85623ed366d903cfddb75177725dc276f359b15/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java#L274) to avoid instantiating an operation builder if the op is not supported?

There's a few places in tests, but that seems fine since it's package private.

karllessard (approved):

PR #172, opened 2020-12-18
[Type Refactor] Move type implementations and mappers to internal package

karllessard:
In this relatively-simple PR, I move the implementations of the typed tensors (`TType`) and their mapping method to the internal package as separate objects. 

While the reason for doing it might seem unclear at this point, it will make more sense in the next PR where the `*Mapper` class could be referenced from a new `@TensorType` annotation. In addition, segregating the implementation classes from their respective interfaces is more scalable for when we'll start supporting different categories of tensors, like sparse and ragged.

Note that these changes has no impact on the user API and is completely backward compatible.

CC\ @deansher

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
@@ -1,0 +1,35 @@
+package org.tensorflow;
+
+import org.tensorflow.internal.c_api.TF_Tensor;
+import org.tensorflow.types.family.TType;
+
+/**
+ * Maps the native memory of a {@link RawTensor} to a n-dimensional typed data space
+ * accessible from the JVM.
+ *
+ * <p>Usage of this class is reserved for internal purposes only.
+ *
+ * @param <T> tensor type mapped by this object
+ * @see {@link TType}
+ */
+public abstract class TensorMapper<T extends TType> {
+
+  /**
+   * Maps the provided dense raw {@code tensor} as a tensor of type {@code T}.
+   *
+   * @param tensor the dense tensor to map, in its raw nature
+   * @return an instance of {@code T}
+   */
+  protected abstract T mapDense(RawTensor tensor);
+
+  /**
+   * Helper for retrieving the native handle of a raw tensor
+   *
+   * @param tensor a raw tensor
+   * @return the native handle of that tensor
+   * @throws IllegalStateException if the tensor has been released
+   */
+  protected static TF_Tensor nativeHandle(RawTensor tensor) {
+    return tensor.nativeHandle();
+  }
+}
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/buffer/ByteSequencer.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/buffer/ByteSequencer.java
@@ -1,0 +1,53 @@
+package org.tensorflow.internal.buffer;
+
+import java.util.Iterator;
+import java.util.function.Function;
+import org.tensorflow.ndarray.NdArray;
+import org.tensorflow.ndarray.NdArraySequence;
+
+/**
+ * Produces sequence of bytes to be stored in a {@link ByteSequenceTensorBuffer}.
+ *
+ * @param <T> source of bytes (byte arrays or strings)
+ */
+public class ByteSequencer<T> implements Iterable<byte[]> {
+
+  /**
+   * Constructor
+   *
+   * @param source source of data
+   * @param byteExtractor method that converts one value of the source into a sequence of bytes
+   */
+  public ByteSequencer(NdArray<T> source, Function<T, byte[]> byteExtractor) {
+    this.source = source;
+    this.byteExtractor = byteExtractor;
+  }
+
+  @Override
+  public Iterator<byte[]> iterator() {
+    return new Iterator<byte[]>() {
+
+      @Override
+      public boolean hasNext() {
+        return scalarIterator.hasNext();
+      }
+
+      @Override
+      public byte[] next() {
+        return byteExtractor.apply(scalarIterator.next().getObject());
+      }
+
+      private final Iterator<? extends NdArray<T>> scalarIterator = source.scalars().iterator();
+    };
+  }
+
+  /**
+   * @return total number of byte sequences that can be produced by this sequencer
+   */
+  long numSequences() {
+    return source.size();
+  }
+
+  private final NdArray<T> source;
+  private final Function<T, byte[]> byteExtractor;
+}
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TBfloat16Mapper.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TBfloat16Mapper.java
@@ -13,15 +13,15 @@ import org.tensorflow.types.TBfloat16;
  * Maps memory of {@link org.tensorflow.proto.framework.DataType#DT_BFLOAT16} tensors
  * to a n-dimensional data space.
  */
-public class TBfloat16Mapper extends TensorMapper<TBfloat16> {
+public final class TBfloat16Mapper extends TensorMapper<TBfloat16> {
 
   @Override
-  public TBfloat16 mapDense(RawTensor tensor) {
+  protected TBfloat16 mapDense(RawTensor tensor) {
     FloatDataBuffer buffer = DataLayouts.BFLOAT16.applyTo(TensorBuffers.toShorts(nativeHandle(tensor)));
     return new DenseTBfloat16(tensor, buffer);
   }
 
-  private static class DenseTBfloat16 extends FloatDenseNdArray implements TBfloat16 {
+  private static final class DenseTBfloat16 extends FloatDenseNdArray implements TBfloat16 {
 
     @Override
     public DataType<TBfloat16> dataType() {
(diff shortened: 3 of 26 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
```diff
@@ -0,0 +1,35 @@
+package org.tensorflow;
+
+import org.tensorflow.internal.c_api.TF_Tensor;
+import org.tensorflow.types.family.TType;
+
+/**
+ * Maps the native memory of a {@link RawTensor} to a n-dimensional typed data space
+ * accessible from the JVM.
+ *
+ * <p>Usage of this class is reserved for internal purposes only.
```
Craigacp, line 10: We might want to have this in a different package as the module system operates on a package level (so when we go to 11 we could not export that package). Though it depends how the methods are used.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/buffer/ByteSequencer.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.function.Function;
+import org.tensorflow.ndarray.NdArray;
+import org.tensorflow.ndarray.NdArraySequence;
+
+/**
+ * Produces sequence of bytes to be stored in a {@link ByteSequenceTensorBuffer}.
+ *
+ * @param <T> source of bytes (byte arrays or strings)
+ */
+public class ByteSequencer<T> implements Iterable<byte[]> {
```
Craigacp, line 13: Maybe `ByteArrayProducer`? `ByteSequencer` sounds like it can change the sequence but my reading of the iterator is that the iteration order is fixed.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TBfloat16Mapper.java
```diff
@@ -0,0 +1,43 @@
...
+import org.tensorflow.types.TBfloat16;
+
+/**
+ * Maps memory of {@link org.tensorflow.proto.framework.DataType#DT_BFLOAT16} tensors
+ * to a n-dimensional data space.
+ */
+public class TBfloat16Mapper extends TensorMapper<TBfloat16> {
+
+  @Override
+  public TBfloat16 mapDense(RawTensor tensor) {
```
Craigacp, line 19: Did you mean to change the access level of `mapDense` here? It's protected in the super class and the other implementations.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TInt64Mapper.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class TInt64Mapper extends TensorMapper<TInt64> {
+
+  @Override
+  protected TInt64 mapDense(RawTensor tensor) {
+    LongDataBuffer buffer = TensorBuffers.toLongs(nativeHandle(tensor));
+    return new DenseTInt64(tensor, buffer);
+  }
+
+  private static class DenseTInt64 extends LongDenseNdArray implements TInt64 {
```
Craigacp, line 23: Should we make this final?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TInt64Mapper.java
```diff
@@ -0,0 +1,42 @@
...
+import org.tensorflow.internal.buffer.TensorBuffers;
+import org.tensorflow.ndarray.buffer.LongDataBuffer;
+import org.tensorflow.ndarray.impl.dense.LongDenseNdArray;
+import org.tensorflow.types.TInt64;
+
+/**
+ * Maps memory of {@link org.tensorflow.proto.framework.DataType#DT_INT64} tensors
+ * to a n-dimensional data space.
+ */
+public class TInt64Mapper extends TensorMapper<TInt64> {
```
Craigacp, line 15: Should these classes all be final?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TBfloat16.java
```diff
@@ -54,7 +50,7 @@
   static final String NAME = "BFLOAT16";
 
   /** Type metadata */
-  DataType<TBfloat16> DTYPE = DataType.create(NAME, 14, 2, TBfloat16Impl::mapTensor);
+  DataType<TBfloat16> DTYPE = DataType.create(NAME, 14, 2, new TBfloat16Mapper());
```
Craigacp, line 53: Should we explicitly put `static final` on here? The `NAME` field is marked as such, and I spent a while looking at this line before I remembered that fields are always static and final on interfaces.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TBool.java
```diff
@@ -45,7 +46,7 @@
   static final String NAME = "BOOL";
 
   /** Type metadata */
-  DataType<TBool> DTYPE = DataType.create(NAME, 10, 1, TBoolImpl::mapTensor);
+  DataType<TBool> DTYPE = DataType.create(NAME, 10, 1, new TBoolMapper());
```
Craigacp, line 49: `static final`?

Craigacp (changes requested):
Small comments.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
karllessard: One of the goal of this abstract class is to be able to access the native handle of the raw tensor from a `TensorMapper` subclass, this is done via `nativeTensor(RawTensor)` and it needs to reside in the `org.tensorflow` package. 

Unless we want to make `RawTensor.nativeHandle()` public (which I would prefer to avoid) or go back to the previous design where both `RawTensor` and `TF_Tensor` are passed to the mapper but that triggered a series of other concerns as discussed in my [previous PR](https://github.com/tensorflow/java/pull/160/files#r535330352).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/buffer/ByteSequencer.java
karllessard: Yeah, I wasn't sure about this one neither... I was hesitating between this and `ByteSequenceProducer` so sounds like the second choice would be a good trade off? (I prefer `ByteSequence` to `ByteArray` to preserve the symmetry with `ByteSequenceTensorBuffer`, or I rename the latter to `ByteArrayTensorBuffer`?).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TBfloat16Mapper.java
karllessard: ah true thanks, I'll keep it protected in subclasses as well, good catch!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TInt64Mapper.java
karllessard: We should yes, thanks

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TBfloat16.java
karllessard: We could but this `DataType` class is going to disappear in my next (and last?) PR about the tensor type refactoring so let's leave it as is for now.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TBool.java
karllessard: Idem as [here](https://github.com/tensorflow/java/pull/172/files#r545846297)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
Craigacp: Ok. I definitely don't want the native handle method to be public. I feel like there is probably a way around this, but it might be more trouble than it's worth, so leaving it here is fine.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TBfloat16.java
Craigacp: Sure.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/buffer/ByteSequencer.java
Craigacp: I think `ByteSequenceProducer` works.

Craigacp (dismissed):
LGTM

karllessard:
Thanks a lot @Craigacp for this quick review!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorMapper.java
saudet: Split packages are definitely something you want to avoid at all costs. These are going to be OSGi users complaining about this as well.

Craigacp: Well we could reflectively access a private method as we'd be in the same module. But that's asking for trouble and sets a bad precedent. There's probably something we could do that's a bit tricksy with inner classes, but I can't currently see all the way through it, and again it's probably more trouble than it's worth.

PR #174, opened 2020-12-22
[Type Refactor] Use Java type system instead of custom one for typing tensors

karllessard:
This PR is the last of importance in the tensor type refactoring currently in progress and it has a lot of backward incompatible changes.

We drop the current `DataType` custom class to leverage the standard Java type system instead, using `.class`. So now, each of the tensor types (i.e. subinterfaces of `TType`, like `TFloat`) carries itself the information required for allocating and mapping tensors of this type, via the annotation `@TensorType`.

All reference to the static variable `DTYPE` of the different tensor types can now be replaced by `.class`. For example:
```java
// Before
Operand<TFloat32> x = tf.dtypes.cast(tf.constant(1), TFloat32.DTYPE);
assertSame(x.dataType(), TFloat32.DTYPE);
if (x.dataType() == TFloat32.DTYPE || x.dataType() == TFloat64.DTYPE) {
    // floating-point only computations...
}

// After
Operand<TFloat32> x = tf.dtypes.cast(tf.constant(1), TFloat32.class); 
assertSame(x.type(), TFloat32.class);
if (TFloating.class.isAssignableFrom(x.type()) {
    // floating-point only computations...
}
```

In most places where the custom `DataType` class has been dropped, it has been replaced with its [protobuf equivalent](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/proto/framework/DataType.java) of the same name which consists of a simple enum. This provides type information at low level from the runtime library directly, and the API of the `Op` framework will naturally guide the users to favor the usage of the tensor type classes (e.g. `TFloat32.class`) instead.

```java
TFloat32 t = TFloat32.scalarOf(1.0f);
assertSame(t.type(), TFloat32.class);
assertSame(t.dataType(), DataType.DT_FLOAT);
```

In a nut shell, the PR consists of:
* Dropping custom `DataType` and `DataTypes` classes to use Java `Class` instead
* Adding annotation and other utilities to register the tensor type classes (subinterfaces of `TType`)
* Replace all reference to `T*.DTYPE` by `T*.class`
* Reshuffle a little bit some functions of the framework to leverage the new type system
* A lot of changes in generated files that could be skipped during the review (everything under `tensorflow-core/tensorflow-core-api/src/gen`)

It is important to note that after this PR has been merged, the ongoing work from @rnett (Kotlin) and @JimClarke5 (Framework) could be resumed.

CC\ @deansher , @Craigacp


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
```diff
@@ -84,33 +86,37 @@ public ByteDataBuffer data() {
...
    */
-  static RawTensor allocate(DataType<?> dtype, Shape shape, long size) {
-    // Minimum requirements for datatypes of variable length cannot be verified in a relevant way so
-    // we only validate them for fixed length datatypes
-    if (!dtype.isVariableLength() && shape.size() * dtype.byteSize() > size) {
+  static RawTensor allocate(Class<? extends TType> type, Shape shape, long size) {
+    TensorTypeInfo<?> typeInfo = TensorTypeRegistry.find(type);
+    long effectiveSize = size;
+    if (effectiveSize < 0) {
+      effectiveSize = shape.size() * typeInfo.byteSize();
```
Craigacp, line 109: What happens if `typeInfo.byteSize()` returns -1?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+      mapper = (TensorMapper<T>)typeAnnot.mapperClass().newInstance();
+    } catch (ReflectiveOperationException e) {
+      throw new IllegalArgumentException("Class \"" + type.getName() + "\" must have a public "
+          + "parameter-less constructor to be used as a tensor mapper");
+    }
+    TensorTypeInfo<T> typeInfo = new TensorTypeInfo<>(type, typeAnnot.dataType(), typeAnnot.byteSize(), mapper);
+    TYPES_BY_CLASS.put(type, typeInfo);
+
+    // If more than one tensor type is mapped to a given native code, the last registered will
+    // have priority. This way, we can allow user to register their own classes to map tensors
```
Craigacp, line 90: This PR doesn't allow user registered classes though right? If so then it should probably say "in the future we could allow"

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Operands.java
```diff
@@ -16,10 +16,16 @@
 package org.tensorflow.op;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
```
Craigacp, line 20: Unnecessary import of `Iterator`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Operands.java
```diff
@@ -16,10 +16,16 @@
...
 import java.util.List;
 import org.tensorflow.Operand;
 import org.tensorflow.OperationBuilder;
 import org.tensorflow.Output;
+import org.tensorflow.Tensor;
```
Craigacp, line 25: Unnecessary import of `Tensor`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/annotation/TensorType.java
```diff
@@ -0,0 +1,52 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 1: Copyright header is missing the comment `*` before each line.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TIntegral.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.tensorflow.types.family;
+
+/**
+ * Common interface for all integral numeric tensors.
+ *
+ * <p>Operations that only accepts integral values as some of their operands enforce that the tensor
+ * types for these operands to be bound to this interface.
+ */
+public interface TIntegral extends TNumber {}
```
Craigacp, line 25: Is this actually plugged into anything yet?

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/Identity.java
```diff
@@ -33,12 +33,12 @@
...
  *             new org.tensorflow.framework.initializers.Identity&lt;&gt;(tf);
  *     Operand&lt;TFloat32&gt; values =
- *             initializer.call(tf.constant(Shape.of(2,2)), TFloat32.DTYPE);
+ *             initializer.call(tf.constant(Shape.of(2,2)), TFloat32.class);
  * </pre>
  *
  * @param <T> The TType for the call operation
  */
-public class Identity<T extends TType> extends BaseInitializer<T> {
+public class Identity<T extends TFloating> extends BaseInitializer<T> {
```
Craigacp, line 41: This more restrictive, do we want that?

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/LeCun.java
```diff
@@ -15,6 +15,7 @@
 package org.tensorflow.framework.initializers;
 
 import org.tensorflow.op.Ops;
+import org.tensorflow.types.family.TFloating;
 import org.tensorflow.types.family.TNumber;
```
Craigacp, line 19: TNumber import is unnecessary now.

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/TruncatedNormal.java
```diff
@@ -31,13 +30,12 @@
...
  *     Operand&lt;TFloat32&gt; values =
- *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.DTYPE);
+ *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.class);
  * </pre>
  *
  * @param <T> The TType for the call operation
- * @param <U> The TNumber for the call operation
  */
-public class TruncatedNormal<T extends TType, U extends TNumber> extends BaseInitializer<T> {
+public class TruncatedNormal<T extends TNumber> extends BaseInitializer<T> {
```
Craigacp, line 38: Shouldn't this be `TFloating`?

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomUniform.java
```diff
@@ -32,13 +32,12 @@
...
  *     Operand&lt;TFloat32&gt; values =
- *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.DTYPE);
+ *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.class);
  * </pre>
  *
  * @param <T> The TType for the call operation
- * @param <U> The TNumber for the call operation
  */
-public class RandomUniform<T extends TType, U extends TNumber> extends BaseInitializer<T> {
+public class RandomUniform<T extends TNumber> extends BaseInitializer<T> {
```
Craigacp, line 40: `TFloating`?

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomNormal.java
```diff
@@ -31,13 +30,12 @@
...
  *     Operand&lt;TFloat32&gt; values =
- *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.DTYPE);
+ *              initializer.call(tf.constant(Shape.of(2,2)), TFloat32.class);
  * </pre>
  *
  * @param <T> The TType for the call operation
- * @param <U> The TNumber for the call operation
  */
-public class RandomNormal<T extends TType, U extends TNumber> extends BaseInitializer<T> {
+public class RandomNormal<T extends TNumber> extends BaseInitializer<T> {
```
Craigacp, line 38: `TFloating`?

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
```diff
@@ -124,16 +124,13 @@ public Hinge(Ops tf, String name, Reduction reduction) {
   public <T extends TNumber, U extends TNumber> Operand<T> call(
       Operand<U> labels, Operand<T> predictions, Operand<T> sampleWeights) {
```
Craigacp, line 125: `U extends TFloating`?

Craigacp (changes requested):
Would be nice to clean up the import generation, as that's induced a lot of noise in this PR, but it doesn't have to be done here. I think that the framework module could do with a pass to enforce stricter types now TFloating and TIntegral exist, but again that doesn't have to happen here.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/family/TIntegral.java
Craigacp: I found something later on which uses it, but forgot to delete this comment.

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/collective/BroadcastRecv.java
```diff
@@ -17,16 +17,17 @@
...
 import org.tensorflow.Operation;
 import org.tensorflow.OperationBuilder;
 import org.tensorflow.Output;
 import org.tensorflow.ndarray.Shape;
+import org.tensorflow.op.Operands;
 import org.tensorflow.op.RawOp;
 import org.tensorflow.op.Scope;
 import org.tensorflow.op.annotation.Endpoint;
 import org.tensorflow.op.annotation.Operator;
+import org.tensorflow.proto.framework.DataType;
```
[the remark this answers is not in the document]
karllessard: It is, it was incorporated with my last changes and I've found a way to remove it, thanks for catching it. Now the process of [collecting the imported classes](https://github.com/tensorflow/java/blob/3f3f384517a5ba633c3b1971651cdcdec3eebfa7/tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_generator.cc#L78) in the op generator is a bit cumbersome and should be reviewed at some point, I agree. It will be even more necessary when we will (finally) activate lint checks that will report unused imports as an error.

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/bitwise/BitwiseOr.java
```diff
@@ -26,7 +26,6 @@
 import org.tensorflow.op.annotation.Endpoint;
 import org.tensorflow.op.annotation.Operator;
 import org.tensorflow.types.family.TNumber;
-import org.tensorflow.types.family.TType;
```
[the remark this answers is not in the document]
karllessard: I'm also not too sure how this import has been removed but it seems to be a good thing since it was never used...

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/data/experimental/AssertCardinalityDataset.java
```diff
@@ -46,16 +47,12 @@
    * @return a new instance of AssertCardinalityDataset
    */
   @Endpoint(describeByClass = true)
-  public static AssertCardinalityDataset create(Scope scope, Operand<?> inputDataset, Operand<TInt64> cardinality, List<DataType<?>> outputTypes, List<Shape> outputShapes) {
+  public static AssertCardinalityDataset create(Scope scope, Operand<?> inputDataset, Operand<TInt64> cardinality, List<Class<? extends TType>> outputTypes, List<Shape> outputShapes) {
```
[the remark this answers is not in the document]
karllessard: I don't see a `AssertCardinalityDataset` class under `org.tensorflow.op.data`, is it what you meant?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
```diff
@@ -83,12 +84,12 @@ public int inputListLength(final String name) {
...
 
   @Override
-  public TFE_TensorHandle getUnsafeNativeHandle(int outputIndex) {
+  TFE_TensorHandle getUnsafeNativeHandle(int outputIndex) {
     return outputHandles[outputIndex];
   }
 
   @Override
-  public Shape shape(int outputIndex) {
+  Shape shape(int outputIndex) {
```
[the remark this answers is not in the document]
karllessard: It should have been package private from the start, as it extends [AbstractOperation](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/AbstractOperation.java) and it was never intended to override the visibility of its endpoint. So it's just a mistake I've found and fixed while making my changes.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
```diff
@@ -39,23 +41,30 @@ public int index() {
...
-   * @param dt any supported tensor data type
+   * @param type any supported tensor type
    * @throws IllegalArgumentException if the actual data type of this object does not match the type
    *     {@code U}.
    */
   @SuppressWarnings("unchecked")
-  public <U extends TType> Output<U> expect(DataType<U> dt) {
-    if (!dt.equals(this.dataType())) {
+  public <U extends TType> Output<U> expect(Class<U> type) {
+    if (type != type()) {
```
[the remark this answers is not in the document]
karllessard: You mean that the tensor type classes would be loaded multiple times by different class loaders? According to [these guys](https://stackoverflow.com/questions/14257357/loading-a-class-twice-in-jvm-using-different-loaders), unless we do extra-effort to make the classes distinct, the JVM will continue to consider them equal. Do you have some specific use case in mind?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
```diff
@@ -164,7 +171,7 @@ TF_Tensor nativeHandle() {
    * @throws ClassCastException if {@code T} is not compatible type with {@link #dataType()}
    */
   <T extends TType> T asTypedTensor() {
-    return (T)dtype.map(this);
+    return (T)typeInfo.mapper().mapDense(this);
```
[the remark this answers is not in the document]
karllessard: I assumed it would, maybe I'm wrong, I'll add a test for it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
karllessard: Actually this `if/else` is kind of reversed, I'll shuffle it a bit to make sure that won't be a problem.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeRegistry.java
karllessard: Ah true, I was planning to support it then postpone that work as a future project but the comment remained, I'll correct it

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/Identity.java
karllessard: Yes, because the `call` method has this code before:
```java
    if (!dtype.isFloating()) {
      throw new IllegalArgumentException("DataType must be a float type: " + dtype.name());
    }
```

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomUniform.java
karllessard: That one has the following check, so I assume that integral values are also accepted:
```java
    if (!dtype.isNumeric())
      throw new IllegalArgumentException("The data type must be numeric. Found : " + dtype.name());
```

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomNormal.java
karllessard: same thing as https://github.com/tensorflow/java/pull/174#discussion_r547489320

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
karllessard: I did not saw any checks enforcing that the tensors to be of a floating type for `Hinge`, should it?

karllessard:
Thanks a lot for the quick review @Craigacp !

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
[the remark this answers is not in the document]
karllessard: yes, that works

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/bitwise/BitwiseOr.java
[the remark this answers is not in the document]
Craigacp: Ok. The removal of this unnecessary TType import is the reason most of the generated files got touched, so it added a lot of noise to the PR.

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/data/experimental/AssertCardinalityDataset.java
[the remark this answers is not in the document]
Craigacp: Ooops, sorry I meant `AssertNextDataset`. There are a bunch of classes that are in both `org.tensorflow.op.data` and `org.tensorflow.op.data.experimental`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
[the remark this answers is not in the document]
Craigacp: Well if they start off in different `URLClassLoader` instances pointed at the same jar file then they will be different, and I think that could happen inside an application container. I don't think we should ever be able to mix them, but given JavaCPP plays tricks with how the libraries are loaded it might happen. Either way, I think it's fine if we decide we don't care about them coming from different class loaders, but it might be worth noting that this check will fail (so if we do get an issue about this bug then there is a note next to it in the code).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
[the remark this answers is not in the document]
Craigacp: Ok, good.

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
Craigacp: Well it says predictions should be between 0 and 1. I guess it could supply the integers rather than probabilities, but it's kinda against the docs.

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/TruncatedNormal.java
karllessard: Check was only one numbers, [see this comment](https://github.com/tensorflow/java/pull/174#discussion_r547489320)

JimClarke5: TrunctedNormal, RandomNormal should be TFloating. 
RandomUniform can be TInteger and TFloating, so that should be TNumber.

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
JimClarke5: Labels can be -1, or 1, if the are 0,1, they are converted to -1, 1, but python example show this as float.
Python is silent on the Predictions, but all the values are between 0 and 1, so should probably be TFLoating.

JimClarke5:
I the TestSession classes, there is still  a  lot of checking on the types to call the right method for evaluating and printing.  For example compare a float value to TFLoat32 data, double value to TFLoat64 data, etc.  Is there now a better way to do this?

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
JimClarke5: The same types should probably apply to SquaredHinge(labels = -1,1 or 0,1) and CategoricalHinge (labels = 0,1)

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomUniform.java
JimClarke5: RandomNormal should be TFloating.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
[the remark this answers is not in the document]
saudet: I don't see what this has to do with JavaCPP. It uses tricks for its own classes to work around issues that happen in containers and OSGi, but it doesn't do anything about user classes. If you're saying that JavaCPP should also take care of cases like this for user classes, then yeah ok, it doesn't do that, but from what I remember, you were against JavaCPP working around limitations of the JDK just because Oracle doesn't approve of those tricks, even though everyone in the industry is using them, right? If you want to blame JavaCPP for something, stop worrying about what everyone does, and please find something else to complain about.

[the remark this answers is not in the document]
Craigacp: It depends if you're doing a findClass in the generated C++ code, as that will pull down something using a class loader but given javacpp loads the same library into the same address space multiple times it's not clear what class loader you'll get back from jni.

However it's not that likely that you're doing that, so it would be difficult to trigger.

And as usual I'm just pointing out that it's library loading behaviour is explicitly against the JNI spec, so what behaviour we get out of the JVM is undefined. It might work on hotspot today, but it's not guaranteed to always work or always have the same behaviour because it's undefined.

karllessard:
> In the TestSession classes, there is still a lot of checking on the types to call the right method for evaluating and printing. For example compare a float value to TFLoat32 data, double value to TFLoat64 data, etc. Is there now a better way to do this?

I don't think this PR is related to such case but an idea like that, since you know that the supported types are all numeric values, you can probably just cast the boxed value returned by `getObject()` to a `Number` and then use the [`doubleValue`](https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html#doubleValue--) of that number?

For example, where `expected` is a double value,
```java
o.asTensor().scalars().forEach(f -> assertEquals(expected, ((Number)f.getObject()).doubleValue()));
```

JimClarke5:
Let me play with your suggestion.  I am redoing the TestSession classes to add support for Placeholders and Feeds, so now would be a good time to clean it up.

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/data/experimental/AssertCardinalityDataset.java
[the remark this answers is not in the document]
karllessard: These are two different kernels in TensorFlow runtime, like other duplicates found in these package as far as I can see. For example, the op under `data` is using `AssertNextDataset` while the other under `experimental` is using `ExperimentalAssertNextDataset`. So it is correct that they coexist, probably they achieve the same purpose but using different implementations. See [here](https://github.com/tensorflow/tensorflow/blob/f02d09c62cb0664439b8a0a9ff69da6e03b5bbdc/tensorflow/core/ops/experimental_dataset_ops.cc#L33)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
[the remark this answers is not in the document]
karllessard: Ok so I'll classify this as a non-issue for now

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/TruncatedNormal.java
karllessard: Ok so I'll do the corrections for `TruncatedNormal` and `RandomNormal`

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
karllessard: Just to make sure I understood @JimClarke5 , so labels could be any numeric values while predictions and sampleWeights must be restricted to floating-points only in all three losses you've mentioned?

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomUniform.java
karllessard: [According to](https://github.com/tensorflow/java/pull/174#discussion_r547550405) @JimClarke5 , this one should remain valid for any numeral type.

tensorflow-framework/src/main/java/org/tensorflow/framework/initializers/RandomNormal.java
karllessard: [According to](https://github.com/tensorflow/java/pull/174#discussion_r547550405) @JimClarke5 , this one should now be restricted to floating-points only, I'll make the change

JimClarke5:
The labels are restricted to the values -1, 0 and 1.  So restricting it to integers seems appropriate.

Jim

tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Hinge.java
JimClarke5: For these classes, Python does not dictate that they be integers, just that they are value restricted to -1, 0, 1. 
I think if we restrict it to TInteger, it falls more in line with the restriction on the values.

karllessard: Ok so then labels (`U`) is bound to `TIntegral` and predictions/weights (`T`) is bound to `TFloating`, I will make this change and we'll adjust later if needed.

karllessard: But in order to change the bound of these generics, I need to move the parameter at the `Loss` class level instead of the `call` method level, otherwise the Java compiler does not allow me to override them. e.g.
```java
public abstract class Loss<T, U> {
    public abstract Operand<T> call(Operand<U> labels, Operand<T> predictions, Operand<T> sampleWeights);
}

public final class Hinge extends Loss<TFloating, TIntegral> {
    @Override
    public abstract Operand<T> call(Operand<U> labels, Operand<T> predictions, Operand<T> sampleWeights) {
    // ...
    }
}
```
Are we ok with this change or should we leave `Loss` parameter-less? ... or should we simply leave it as is (i.e. only bound to `TNumber`) and review this in a separate PR?

JimClarke5: I don't understand what you need to do? With ` Hinge extends Loss<TFloating, TIntegral>`,
Why doesn't `call(Operand<TInt32> labels, Operand<TFloat32> predictions, Operand<TFloat32> sampleWeigthts)` work?

karllessard: That's the thing, it would work that way. But right now the `Loss` class is unparameterized and the parameters are carried at the [method level](https://github.com/tensorflow/java/blob/3f3f384517a5ba633c3b1971651cdcdec3eebfa7/tensorflow-framework/src/main/java/org/tensorflow/framework/losses/Loss.java#L68). I'm not able to override the generic bounds at the method level, only at the class level but that would require me to do a bit of refactoring like I've presented above.

JimClarke5: Ok, now I see. I would change `Loss` to to be parameterized. That will require touch all the `Loss` subclasses though, but I think that makes more sense.

JimClarke5: BTW: Metric is parameterized, so I guess I am learning :-).

Craigacp: In the interests of getting this PR integrated, lets leave firming up these types for another PR.

karllessard: I agree, so I'm pushing only the other changes that we've discussed.

karllessard:
Ok so I've pushed the last changes that, I think, covers up comments from @Craigacp and @JimClarke5 . 

@deansher , are you interested to take a look as well, as you were a major instigator of that proposal as well? If you want but don't have time right now, just let me know and I'll wait before merging, thanks!

deansher:
Thank you, Karl — I would love to take a look, if waiting until early New
Year's week to merge doesn't seem like a drag.

Dean

karllessard:
@rnett , @JimClarke5 : I don't expect that there will be major changes after Dean's review so if you want to get unblocked faster, I invite you to rebase your work on this PR branch and you can even start producing new PRs that we could merge after this one.

rnett:
I'm working on the `reified` generation for the Kotlin API, and I'm noticing that lots of methods have unnecessary type parameters that makes the `reified` much harder (since you have to specify all type parameters if you specify one).  The best example is probably `cast`, which has the signiture:
```java
<U extends TType, T extends TType> Cast<U> cast(Operand<T> x, Class<U> DstT, Cast.Options... options)
```

`T` is completely unnecessary and could be replaced with `?` without issue, but it prevents `cast<TInx32>(x)` usage from Kotlin.  This shows up in a number of Ops, mostly with the unnecessary type parameters on the input.  It essentially needs a "is this type param only bounded by TType and only used on inputs" check.

It might be out of scope of this PR to change this now, since it probably is best done in the cc generation code (to change the op class `create` methods).

karllessard:
That's very interesting @rnett , and yes it should be addressed outside this PR, I've created a new issue so we can continue the discussion from there: #176

deansher (changes requested):
I love how this came out! I did propose changes, but they are nits.

Thanks, @karllessard , for inviting me back into the loop on this. It's exciting to see it land! What a lot of work!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
```diff
@@ -84,33 +86,39 @@ public ByteDataBuffer data() {
...
    * @param shape shape of the tensor
-   * @param size size of the tensor
+   * @param size size in bytes of the tensor, or -1 to compute the size from the shape
    * @return allocated tensor
    */
-  static RawTensor allocate(DataType<?> dtype, Shape shape, long size) {
-    // Minimum requirements for datatypes of variable length cannot be verified in a relevant way so
-    // we only validate them for fixed length datatypes
-    if (!dtype.isVariableLength() && shape.size() * dtype.byteSize() > size) {
+  static RawTensor allocate(Class<? extends TType> type, Shape shape, long size) {
```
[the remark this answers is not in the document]
karllessard: In fact, passing a totally or partially unknown shape to this constructor should be forbidden, I'll add a check for this.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -57,35 +49,35 @@
...
    * tensor is explicitly set instead of computing it from the datatype and shape, which could be
    * larger than the actual space required to store the data but not smaller.
    *
-   * @param <T> the tensor element type
-   * @param dtype datatype of the tensor
+   * @param <T> the tensor type
+   * @param type the tensor type class
    * @param shape shape of the tensor
-   * @param size size, in bytes, of the tensor
+   * @param size size in bytes of the tensor or -1 to compute the size from the shape
```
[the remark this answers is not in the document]
karllessard: same as in https://github.com/tensorflow/java/pull/174#discussion_r549924041

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeInfo.java
```diff
@@ -0,0 +1,76 @@
...
+ * Registered information about a tensor type.
+ *
+ * @param <T> the tensor type
+ */
+public final class TensorTypeInfo<T extends TType> {
+
+  /**
+   * Returns the class of this tensor type
+   */
+  public Class<T> typeClass() {
```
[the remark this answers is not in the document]
karllessard: I was hesitating on this one as well so sounds like we are two now, meaning that it should be `type()`, I'll rename it.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/RawTensorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class RawTensorTest {
+
+  @Test
+  public void rawToTypedTensor() {
+    RawTensor rawTensor = RawTensor.allocate(TFloat32.class, Shape.of(2, 2), -1);
+    TFloat32 floatTensor = rawTensor.asTypedTensor();
+    assertSame(floatTensor.asRawTensor(), rawTensor);
+    try {
+      TInt32 intTensor = rawTensor.asTypedTensor();
```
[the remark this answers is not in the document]
karllessard: Good point, I've never tried that before. You want me to update the doc or try to throw an `ClassCastException` in this case as well? 

Note that `asTypedTensor()` is an internal package-private method and where it is currently being used, it is implicitly enforced that the returned type matches the type of the tensor for all cases.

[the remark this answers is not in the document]
deansher: Updating the doc seems good to me for both of the reasons you give.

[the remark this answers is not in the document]
karllessard: To be honest I'm not too sure how to document this. It looks to me that this behavior is some kind of "defect" in the generic specification, since the explicit `<TInt32>` parameterization of the method invocation seems to be overridden implicitly to the type inferred by the target (`Object` in this case). 

So basically, `<TInt32>` is ignored and that is probably the case for Java type inference in general when dealing with type parameters inferred by a target, is it worth documenting it here then?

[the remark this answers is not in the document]
Craigacp: The type will be erased at runtime so the cast is erased to its bound (`TType`) which always succeeds. It might be better to have it explicitly return `TType` rather than have it promise something that can't be enforced by the type system. This is going to pollute the internals of our code with casts, but at least we'll remember to put them in rather than having it mysteriously blow up.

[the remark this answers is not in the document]
karllessard: Ok sounds fair, let's do that, anyway all these additional casts will happen internally only.

karllessard:
I've just pushed a last version that should now covered all discussed topics during the review, I'll merge it as soon as I receive a green light from your behalf and we will be done with that massive refactoring (and hopefully the last one of this nature).

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/RawTensorTest.java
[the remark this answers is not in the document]
deansher: On further reflection, I lean the same way as @Craigacp: have this method return `TType`.

Here's some analysis to support this choice: At runtime, the current behavior is "apply the mapper from `typeInfo` and return whatever `TType` it produces." Since the type parameter `T` is erased at runtime, the only way to guarantee a return of `T` would be to have the method take the tensor type class as a runtime parameter and explicitly verify that the mapper returns that type class. I might advocate this approach for a public method or even a widely used package-private method, but not for a rarely used package-private method.

But also, a review of the call points of this method raises another issue: whenever we treat the return value of `asTypedTensor` as simply a `Tensor` (rather than a `TType`), we "forget" at compile time the semantic upgrade that was presumably the point of the `asTypedTensor` call in the first place.

As an example, let's explore what happens underneath the following method:
~~~ Java
class EagerOperation extends AbstractOperation {
  // ...
  /**
   * Returns the tensor of the {@code outputIdx}th output of this operation.
   *
   * <p>This is only supported in an eager execution environment.
   *
   * @param outputIdx index of the output of this operation
   * @return output tensor
   */
  @Override
  Tensor tensor(int outputIndex) {
    Tensor tensor = outputTensors.get(outputIndex);
    if (tensor == null) {
      tensor = resolveTensor(outputIndex);
    }
    return tensor;
  }
  // ...
}
~~~
(I copied the Javadoc above the `@Override` for easy reference.)

The caller should presumably be agnostic as to whether this returns a `RawTensor` or a `TType`. But looking deeper into the call path, here's how the returned `Tensor` is  constructed:
~~~ Java
  private static Tensor resolveTensorHandle(TFE_TensorHandle handle, EagerSession session) {
    requireTensorHandle(handle);
      TF_Status status = TF_Status.newStatus();
      TF_Tensor tensor = TFE_TensorHandleResolve(handle, status).withDeallocator();
      status.throwExceptionIfNotOK();
      return RawTensor.fromHandle(tensor, session).asTypedTensor();
    }
  }
~~~
What's the point of `asTypedTensor()` on the last substantive line, above? It causes this method to return a special kind of `Tensor` -- a `TType` -- but that fact is immediately forgotten by the type system and is not even asserted in the Javadoc. Presumably, we should either drop the call to `.asTypedTensor()` or change the whole call path to return `TType`.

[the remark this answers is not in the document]
deansher: :-) But not for this PR!

deansher (approved):
The Eagle has landed.

JimClarke5 (approved):
Good job @karllessard

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeRegistry.java
```diff
@@ -0,0 +1,104 @@
...
+public final class TensorTypeRegistry {
+
+  /**
+   * Find registered information about a tensor type from its equivalent data type
+   *
+   * @param dataType data type
+   * @return type registered information
+   * @throws IllegalArgumentException if no tensor type for this data type has been registered
+   */
+  public static <T extends TType> TensorTypeInfo<T> find(DataType dataType) {
```
Craigacp, line 47: This is a place where it will infer the type bound from the calling context, and if that's wrong then we'll get a weird class cast error when people use it. It might be better to return the wildcard as at least the user will get a warning when they make the cast.

Craigacp (approved):
The typing in `TensorTypeRegistry` should probably be relaxed as I've indicated, but that can wait till the next PR.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeRegistry.java
karllessard: Again note that the `TensorTypeRegistry` is (for now) an internal class so we do have some control over the context it is being called. But like you've suggested, let's review this later.

karllessard:
All right, let's merge this before someone changes his mind :) Thank you all for your reviews and comments!

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/registry/TensorTypeRegistry.java
Craigacp: Yeah, I know. But I'm in favour of having the compiler warn me before I make a silly mistake, and 6 months from now I won't necessarily remember that this method doesn't quite live up to its contract.

PR #178, opened 2020-12-29
Initialization imprvements

rnett:
Makes three simple improvements to initialization:
 * Makes `Ops.initAdd` no-op in eager mode, since the op has already been ran.
 * Adds a `runInit()` method to `Session` that runs the graph's initializers.
 * Adds a `doInitialization()` method to `Session.Runner` to add the initializers as targets.
 
There's no changes that conflict w/ the type refactor, so I'll just wait till it's merged and rebase onto master.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
+++ b/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
@@ -2755,7 +2755,7 @@ public final class Ops {
    *
    *  <p>Registered initializers are then grouped as a single unit of computation by adding
    *  and executing an {@link org.tensorflow.op.core.Init#create(Scope) init} operation from a graph
-   *  session.  No-ops if the session is eager.
+   *  session.  This is a no-op if executed in an eager session.
    *
    * @param scope
    * @param initializer
--- a/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
+++ b/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
@@ -347,9 +347,9 @@ public final class Ops {
 
   public final SignalOps signal;
 
-  public final QuantizationOps quantization;
-
   public final TrainOps train;
+
+  public final QuantizationOps quantization;
 
   private final Scope scope;
 
@@ -372,8 +372,8 @@ public final class Ops {
     math = new MathOps(this);
     audio = new AudioOps(this);
     signal = new SignalOps(this);
-    quantization = new QuantizationOps(this);
     train = new TrainOps(this);
+    quantization = new QuantizationOps(this);
   }
 
   /**
@@ -2755,11 +2755,10 @@ public final class Ops {
    *
    *  <p>Registered initializers are then grouped as a single unit of computation by adding
    *  and executing an {@link org.tensorflow.op.core.Init#create(Scope) init} operation from a graph
-   *  session.
+   *  session.  No-ops if the session is eager.
    *
    * @param scope
    * @param initializer
-   * @throws IllegalArgumentException if the execution environment in scope is not a graph
    * @see org.tensorflow.op.core.Init#create(Scope) init
    */
   public void initAdd(Op initializer) {
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
@@ -294,6 +294,16 @@ public final class Session implements AutoCloseable {
     }
 
     /**
+     * Make {@link #run} execute the graph's initializers.
+     *
+     * @return this session runner
+     */
+    public Runner doInitialization(){
+      graph.initializers().forEach(this::addTarget);
+      return this;
+    }
+
+    /**
      * Set options (typically for debugging) for this run.
      *
      * <p>The options are presented as a <a
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
@@ -294,16 +294,6 @@ public final class Session implements AutoCloseable {
     }
 
     /**
-     * Make {@link #run} execute the graph's initializers.
-     *
-     * @return this session runner
-     */
-    public Runner doInitialization(){
-      graph.initializers().forEach(this::addTarget);
-      return this;
-    }
-
-    /**
      * Set options (typically for debugging) for this run.
      *
      * <p>The options are presented as a <a
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
@@ -490,6 +490,19 @@ public final class Session implements AutoCloseable {
     runner().addTarget(op.op()).run();
   }
 
+
+  /**
+   * Execute the graph's initializers.
+   *
+   * <p>This method is equivalent to {@code session.runner().addTarget(Ops.create(session.graph()).init()).run()}.
+   *
+   */
+  public void runInit(){
+    Runner runner = runner();
+    graph.initializers().forEach(runner::addTarget);
+    runner.run();
+  }
+
   /**
    * Saves the actual state of the variables of this session's graph.
    *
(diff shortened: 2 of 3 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java
```diff
@@ -89,16 +89,18 @@ public static Init create(Scope scope) {
...
    * @param scope
    * @param initializer
-   * @throws IllegalArgumentException if the execution environment in scope is not a graph
    * @see org.tensorflow.op.core.Init#create(Scope) init
    */
   @Endpoint(name = "initAdd")
   public static void add(Scope scope, Op initializer) {
     ExecutionEnvironment exEnv = scope.env();
+    if(exEnv.isEager()){
+      return;
```
Craigacp, line 102: Why leave the exception in there if it's not triggered?

rnett: Because in theory you could have a non-eager execution environment that isn't Graph.  Is there any possibility of that being a thing?

Craigacp: Not to my knowledge. There are a lot of other things that assume either eager or graph execution and that would blow up if there was a third type. At the very least the exception error message should be fixed to note that this op succeeds on graph and eager sessions.

karllessard:
I remember that we we've added the [Init](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java) ops, we discussed about having an specialized endpoint like this to initialize the variables but then ended up deciding that we should have something more generic that allows you to run any single op in a session.
```java
try (Graph g = new Graph()) {
  Variable<TInt32> x = tf.variable(tf.constant(10));
  Variable<TInt32> y = tf.variable(tf.constant(20));

  try (Session s = new Session(g)) {
    s.run(tf.init());
    ...
   }
}
```
Just double checking, is there any specific reason to go back to a specialized endpoint for variable initialization only? @Craigacp , any comment on this?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java
```diff
@@ -89,18 +89,20 @@ public static Init create(Scope scope) {
...
+    }
     if (!(exEnv instanceof Graph)) {
-      throw new IllegalArgumentException("initAdd is only supported on Graph sessions.");
+      throw new IllegalArgumentException("initAdd is only supported on Graph or eager sessions.");
```
karllessard, line 105: I think pretty much all over the code, while we "might" have to support later a third execution environment type, we kind of assume that there is only to. So I would only simplify this method with:
```java
if (exEnv.isGraph()) {
   Graph graph = (Graph) exEnv;
   graph.addInitializer(initializer);
}
```

rnett:
> Just double checking, is there any specific reason to go back to a specialized endpoint for variable initialization only?

Two reasons: discoverability that initialization is a thing that needs to be ran (i.e. if you depend on a network that uses init that you didn't write, you may not even know it exists.  I got burnt by forgetting to run it a few times), and doing `s.run(tf.init())` creates a new init op each time, where `runInit()` just runs each initialization op w/o creating anything.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -293,6 +293,16 @@ public Runner addTarget(Op op) {
       return addTarget(op.op());
     }
 
+    /**
+     * Make {@link #run} execute the graph's initializers.
+     *
+     * @return this session runner
+     */
+    public Runner doInitialization(){
```
Craigacp, line 301: This should be called `addInitialization`, and I'm not sure it's a great idea to do this in a runner with other things, so it might want to be a package-private method only visible through the `runInit` method, then fix the docs for `runInit` so it doesn't reference a private method.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java
```diff
@@ -89,21 +89,19 @@ public static Init create(Scope scope) {
    *
    * <p>Registered initializers are then grouped as a single unit of computation by adding
    * and executing an {@link org.tensorflow.op.core.Init#create(Scope) init} operation from a graph
-   * session.
+   * session.  No-ops if the session is eager.
```
Craigacp, line 92: "This is a no-op if executed in an eager session"

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -2755,11 +2755,10 @@ public Init init() {
    *
    *  <p>Registered initializers are then grouped as a single unit of computation by adding
    *  and executing an {@link org.tensorflow.op.core.Init#create(Scope) init} operation from a graph
-   *  session.
+   *  session.  No-ops if the session is eager.
```
Craigacp, line 2758: "This is a no-op if executed in an eager session"

Craigacp (commented):
I think I'm fine with adding this back in, the previous discussion about it happened here - https://github.com/tensorflow/java/pull/36#issuecomment-605657388

Craigacp:
> > Just double checking, is there any specific reason to go back to a specialized endpoint for variable initialization only?
> 
> Two reasons: discoverability that initialization is a thing that needs to be ran (i.e. if you depend on a network that uses init that you didn't write, you may not even know it exists. I got burnt by forgetting to run it a few times), and doing `s.run(tf.init())` creates a new init op each time, where `runInit()` just runs each initialization op w/o creating anything.

All TF 1 networks require the init op to be executed, it's been dropped with TF 2 in Python, but that's because the variables changed and things are implicitly eager whereas we still have Graph & Eager modes which are quite different. Until we paper over that more (and it might not be possible to do so without more improvements in the C API) then TF Java will still always need the init operator.

karllessard:
Again, if we focus on the functional API instead, we can probably call `session.run(tf.init())` implicitly when `function.call` is invoked, or something like that?

rnett:
Depends on how you store the session in the ConcreteFunction, but yeah probably.

Idk about Python TF2, but I'm still using it for variable initialization in my PR for variables, and I don't see a way around it.  Python uses `init_scope()` (as a context manager) which I think is the same or similar, I don't know if we have an equivilant.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
rnett: I'll just remove it if we don't want it public, I wasn't using it in `runInit`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -490,6 +490,19 @@ public void run(Op op) {
     runner().addTarget(op.op()).run();
   }
 
+
+  /**
+   * Execute the graph's initializers.
+   *
+   * <p>This method is equivalent to {@code session.runner().addTarget(Ops.create(session.graph()).init()).run()}.
```
karllessard, line 497: In fact, it is more a replacement for `session.run(Ops.create(session.graph).init())`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java
```diff
@@ -89,21 +89,19 @@ public static Init create(Scope scope) {
...
-   * @throws IllegalArgumentException if the execution environment in scope is not a graph
    * @see org.tensorflow.op.core.Init#create(Scope) init
    */
   @Endpoint(name = "initAdd")
   public static void add(Scope scope, Op initializer) {
     ExecutionEnvironment exEnv = scope.env();
-    if (!(exEnv instanceof Graph)) {
-      throw new IllegalArgumentException("initAdd is only supported on Graph sessions.");
+
+    if(exEnv.isGraph()) {
```
karllessard, line 102: Nit: there is a missing space after `if`

karllessard:
Thanks @rnett , I like this simplified version and I personally like to make it easier to initialize the variables. I've left a few minor comments, if you want to take a look before I merge.

Still to be discuss, I would like to see at some point the Functional API taking care of initializing the variables once before calling the graph without the need for the user to do it.

karllessard (dismissed):

rnett:
I'll change those.

I agree about variable initialization, we could have the session track whether it's initialized, and run it on the first run call if not?  We'd probably want to prevent the manual usage of the `init` op then.

karllessard (approved):

PR #182, opened 2021-01-05
Expose input and output types from Signature

jbentleyEG:
Currently you can get the names of the inputs and outputs from a Signature, but there is no publicly accessible method that allows you to get the return types other than toString().

I was unable to get the tests building on my machine, but hopefully this change is small enough that it won't be an issue.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Signature.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Signature.java
@@ -36,9 +36,9 @@ public class Signature  {
 
   public static class TensorDescription {
     public final DataType dataType;
-    public final long[] shape;
+    public final Shape shape;
 
-    public TensorDescription(DataType dataType, long[] shape) {
+    public TensorDescription(DataType dataType, Shape shape) {
       this.dataType = dataType;
       this.shape = shape;
     }
@@ -189,7 +189,10 @@ public class Signature  {
   private Map<String, TensorDescription> buildTensorDescriptionMap(Map<String, TensorInfo> dataMapIn) {
     Map<String, TensorDescription> dataTypeMap = new HashMap<>();
     dataMapIn.forEach((a, b) -> {
-      dataTypeMap.put(a, new TensorDescription(b.getDtype(), b.getTensorShape().getDimList().stream().mapToLong(d -> d.getSize()).toArray()));
+      long[] tensorDims = b.getTensorShape().getDimList().stream().mapToLong(d -> d.getSize()).toArray();
+      Shape tensorShape = Shape.of(tensorDims);
+      dataTypeMap.put(a, new TensorDescription(b.getDtype(),
+                                               tensorShape));
     });
     return dataTypeMap;
   }
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Signature.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Signature.java
@@ -15,9 +15,11 @@
  */
 package org.tensorflow;
 
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


karllessard:
Thanks @jbentleyEG , looks good but do you have an example on how you actually use it? (maybe adding a small unit test could be great)

Also I'm wondering what is better: return types mapped to their tensor name, like you did, or simply return the whole `TensorInfo` input/output list, which would also expose their shape, wdyt? Would it also fit in your use case?

jbentleyEG:
Agreed, I will re-write it to expose the shape as well. That isn't required for my use case, but I imagine that others will need that.

How about just making it a pass-through to the signatureDef method:

 ```
 /**
   * Returns the names of the inputs in this signature mapped to the expected tensor
   */
  public Map<String, TensorInfo> getInputs() {
    return signatureDef.getInputsMap();
  }
```

**Comment by jbentleyEG** [comment]
@googlebot I signed it!

**Comment by karllessard** [comment]
> Agreed, I will re-write it to expose the shape as well. That isn't required for my use case, but I imagine that others will need that.
> 
> How about just making it a pass-through to the signatureDef method:
> 
> ```
> /**
>   * Returns the names of the inputs in this signature mapped to the expected tensor
>   */
>  public Map<String, TensorInfo> getInputs() {
>    return signatureDef.getInputsMap();
>  }
> ```

That is what I had initially in mind as well, but now I'm thinking that we might want to wrap up that `TensorInfo` class for our own so we can return a `Shape` instead of a `TensorShapeProto`, since the former is being used everywhere in the Java client.

For example we could wrap and convert this in a inner class, like `Signature.TensorInfo` or `Tensor.Metadata`? Just throwing some ideas. And to limit the amount of work, I would simply focus on the datatype (this one is fine the return the `DataType` enum found in the protos) and the shape (converted to a `Shape`), leaving the other fields of `TensorInfo` for later. Wdyt?

**Comment by jbentleyEG** [comment]
That makes sense. My original concern about just exposing the whole thing is that it's a mutable object, an inner class bypasses that. I'll submit an update later today.

**Comment by rnett** [comment]
For tests, did you try running maven with the target `integration-test` instead of `test` (in `tensorflow-core-api`, with the `dev` profile)?  I ran into the same thing earlier.

**Comment by karllessard** [comment]
Btw @jbentleyEG , do you still intend to push a new PR for this?

**Comment by rnett** [comment]
@jbentleyEG @karllessard I can take this over, if necessary.

**Comment by jbentleyEG** [comment]
Sorry, got pulled onto another task and got distracted away from this. Yes, I still plan to do a PR for this.

**Comment by jbentleyEG** [comment]
@karllessard PR updated

File: tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Signature.java [hunk 1]
```diff
@@ -32,6 +34,16 @@
   /** The default signature key, when not provided */
   public static final String DEFAULT_KEY = "serving_default";
 
+  public static class TensorDescription {
+    public final DataType dataType;
+    public final long[] shape;
```
karllessard, line 39: Hi @jbentleyEG , looks good but as we discussed, can we return the shape as a `Shape` object to the users instead of a `long[]`?

karllessard: Hi @jbentleyEG, just checking, do you agree to make the proposed change above?

jbentleyEG: Sorry for the flakiness, I keep getting pulled away by other tasks. I wasn't able to generate the Shape directly from the TensorShapeProto, but I was able to recreate it from the dimensions.

karllessard (approved):
Thanks @jbentleyEG !

PR #191, opened 2021-01-22
Created TensorFormat enum

sevarac:
Created enum for TensorFormat. Some additional enum values are open for discussion, at the moment there are values that are used by Tensorflow and cuDnn

### DIFF
--- a/ndarray/src/main/java/org/tensorflow/ndarray/buffer/layout/TensorFormat.java
+++ b/ndarray/src/main/java/org/tensorflow/ndarray/buffer/layout/TensorFormat.java
@@ -1,0 +1,30 @@
+package org.tensorflow.ndarray.buffer.layout;
+
+/**
+ * Specifies the data format in tensor.
+ * 
+ * With the default format "NHWC", the data is stored in the order of:
+ *     [batch, height, width, channels].
+ * 
+ * Alternatively, the format could be "NCHW", the data storage order of:
+ *     [batch, channels, height, width].
+ * 
+ * Additional format NCHW_VECT_C is specified by 
+ * https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnTensorFormat_
+ * although not sure if it is used or needed in tf
+ * 
+ *  Even More formats are specified in https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#WXWZ-tensor-descriptor
+ *  CHWN   4d tensor description
+ *  NCDHW  5d tensor description
+ *  NDHWC
+ *  CDHWN
+ * 
+ * https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnTensorFormat_t
+ * 
+ // https://github.com/tensorflow/java/blob/f85623ed366d903cfddb75177725dc276f359b15/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/nn/MaxPoolGradGrad.java
+
+ */
+public enum TensorFormat {
+    NCHW, 
+    NHWC;
+}


Craigacp:
I don't think we can look at PRs until the CLA has been signed, so that's what this is waiting on.

sevarac:
@googlebot I signed it


др Зоран Шеварац | Zoran Ševarac, PhD
Катедра за софтверско инжењерство | Department of Software Engineering
Ванредни професор | Associate Professor
[Fakultet organizacionih nauka]
Универзитет у Београду | University of Belgrade
Факултет организационих наука | Faculty of Organizational Sciences
Јове Илића 154, Београд, Србија | Jove Ilića 154, Belgrade, Serbia

t: +381 11 3950 853
e: zoran.sevarac@fon.bg.ac.rs<mailto:zoran.sevarac@fon.bg.ac.rs>
w: http://www.fon.bg.ac.rs
________________________________

JimClarke5 (commented):
Is there a reason we shouldn't add the other four enums as described in the link?

- CHWN   4d tensor description
- NCDHW  5d tensor description
- NDHWC
- CDHWN

sevarac:
The main question is are they are used by tensorflow at all? Maybe someone from Tensorflow team could answer this.
I've found these layouts in cuDNN documentation so there is good chance that they are used. 
It would be best to add them too if they are used, otherwise not sure if it makes sense to add them

ndarray/src/main/java/org/tensorflow/ndarray/buffer/layout/TensorFormat.java
```diff
@@ -0,0 +1,30 @@
+package org.tensorflow.ndarray.buffer.layout;
```
karllessard, line 1: Thanks @sevarac , as we discussed briefly during our last session, this enum should probably be moved at the `tensorflow-core-api` or `tensorflow-framework` level. It would be helpful to know which one if you can provide some quick examples of usage. I think @JimClarke5 had some in mind too.

sevarac: Sounds good, in general everywhere where  public Options dataFormat(String dataFormat)  is used 
there should be now public Options dataFormat(TensorFormat dataFormat) 
which includes a bunch of classes mainly layers  https://github.com/tensorflow/java/search?q=dataFormat

JimClarke5: It is used in `losses.Losses` for `losses.CategoricalCrossentropy` and `metrics.CategoricalCrossentropy`.
```
  public static final int CHANNELS_LAST = -1;
  public static final int CHANNELS_FIRST = 1;
```

Once this PR is merged, I'll change the logic in losses and metrics.

karllessard: This could be done in the C++ op generator, by looking at any argument called `dataFormat`. Though I don't think this naming convention is enforced to the kernel developers, which might lead to mistakes. But there will be possible workarounds as well so I'm fine if you want to give a try making that change as well

JimClarke5:
BTW: I will also need this in `layers`.

ndarray/src/main/java/org/tensorflow/ndarray/buffer/layout/TensorFormat.java
karllessard: Hi @sevarac , so any chance that you can move this enum to a different location, as [I've suggested before](https://github.com/tensorflow/java/pull/191/files#r574243992)?

If I need to pick one, I'll suggest`tensorflow-framework` over `tensorflow-core-api`, what about under [`org.tensorflow.framework.utils`](https://github.com/tensorflow/java/tree/master/tensorflow-framework/src/main/java/org/tensorflow/framework/utils)?

karllessard: Also please don't forget to add the header notice in your file (like [this one](https://github.com/tensorflow/java/blob/master/tensorflow-framework/src/main/java/org/tensorflow/framework/utils/ShapeUtils.java), for example).

karllessard (changes requested):

PR #192, opened 2021-01-22
Add docs generation script.

MarkDaoust:
Here's a basic working version of a docs-generation script for tensorflow.org.

Is this new tools/ directory the right place to put this script?
Does this need a Op generation step like the old tensorflow java api?:

https://www.tensorflow.org/api_docs/java/org/tensorflow/op/core/package-summary
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docs/build_java_api_docs.py#L64-L75

The script generates this file structure:

https://gist.github.com/MarkDaoust/f3abf18ceef62c8f3d4fbbec0938a3f0

Does anything look out of place to you?

If you want to try it yourself you'll need to install [tensorflow's java-doc generator](https://github.com/tensorflow/docs/tree/master/tools/tensorflow_docs/api_generator/gen_java) with:

```
pip install git+https://github.com/tensorflow/docs
```

### DIFF
--- a/tools/build_java_api_docs.py
+++ b/tools/build_java_api_docs.py
@@ -1,0 +1,72 @@
+# Lint as: python3
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Generate TensorFlow Lite Java reference docs for TensorFlow.org."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import pathlib
+import shutil
+import tempfile
+
+from absl import app
+from absl import flags
+
+from tensorflow_docs.api_generator import gen_java
+
+FLAGS = flags.FLAGS
+
+# These flags are required by infrastructure, not all of them are used.
+flags.DEFINE_string('output_dir', '/tmp/java_api/',
+                    ("Use this branch as the root version and don't"
+                     ' create in version directory'))
+
+flags.DEFINE_string('site_path', 'java/api_docs/java',
+                    'Path prefix in the _toc.yaml')
+
+flags.DEFINE_string('code_url_prefix', None,
+                    '[UNUSED] The url prefix for links to code.')
+
+flags.DEFINE_bool(
+    'search_hints', True,
+    '[UNUSED] Include metadata search hints in the generated files')
+
+# __file__ is the path to this file
+TOOLS_DIR = pathlib.Path(__file__).resolve().parent
+REPO_ROOT = TOOLS_DIR.parent
+
+
+def main(unused_argv):
+  merged_source = pathlib.Path(tempfile.mkdtemp())
+  (merged_source / 'java/org').mkdir(parents=True)
+
+  shutil.copytree(REPO_ROOT/'tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/',
+                  merged_source/'java/org/tensorflow')
+  shutil.copytree(REPO_ROOT/'tensorflow-framework/src/main/java/org/tensorflow/framework',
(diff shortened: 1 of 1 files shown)


Craigacp:
It's missing the generated files. They are checked in in `tensorflow-core/tensorflow-core-api/src/gen/java` and need to be included in the docs generation. As they are checked in we don't need an ops generation step as the source root should always be current.

I mostly got this working but couldn't quite figure out how it was pulling in the css and styling (probably because my build environment was Ubuntu not Google's version of Debian). I'll pull out the version I had tomorrow and see if there is anything else missing.

MarkDaoust:
> They are checked in in tensorflow-core/tensorflow-core-api/src/gen/java

Done. The most recent commit overlays those files onto the others before generating the docs. Seems to work. Both the generated and [non-generated ops](https://github.com/tensorflow/java/tree/master/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/nn) seem to be there in the new output listing:

https://gist.github.com/MarkDaoust/020b56743a71eb7c6a34cb7e9a0a7e0f

karllessard:
Thanks @MarkDaoust , I think `/tools` is fine for now. @Craigacp , is there anything else you wanted to check before merging the script?

Craigacp:
I looked through my version of this, and I had to modify the shell script to add a bunch of extra jars (JavaCPP and it's dependencies, and protobuf) as otherwise javadoc couldn't resolve the classes properly. Do the AbstractTF* classes look like they were generated right?

MarkDaoust:
I do get these sorts of warnings when I run it:

```
Constructing Javadoc information...
/tmp/tmpqinpxe9o/java/org/tensorflow/Output.java:19: error: package org.bytedeco.javacpp does not exist
import org.bytedeco.javacpp.Pointer;
                           ^
/tmp/tmpqinpxe9o/java/org/tensorflow/proto/framework/DataType.java:15: error: package com.google.protobuf does not exist
    implements com.google.protobuf.ProtocolMessageEnum {
                                  ^
```

>  I had to modify the shell script that is called by gen_java to add a bunch of extra jars (JavaCPP and it's dependencies, and protobuf)

If you've still got that code handy please send it as a PR to https://github.com/tensorflow/docs where that code lives.

Or just post the code here and I'll take care of it.

saudet:
> I looked through my version of this, and I had to modify the shell script that is called by `gen_java` to add a bunch of extra jars (JavaCPP and it's dependencies, and protobuf) as otherwise javadoc couldn't resolve the classes properly. Do the AbstractTF* classes look like they were generated right?

No, those are handwritten. Anything that isn't in `src/gen` isn't generated.

Craigacp:
> > I looked through my version of this, and I had to modify the shell script that is called by `gen_java` to add a bunch of extra jars (JavaCPP and it's dependencies, and protobuf) as otherwise javadoc couldn't resolve the classes properly. Do the AbstractTF* classes look like they were generated right?
> 
> No, those are handwritten. Anything that isn't in `src/gen` isn't generated.

I meant did the javadoc for those classes look like it was generated correctly, which given Mark's response it presumably isn't.

I had to add this to the doclet path:
```bash
JARS=antlr-complete-3.5.2.jar:doclava-android-6.0.1.jar:guava-29.0-jre.jar:jsilver-android-6.0.1.jar:tagsoup-1.2.1.jar
```
and these to the classpath:
```bash
javacpp-1.5.4.jar:protobuf-java-3.8.0.jar:osgi.annotation-7.0.0.jar
```

I think the doclet jars might just be an artifact of the fact that I couldn't replicate the environment the docs build in. The classpath ones are because those are referenced from classes.

I dunno why it needs the osgi one, it's referenced by JavaCPP and would crash without it, but I couldn't figure out why it was necessary. @saudet any idea?

saudet:
@timothyjward added a couple of annotations in the code for OSGi, but I don't know why that would make it crash...

Craigacp:
It might need to reference them if they appear transitively on subclasses.

timothyjward:
> @timothyjward added a couple of annotations in the code for OSGi, but I don't know why that would make it crash...

Those are class retention annotations used to generate OSGi metadata in the manifest of JavaCPP. They don't exist at runtime, just in the class bytecode. The normal rules for annotation processing are to ignore the annotations if you can't load them or don't understand them, so I can see why there might be a warning (at most) from another tool, but not a crash.

karllessard:
I think we are good to merge this now, even if we don't have a clear idea why we need osgi jar to prevent the tool from breaking, as long as it works its fine. Thanks @MarkDaoust !

karllessard (approved):

MarkDaoust:
Thanks guys. I'll try staging this and I'll see if I can get back you with a link where you can preview it.

PR #193, opened 2021-01-23
Remove extra generics from op generation

### LINKED ISSUE #176: Simplify generic parameterization on some operations
I'm working on the `reified` generation for the Kotlin API, and I'm noticing that lots of methods have unnecessary type parameters that makes the `reified` usage much less nice (since you have to specify all type parameters if you specify one).  The best example is probably `cast`, which has the signiture:
```java
<U extends TType, T extends TType> Cast<U> cast(Operand<T> x, Class<U> DstT, Cast.Options... options)
```

`T` is completely unnecessary and could be replaced with `?` without issue, but it prevents `cast<TInx32>(x)` usage from Kotlin.  This shows up in a number of Ops, mostly with the unnecessary type parameters on the input.  It essentially needs a "is this type param only bounded by TType and only used on inputs" check.

_Originally posted by @rnett in https://github.com/tensorflow/java/issues/174#issuecomment-751537426_

rnett:
Fixes #176.  I leave a type parameter in if it is used by two or more inputs or outputs.


tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_specs.cc
```diff
@@ -391,8 +391,64 @@ OpSpec OpSpec::Create(const OpDef& op_def, const ApiDef& api_def) {
   for (const auto& endpoint_def : api_def.endpoint()) {
     op.endpoints_.push_back(CreateEndpoint(op_def, api_def, endpoint_def));
   }
+  op.remove_extra_generics();
   return op;
 }
 
+void OpSpec::remove_extra_generics() {
```
karllessard, line 398: Please use CamelCasing for method names in C++: `RemoveExtraGenerics`

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_specs.cc
```diff
@@ -391,8 +391,64 @@ OpSpec OpSpec::Create(const OpDef& op_def, const ApiDef& api_def) {
...
+  std::map<string, int> generics;
+
+  for (const ArgumentSpec& output : this->outputs()) {
+    if (output.type().kind() == Type::GENERIC && !output.type().wildcard()) {
+      if(generics.find(output.type().name()) == generics.end()) {
```
karllessard, line 403: Nit: missing space after `if` (and in a few other places in the same method)

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_specs.cc
```diff
@@ -391,8 +391,64 @@ OpSpec OpSpec::Create(const OpDef& op_def, const ApiDef& api_def) {
...
+    if (input.type().kind() == Type::GENERIC && !input.type().wildcard()) {
+      if(generics.find(input.type().name()) == generics.end()) {
+        generics[input.type().name()] = 1;
+      } else {
+        generics[input.type().name()] = generics.find(input.type().name())->second + 1;
+      }
+    }
+  }
+
+
```
karllessard, line 421: Nit: extra line

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_specs.cc
```diff
@@ -391,8 +391,64 @@ OpSpec OpSpec::Create(const OpDef& op_def, const ApiDef& api_def) {
...
+  for (ArgumentSpec& output : this->outputs_) {
+    if (output.type().kind() == Type::GENERIC && !output.type().wildcard()) {
+      if(generics.find(output.type().name()) == generics.end() || generics[output.type().name()] <= 1) {
```
karllessard, line 424: it sounds to me that it is not possible that `generics.find(output.type().name()) == generics.end()` here, let's remove that condition to avoid confusion. Same thing below.

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_specs.h
```diff
@@ -151,6 +152,7 @@ class OpSpec {
   const std::vector<AttributeSpec>& optional_attributes() const {
     return optional_attributes_;
   }
+  void remove_extra_generics();
```
karllessard, line 155: This method can and should be private

karllessard (changes requested):
Great job @rnett , I like this improvement. Just a few nits to look and we are good to merge.

karllessard:
Following your [comments](https://github.com/tensorflow/java/issues/176#issuecomment-766194814) about missing files, it is not a concern. When we upgrade the TF runtime version, we need to delete the `src/gen` folder before regenerating the ops but that has not always been done. So there are a few remnants of the previous versions that you are getting rid of in this PR.

For example, `TensorScatterAdd` has been [renamed](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_TensorScatterAdd.pbtxt) to `TensorScatterNdAdd` and `DummySeedGenerator` has been [moved](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/api_def_DummySeedGenerator.pbtxt) to `random.experimental`. So you are only deleting some obsolete versions of them in here.

rnett:
@karllessard will the CI do op generation, or should I re-run it locally to check the fixes?

karllessard:
The CI will regenerate the op and I've triggered a full native build on that PR to make sure we don't break anything, I'll have to wait until it is finished before merging

karllessard (approved):

PR #195, opened 2021-01-26
Misc fixes and docs update

### LINKED ISSUE #168: Match test configuration for ndarray and framework to core
`ndarray` and `tensorflow-framework` both use version `2.22.2` of the surefire plugin, while `tensorflow-core` uses `2.22.0`.  Is there a reason for that or can I bump `core`'s version?

They also configure it with:
```xml
<skipTests>false</skipTests>
```
This is rather annoying because it makes `-DskipTests` not skip the tests, which is what IntelliJ (and I would assume most other IDEs) use for their "Skip Tests" button.  Is there a reason for this or can I remove it?

rnett:
Fixes #168 (makes `ndarray` and `framework` respect `-DskipTests`), gitignores bazel config files, and adds a `CONTRIBUTING.md` file as discussed in https://github.com/tensorflow/java/issues/176#issuecomment-766407236.  Also adds a link to the gitter channel.

If there's anything else you think should be added to `CONTRIBUTING.md` let me know, these are all of the tips that I'm aware of.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,57 +1,99 @@
-# Building and contributing to TensorFlow Java
+# Building and Contributing to TensorFlow Java
 
 ## Building
 
-To build all the artifacts, simply invoke the command `mvn install` at the root of this repository (or
-the Maven command of your choice). It is also possible to build artifacts with support for MKL enabled with
+To build all the artifacts, simply invoke the command `mvn install` at the root of this repository (or the Maven command of your choice). It is also
+possible to build artifacts with support for MKL enabled with
 `mvn install -Djavacpp.platform.extension=-mkl` or CUDA with `mvn install -Djavacpp.platform.extension=-gpu`
 or both with `mvn install -Djavacpp.platform.extension=-mkl-gpu`.
 
 When building this project for the first time in a given workspace, the script will attempt to download
-the [TensorFlow runtime library sources](https://github.com/tensorflow/tensorflow) and build of all the native code
-for your platform. This requires a valid environment for building TensorFlow, including the [bazel](https://bazel.build/)
+the [TensorFlow runtime library sources](https://github.com/tensorflow/tensorflow) and build of all the native code for your platform. This requires a
+valid environment for building TensorFlow, including the [bazel](https://bazel.build/)
 build tool and a few Python dependencies (please read [TensorFlow documentation](https://www.tensorflow.org/install/source)
 for more details).
 
-This step can take multiple hours on a regular laptop. It is possible though to skip completely the native build if you are
-working on a version that already has pre-compiled native artifacts for your platform [available on Sonatype OSS Nexus repository](#Snapshots).
-You just need to activate the `dev` profile in your Maven command to use those artifacts instead of building them from scratch
+This step can take multiple hours on a regular laptop. It is possible though to skip completely the native build if you are working on a version that
+already has pre-compiled native artifacts for your platform [available on Sonatype OSS Nexus repository](#Snapshots). You just need to activate
+the `dev` profile in your Maven command to use those artifacts instead of building them from scratch
 (e.g. `mvn install -Pdev`).
 
-Note that modifying any source files under `tensorflow-core` may impact the low-level TensorFlow bindings, in which case a
-complete build could be required to reflect the changes.
+Modifying the native op generation code (not the annotation processor) or the JavaCPP configuration (not the abstract Pointers) will require a
+complete build could be required to reflect the changes, otherwise `-Pdev` should be fine.
 
+### GPU Support
+
+Currently, due to build time constraints, the GPU binaries only support compute capacities 3.5 and 7.0.  
+To use with un-supported GPUs, change the value [here](tensorflow-core/tensorflow-core-api/build.sh#L27) and build the binaries yourself. While this
+is far from ideal, we are working on getting more build resources, and for now this is the best option.
+
+To build for GPU, pass `-Djavacpp.platform.extension=-gpu` to maven. By default, the CI options are used for the bazel build.  
+Using Tensorflow's configure script and copying the resulting `.tf_configure.bazelrc` to `tensorflow-core-api` can be used to override these options (
+like cuda locations). See the [Working with Bazel generation](#working-with-bazel-generation) section for details. If you do this, make sure
+the `TF_CUDA_COMPUTE_CAPABILITIES` value in your `.tf_configure.bazelrc` matches the value set in `build.sh`.
 
 ## Running Tests
 
-`ndarray` can be tested using the maven `test` target.  `tensorflow-core` and `tensorflow-framework`, however, 
-should be tested using the `integration-test` target, due to the need to include native binaries.
-It will **not** be ran when using the `test` target of parent projects, but will be ran by `install` or `integration-test`.
+`ndarray` can be tested using the maven `test` target.  `tensorflow-core` and `tensorflow-framework`, however, should be tested using
+the `integration-test` target, due to the need to include native binaries. It will **not** be ran when using the `test` target of parent projects, but
+will be ran by `install` or `integration-test`. If you see a `no jnitensorflow in java.library.path` error from tests it is likely because you're
+running the wrong test target.
+
+### Native Crashes
+
+Occasionally tests will fail with a message like:
+
+```
+Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test(default-test)on project tensorflow-core-api:There are test failures.
+
+    Please refer to C:\mpicbg\workspace\tensorflow\java\tensorflow-core\tensorflow-core-api\target\surefire-reports for the individual test results.
+    Please refer to dump files(if any exist)[date]-jvmRun[N].dump,[date].dumpstream and[date]-jvmRun[N].dumpstream.
+    The forked VM terminated without properly saying goodbye.VM crash or System.exit called?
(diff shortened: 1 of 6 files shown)


rnett:
To add when I'm not building:
```
Currently, do to build time constraints, the GPU binaries only support compute capacities 3.5 and 7.0.  To use with un-supported GPUs, change the value [here](tensorflow-core/tensorflow-core-api/build.sh#L27) build the binaries yourself.

To build for GPU, pass `-Djavacpp.platform.extension=-gpu` to maven.  Tensorflow's configure script can be used to set options (like cuda locations) for the native build.  Make sure the `TF_CUDA_COMPUTE_CAPABILITIES`  value in your `.tf_configure.bazelrc` matches the value set in `build.sh`.

```

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+for your platform. This requires a valid environment for building TensorFlow, including the [bazel](https://bazel.build/)
+build tool and a few Python dependencies (please read [TensorFlow documentation](https://www.tensorflow.org/install/source)
+for more details).
+
+This step can take multiple hours on a regular laptop. It is possible though to skip completely the native build if you are
+working on a version that already has pre-compiled native artifacts for your platform [available on Sonatype OSS Nexus repository](#Snapshots).
+You just need to activate the `dev` profile in your Maven command to use those artifacts instead of building them from scratch
+(e.g. `mvn install -Pdev`).
+
+Note that modifying any source files under `tensorflow-core` may impact the low-level TensorFlow bindings, in which case a
```
karllessard, line 21: Actually, I think a full native build is only required if we modify the op generator sources. It should be safe to reuse the pre-compiled native artifacts when modifying any of the Java source files

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+### Formatting
+
+Java sources should be formatted according to the [Google style guide](https://google.github.io/styleguide/javaguide.html).
+It can be included in [IntelliJ](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) and 
+[Eclipse](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml).
+[Google's C++ style guide](https://google.github.io/styleguide/cppguide.html) should also be used for C++ code.
+
+### Code generation
+
+Code generation for `Ops` and related classes is done during `tensorflow-core-api`'s `install`, using the annotation processor in
```
karllessard, line 42: It happens in the `compile` phase, not `install`

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+It can be included in [IntelliJ](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) and 
+[Eclipse](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml).
+[Google's C++ style guide](https://google.github.io/styleguide/cppguide.html) should also be used for C++ code.
+
+### Code generation
+
+Code generation for `Ops` and related classes is done during `tensorflow-core-api`'s `install`, using the annotation processor in 
+`tensorflow-core-generator`. If you change or add any operator classes (annotated with `org.tensorflow.op.annotation.Operator`), 
+endpoint methods (annotated with `org.tensorflow.op.annotation.Endpoint`), or change the annotation processor, be sure to re-run a 
+full `mvn install` in `tensorflow-core-api`.
```
karllessard, line 45: Are you sure about that? I don't recall exactly but the annotation processor will run even with the `dev` profile

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Working with Bazel generation
+
+`tensorflow-core-api` uses Bazel-built C++ code generation to generate most of the `@Operator` classes.  To get it to build, you will likely need to 
+clone the [tensorflow](https://github.com/tensorflow/tensorflow) project, run its configuration script (`./configure`), and copy the resulting 
+`.tf_configure.bazelrc` to `tensorflow-core-api`.
```
karllessard, line 51: Why this extra step? I never had to do it before. 

We have already a [`bazelrc`](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/.bazelrc) configuration file saved in our repo, I've noticed that it has not been updated after each TF version upgrade like it should, maybe that's the issue?

karllessard (changes requested):
Thank you @rnett , that's valuable documentation that was definitely missing

CONTRIBUTING.md
rnett: Agreed, that was in the readme so I didn't want to change it.  I'll update it.

CONTRIBUTING.md
rnett: Yeah, this isn't (meant to be) related to `dev`, it's just a "make sure you re-run codegen if you add ops" warning.

CONTRIBUTING.md
rnett: Maybe?  But if you have to do something like enable CUDA, or point it to non-default CUDA dirs, or use a different Python binary (that was my issue, I needed to point it to my conda install), you need the extra configuration.  It's handled in `build.sh` for maven builds, but if you aren't going through build.sh I had issues.  I had continual issues with needing `--incompatible_restrict_string_escapes=false` as well.

rnett: It was also necessary to get my bazel IDE plugin working, there were times when `bazel build` would work but the plugin's command wouldn't.  I can move it to more of a "if your build doesn't work try this" section.

karllessard: Yes, or something like: "The bazel build is configured for TF Java [continuous integration](link to CI workflow). If your current local setup is different, you might need to reconfigure bazel as follow:..."

rnett:
I changed `build.sh` to use a set `TF_CUDA_COMPUTE_CAPABILITIES` by default as per #200, @karllessard can you add the CI build tag just in case?

karllessard:
@rnett , the GPU build failed with that change: 
```
Cuda Configuration Error: Invalid compute capability: compute_'35
INFO: Elapsed time: 15.494s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```
https://github.com/tensorflow/java/runs/1826080574?check_suite_focus=true

rnett:
It's the single quotes, I think.  Should be fixed.

karllessard:
@rnett , failed again for a very awkward reason this time. I can simply retrigger a build, sometimes things get wrong on GitHub Action hosted servers but I'll you validate first:
```
Analyzing: 6 targets (180 packages loaded, 9563 targets configured)
Analyzing: 6 targets (199 packages loaded, 14836 targets configured)
WARNING: Download from https://storage.googleapis.com/mirror.tensorflow.org/github.com/llvm/llvm-project/archive/7e825abd5704ce28b166f9463d4bd304348fd2a9.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
INFO: Analyzed 6 targets (204 packages loaded, 18538 targets configured).
INFO: Found 6 targets...
[0 / 1,314] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[2,727 / 6,848] Creating runfiles tree bazel-out/k8-opt-exec-50AE0418/bin/external/org_tensorflow/tensorflow/cc/ops/array_ops_gen_cc.runfiles; 0s local ... (2 actions, 1 running)
ERROR: /root/.cache/bazel/_bazel_root/47b4c4ffab8583ddc67d4f1b3159b1ed/external/nccl_archive/BUILD.bazel:53:1: C++ compilation of rule '@nccl_archive//:device_lib' failed (Exit 1)
/opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/../../../../include/c++/7/limits: In static member function 'static constexpr long double std::numeric_limits<long double>::denorm_min()':
/opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/../../../../include/c++/7/limits:1799:65: error: expected ')' before numeric constant
       denorm_min() _GLIBCXX_USE_NOEXCEPT { return __LDBL_DENORM_MIN__; }
                                                                 ^~~
/opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/../../../../include/c++/7/limits:1799:68: error: cannot convert 'double (*)(const char*) throw ()' to 'long double' in return
       denorm_min() _GLIBCXX_USE_NOEXCEPT { return __LDBL_DENORM_MIN__; }
```
https://github.com/tensorflow/java/runs/1827956486?check_suite_focus=true

rnett:
Yeah that doesn't sound like anything I changed.

karllessard:
You are right, master is also [failing](https://github.com/tensorflow/java/actions/runs/535978816) now for the same reasons... (sigh)... @saudet , do you think it is a wrong GCC installation or something like that?

saudet:
@karllessard Tried upgrading to GCC 8 and that didn't change anything, so probably not an issue with GCC:
https://github.com/saudet/tensorflow-java/actions/runs/541859636

saudet:
@karllessard I'm not having any more luck with TensorFlow 2.4.1 and CUDA 11.0 either, so it's not something that used to be broken for sure. It actually seems like something recently broke. It's obviously downloading something somewhere for CUDA that isn't pinned to a specific version, and that got upgraded unintentionally somewhere upstream, which broke TF's build, at least on CentOS. I assume this is going to get fixed upstream, eventually...

karllessard:
Thanks @saudet , maybe we can go back to this container with Cuda we used to have before we start installing it manually? It stopped working as well but might have been fixed upstream since? Also that shortened the build time for Linux, which seems less stable these days.

saudet:
Using the container isn't going to change anything about CUDA. That's not
the issue here.

karllessard:
@rnett , were you able at least to build your version with GPU support on your machine? It might take awhile before this issue gets resolved by itself, I'll merge this if you confirm that it worked on your side.

rnett:
Yeah, although I had `TF_CUDA_COMPUTE_CAPABILITIES` set in my bazelrc so I'm not 100% it was tested.  I'll run one tonight or tomorrow without it and let you know.

rnett:
Ok, it works locally.

karllessard (approved):

PR #207, opened 2021-02-04
Better cross-environment error messages

rnett:
This PR exposes `env()` to `Operation` and `Op`, and adds nicer error messages when using inputs from other environments.

@karllessard it's currently failing the [`failWhenMixingOperationsOnDifferentGraphs` test](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/GraphOperationBuilderTest.java#L35-L35), but looking at it it looks like the test is wrong.  It fails when an exception is thrown, which given the name seems backwards.  Mixing graphs like this fails anyways, just not until the session is ran (it gives a very cryptic `Graph is invalid, contains a cycle` exception, but I'm assuming that's due to the mixing).

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
@@ -69,6 +69,11 @@ class EagerOperation extends AbstractOperation {
   }
 
   @Override
+  public EagerSession env() {
+    return session;
+  }
+
+  @Override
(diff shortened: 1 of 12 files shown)


karllessard:
@rnett , I agree the broken test seems reverted, it should fail if there are no exception thrown, that is strange. You can update it or simply remove it since you seem to cover these cases in another test suite now.

rnett:
It's fixed, should be good to merge.

karllessard (approved):
Thanks @rnett !

PR #209, opened 2021-02-05
Update LICENSE

theadactyl:
Updating license according to current guidance. (removing "copyright" bit)


karllessard (approved):

karllessard:
Thanks @theadactyl , so for guidance, should we stop including this copyright in all of our source files now?

theadactyl:
I believe so. The copyright language as far as I was told is outdated.

PR #210, opened 2021-02-05
Number constant ops

rnett:
Adds a `constant` ops that takes `Number` and a `TNumber` class, and one that does the same based off of an `Operand`'s runtime type.  The conversion is done using `Number.doubleValue()` etc rather than `cast`, since it avoids adding an op, and is necessary to call the other `constant` ops.  It's useful when you want to create a composite op that works with any numeric types, but need to add a constant, or for enabling things like the `+` operator in Kotlin to not need a new overload for each scalar type.


karllessard (approved):
Thanks @rnett !

Issue #205, opened 2021-02-03
Functional API: Execution environment agnostic function


rnett:
cc @karllessard 

This is somewhat of a sub-task of #181.  The biggest pain I've ran into when using ConcreteFunction is that it only has tensor `call` methods, when it's mostly going to be used with `Operand`.  This is a fairly simple issue on the surface.  But, there's no way to execute `ConcreteFunction` in graph mode, i.e. if they are nested.  The function used to generate the graph-mode outputs (which are wrapped in Signature) isn't saved.  Now, it's easy enough to do this Java-side, in a sub class so that `ConcreteFunction` still supports loading.  However, there's other issues such as supporting inputs with different shapes and dtypes that made me realize that what I'm trying to do here is closer to Python's [`Function`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/eager/def_function.py#L466-L466) and we may want to handle it with a new abstraction.  There's also `TF_Function` and `TF_GraphCopyFunction` and `TFE_ContextAddFunction` which seems like it would allow attaching a `ConcreteFunction` to a graph without having to re-execute the builder in a new graph.

So I'd propose two things:
 * Implement Graph-mode and Eager-mode use of `ConcreteFunction`s using the native `TF_Function` APIs (the fact that the eager one doesn't mention gradients makes me a little worried, but I would think we can handle that manually later if necessary).
 * Add a `Function` class that acts like `tf.function`, in that it creates `ConcreteFunction`s as necessary for the argument shapes and dtypes.  Additionally, since this will save the graph-creator lambda, we can have a `debug` flag that re-runs the lambda.

We also need to do something with variable handling, although that will probably need to wait on #179.  Python seems to use an implicit variable-creation context to create them at the call-site and only allows it on the first call.  I'd be fine with throwing errors and forcing the user to extract them, I think.  I need to look into the details a bit more before I propose anything for this though.  Variable scopes might be worth doing anyways for freezing, although hopefully explicit as part of `Ops`/`Scope`.

We'll need to pay attention to Graph states, like random seeds, too.

rnett:
More thoughts about `Function`:
 * Inputs: Python auto-wraps everything in tensors, we can't do this.  I'd like to allow non-tensor arguments, re-creating the graph if they change (treating them like shapes/dtypes).  This also allows for some control flow structures inside the method, since we don't have access to autograph.  I'd also like to include collections, and re-build if their structure changes (i.e. map keys, size of a list).
[...]

PR #211, opened 2021-02-06
Native function support

rnett:
Adds support for creating `TF_Function`s from graphs and executing them in eager and graph mode.

I'm undecided as to whether I should keep the positional `GraphFunction` or just make the base function use named arguments.  Thoughts @karllessard?

I'm also considering making the `create` methods automatically find the minimal subgraph for the passed inputs and outputs (and throw if that's impossible, as the function would break anyways).  This would help with gradient handling.  I'd add another overload to create everything (which could be used by ConcreteFunction), and possibly ones to detect inputs and outputs from the graph.

Before merging, I need to handle gradients for graph mode (it doesn't seem to need to be specified in eager mode, I need to check how Python does it), and add overloads with control outputs.


rnett:
Notes to self:
 * Look into partitioned_call, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/functional_ops.py#L1126-L1126
 * SymbolicGradient for eager mode function gradients: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/gradients_util.py#L308-L308
 * See also https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/eager/function.py#L907-L907
 * `TF_AddGradients` seems to call it.
 * proto: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt#L53057-L53057

Assuming that eager gradients is done by calling SymbolicGradient, I probably want an option to do the gradient calculation when the function is created (like for graph) and use that function in eager, rather than re-calculating it each time.  Depends on what gradient tape looks like.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+ * A Function created from a TensorFlow {@link Graph}. Callable in Graph or Eager mode via the {@link
+ * Ops#callFunction(GraphFunction, Operand[])} ops or the {@link #call(Ops, Operand[])} methods.
+ *
+ * <p>By default, resolves arguments and returns outputs by position.
+ * Names can be specified for the inputs and outputs (and used when calling) using {@link NamedGraphFunction} and {@link
+ * #create(Graph, String, boolean, Map, Map, String)}.
+ *
+ * Functions can also be created from {@link ConcreteFunction}s using {@link #create(ConcreteFunction, boolean)}.
+ */
+public class GraphFunction implements AutoCloseable {
```
karllessard, line 50: `GraphFunction` is a bit confusing as a name, it sounds like it is a function attached to a `Graph` which is not necessarily a case. I'd turn this over to `FunctionGraph`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -221,10 +222,8 @@ public Runner fetch(String operation) {
      * @return this session runner
      */
     public Runner fetch(String operation, int index) {
-      Operation op = operationByName(operation);
-      if (op != null) {
-        outputs.add(op.output(index));
-      }
+      Operation op = graph.operationOrError(operation);
```
karllessard, line 225: This will break current behavior by throwing an error where it was silently failing before... but I think it's for the best, so let's keep it that way

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/presets/tensorflow.java
```diff
@@ -184,15 +184,23 @@
...
+                new Info("TF_Tensor").pointerTypes("TF_Tensor").base("org.tensorflow.internal.c_api.AbstractTF_Tensor"))
+            .put(new Info("TF_Session").pointerTypes("TF_Session")
+                .base("org.tensorflow.internal.c_api.AbstractTF_Session"))
+            .put(new Info("TF_SessionOptions").pointerTypes("TF_SessionOptions")
+                .base("org.tensorflow.internal.c_api.AbstractTF_SessionOptions"))
+            .put(new Info("TF_Graph").pointerTypes("TF_Graph").base("org.tensorflow.internal.c_api.AbstractTF_Graph"))
+            .put(new Info("TF_Graph::graph").javaText("public native @MemberGetter @ByRef Graph graph();"))
+            .put(new Info("TF_Graph::refiner").javaText("public native @MemberGetter @ByRef ShapeRefiner refiner();"))
+            .put(new Info("TF_Function").pointerTypes("TF_Function")
+                .base("org.tensorflow.internal.c_api.AbstractTF_Function"))
```
karllessard, line 203: Are you sure about this reformatting? Something seems off a bit.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Function.java
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * Call {@code function}, adding it to the execution environment if it isn't already present.
+   *
+   * @param function the function to call
+   * @param inputs the inputs to the function
+   * @return the outputs of the function
+   */
+  @Endpoint(name = "callFunction")
+  public static List<Operand<?>> callFunction(Scope scope, GraphFunction function, List<Operand<?>> inputs) {
```
karllessard, line 49: Why not simply call these endpoints `call`? I like the simplicity of `tf.call(function, x).get("y")`

The function instance type is enough to differentiate the different signatures

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+  GraphFunction(TF_Function nativeHandle, int numInputs, int numOutputs) {
+    this.numInputs = numInputs;
+    this.numOutputs = numOutputs;
+    scope = new PointerScope();
+    this.nativeHandle = nativeHandle;
+    scope.attach(nativeHandle);
+  }
+
+  /**
+   * This causes native errors when I use it (Linux GPU, 6.1 CC), but I'm leaving it because how to enable XLA JIT is
```
karllessard, line 66: maybe change this javadoc to a FIXME?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    return (TF_Operation) handle;
+  }
+
+  private static TF_Output resolveToOutput(List<Operand<?>> operands) {
+    TF_Output handles = new TF_Output(operands.size());
+    for (int i = 0; i < operands.size(); i++) {
+      Operand<?> input = operands.get(i);
+      //TODO body.checkInput(input)
```
karllessard, line 108: can you explain more in detail here what is left TODO?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+    return handles;
+  }
+
+  /**
+   * Create a {@link GraphFunction} with positional argument and output resolution. Includes all ops in {@code body},
+   * {@code inputs} and {@code outputs} must be the inputs and outputs of {@code body}.
+   *
+   * @param body the graph to use as the function body
+   * @param name the name of the function
+   * @param appendHash whether to append the has of the signature to the name
```
karllessard, line 122: ".. the has*h*..."

btw, just curious, why do we let the user decide to append a hash or not to the signature name? Can we take this decision for him?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+
+  /**
+   * Create a {@link NamedGraphFunction} from a {@link ConcreteFunction}.
+   *
+   * @param function the {@code ConcreteFunction} to create the function from
+   * @param appendHash whether to append a hash of the signature to the name
+   */
+  public static NamedGraphFunction create(ConcreteFunction function, boolean appendHash) {
```
karllessard, line 220: I'm not sure to understand why do we need both `NamedGraphFunction` and `ConcreteFunction`, they seem to achieve pretty much the same goal, which is to provided named inputs/outputs on an encapsulated function, right?

In the Python paradigm, a `Function` is a method that wraps an executable graph (or eager code) while a `ConcreteFunction` is the realization of this function for specific type of input arguments (since Python is non-typed, the function is realized (or "concretized") when its invoke for the first time with a given set of input types.

In Java, as everything is typed, we might not need to make this distinction. So why not just calling it simply `Function` and merge both classes together? The only difference I see is that one runs a main `Graph` (using a session) while the other runs a secondary graph (using a ops). But there might be a way to abstract this to the user.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
```diff
@@ -0,0 +1,301 @@
...
+    scope.close();
+  }
+
+  /**
+   * Get the native handle of the function's gradient, so that it can be attached to a Graph.  Not implemented yet.
+   *
+   * TODO implement
+   */
+  TF_Function gradNativeHandle() {
+    return null;
```
karllessard, line 271: Any idea how to implement this?

karllessard (changes requested):

karllessard:
Hi @rnett , I'm really happy that you tackle this feature that was in my backlog for too long now, let's get it done. I lean to an approach that would abstract as much as possible all the implementation details to the user, whether the function is a graph, a subgraph or code executed eagerly. We'll see where it goes in the scope of this PR

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: I need to do some tests, but for Graph I'm thinking I'd call the existing gradient op on it, and then construct a function using the op list constructor.  Eager is up in the air until we get gradient tape support.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: I'd lean in favor of always appending it, but I could see a case where you want to create a function with a specific name, so I left it in.  It's definitely necessary most of the time to avoid name collisions for functions with different arguments (aka overloading support).

rnett:
Yeah, I very much agree, I didn't intend to put that in this PR, just the native support.  I've put my thoughts on it in #205.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: Pretty much, yeah, but `NamedGraphFunction` has to manage the `TF_Function` handle and doesn't require a signature.  `ConcreteFunction` also allows for Tensor execution.  As for function, see #205, this is generally modeling Python's `DefinedFunction`/`_EagerDefinedFunction`/defun, which is separate from the `Function` class.

I could look into merging them, but it would make it a bit harder to declare the `GraphFunction`.  Would you be in favor of getting rid of the positional `GraphFunction` altogether and merging `NamedGraphFunction` with `ConcreteFunction`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: It's from #207, I'd originally written this in that branch.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Function.java
rnett: I had issues with `call(NamedGraphFunction, List<>)`, in that I'd like to have it return a Map using the named outputs even if named inputs aren't used, but that collides with `call(GraphFunction, List<>)` which returns a List.  Let's delay this till after the merging w/ ConcreteFunction conversation?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/presets/tensorflow.java
rnett: Ah yeah, a lot of the JavaCPP preset files are done w/ 4 spaces (so are all the generated ones, but I don't think I can do anything about that).  I'll fix that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
rnett: It won't actually change things, I just copied the Session function to Graph.  The null check there was extraneous.  But I agree about the behavior, I'll note it in the docs.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: I agree, though lets have the "merge with ConcreteFunction" discussion first.  I didn't want to use `FunctionGraph` because Python has a `FuncGraph` that subclasses Graph, and I'm not sure if we'll end up needing one for further Function support (I'd prefer a subclass of Ops, which should handle most of the same stuff).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
rnett: I'm expecting it to get merged before this branch anyways, in which case I'll uncomment it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphFunction.java
karllessard: Yes, I'm for merging everything together if that can simplify the API for the user. Or we can continue to segregate the logic into multiple classes but only one should be public.

rnett: When trying to make `GraphFunction` (now called `DefinedFunction`) package local, I ran into the issue where I'd like to have a `Ops.call(ConcreteFunction, Map<String, Operand<?>)` function, but the necessary methods (and `DefinedFunction`) won't be accessible there.  There's a few options, as far as I can see:

1. Only have a call method on `ConcreteFunction`, like `ConcreteFunction.call(Ops tf, Map<String, Operand<?>)`.
2. Have a `@Operator` Function class in the main package
3. Have a public call method on `ConcreteFunction` that takes `Scope` instead of `Ops`, call that from the operator class
4. Move `ConcreteFunction`, `DefinedFunction`, the Operator class, and any future function classes to a new package (`function`, presumably).

4 is my preference, thoughts?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DefinedFunction.java
```diff
@@ -0,0 +1,212 @@
...
+ * A Function created from a TensorFlow {@link Graph}. Callable in Graph or Eager mode via the {@link
+ * Ops#callFunction(DefinedFunction, Operand[])} ops or the {@link #call(Ops, Operand[])} methods.
+ *
+ * <p>By default, resolves arguments and returns outputs by position.
+ * Names can be specified for the inputs and outputs (and used when calling) using {@link NamedGraphFunction} and {@link
+ * #create(Graph, String, boolean, Map, Map, String)}.
+ *
+ * Functions can also be created from {@link ConcreteFunction}s using {@link #create(ConcreteFunction, boolean)}.
+ */
+public class DefinedFunction implements AutoCloseable {
```
karllessard, line 47: I'm still not convinced that we need this extra class as part of the public API. Can't we have everything fronted by `ConcreteFunction`, eventually changing it to an interface that can have multiple implementations?

rnett: I agree, but I'm running to access modifiers:
> When trying to make `DefinedFunction` package local, I ran into the issue where I'd like to have a `Ops.call(ConcreteFunction, Map<String, Operand<?>)` function, but the necessary methods (and `DefinedFunction`) won't be accessible there. There's a few options, as far as I can see:
> 
> 1. Only have a call method on `ConcreteFunction`, like `ConcreteFunction.call(Ops tf, Map<String, Operand<?>)`.
> 2. Have a `@Operator` Function class in the main package
> 3. Have a public call method on `ConcreteFunction` that takes `Scope` instead of `Ops`, call that from the operator class
> 4. Move `ConcreteFunction`, `DefinedFunction`, the Operator class, and any future function classes to a new package (`function`, presumably).
> 

4 is my preference, thoughts?

karllessard: So like we discussed, let's add endpoints to `ConcreteFunction` that accepts a `Scope` object and do the actual job. Anyway, most users will always go through the ops so it's ok to have these kind of endpoints in the public API of the core classes even if they are more meant for internal usage.

We also agreed that only `ConcreteFunction` should remain and probably be converted to an interface that has 2 different possible implementations: one with a graph (like the actual one) and one with a function graph (the new one). Maybe add these implementations in the `org.tensorflow.internal` package?

And last, can we check if it will be possible to build `ConcreteFunction` instances out of function graphs found in a saved model that has been loaded?

rnett: I've started on this, and as far as I can tell, the only thing `ConcreteFunction` provides that the native functions doesn't is the ability to call with `Tensor`s, and even that's relocatable by using an temporary EagerSession.  So do we even need it?  The `Signature` builders can be ported just fine.  The only thing I could think of that would cause issues is Saving, which I haven't looked at enough to be able to tell.  If I understand it right it puts everything under a default named function anyways, so we could just create a graph and add the function to it.

karllessard: | So do we even need it? 

The goal is that we don't! Maybe I'll preserve `ConcreteFunction` as the name for the common function interface since it is still accurate without making breaking changes again. But if you think it is simple to do everything within a single class instead of one interface and two implementations, you can give it a try and I'll take a look at the PR but I have slightly a preference for the latter.

rnett: Will do.  I'll start it without saving, and figure that out after.  I'm also not sure how variables are handled within `TF_Function`, which may cause some differences from `ConcreteFunction` (especially wrt saving), but I suspect that resource variables will be handled just fine, it's only the legacy ones that will have issues, so imo we don't need to worry about it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -39,16 +57,87 @@
  */
 public class ConcreteFunction implements AutoCloseable {
 
+  private static TF_Operation outputHandle(Operand<?> operand) {
```
karllessard, line 60: Nit: can we move the private parts at the bottom of the class? It's the Google way to do it and I kind of like it now :)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -154,141 +240,235 @@ public Signature signature() {
...
+   * TODO implement
+   */
+  TF_Function gradNativeHandle() {
+    return null;
   }
 
-  private final Graph graph;
-  private final Session session;
   private final Signature signature;
-  private final Ownership ownership;
```
karllessard, line 286: Why did you remove the ownership part? It is quite important to prevent unexpected behavior when closing a function that comes from a saved model vs. one that has been instantiated manually.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -154,141 +240,235 @@ public Signature signature() {
...
+  }
 
-    final SignatureDef signatureDef = signature.asSignatureDef();
-    final Session.Runner runner = session.runner();
 
-    signatureDef.getInputsMap().forEach((argName, t) -> {
-      Tensor tensor = arguments.get(argName);
-      if (tensor == null) {
-        throw new IllegalArgumentException(String.format("Missing argument [%s]", argName));
+  public Map<String, Operand<?>> call(Scope scope,
```
karllessard, line 252: javadoc?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -163,11 +163,12 @@ public Exporter withFunction(ConcreteFunction function) {
...
-        session = function.session();
-      } else if (session != function.session()) {
-        throw new UnsupportedOperationException("Saving multiple functions with different graphs/sessions is not supported yet.");
-      }
+      //TODO fix saving
+//      if (session == null) {
+//        session = function.session();
+//      } else if (session != function.session()) {
+//        throw new UnsupportedOperationException("Saving multiple functions with different graphs/sessions is not supported yet.");
+//      }
```
karllessard, line 171: We can't merge this PR until we support them in saved models properly

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Function.java
```diff
@@ -0,0 +1,45 @@
...
+import org.tensorflow.op.Scope;
+import org.tensorflow.op.annotation.Endpoint;
+import org.tensorflow.op.annotation.Operator;
+
+/**
+ * Ops for calling {@link ConcreteFunction}.  Even though the C API docs say the name of the Op needs to be the name of
+ * the function, they mean the type.
+ */
+@Operator(name = "call")
+public abstract class Function {
```
karllessard, line 31: Can we rename this class to `FunctionHelper` or something like that, to prevent clashing with `java.util.Function`?

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/ConcreteFunctionTest.java
```diff
@@ -83,27 +82,27 @@ public void chainFunctions() {
...
-    }
+//    try (Graph g = new Graph()) {
+//      Signature signature = plusFive(Ops.create(g));
+//      Session s;
+//      try (ConcreteFunction f = ConcreteFunction.create(signature, g)) {
+//        s = f.session();
+//      }
+//      assertThrows(IllegalStateException.class, () -> s.run(Init.DEFAULT_NAME));
+//      g.toGraphDef();  // check that graph is still valid
+//    }
```
karllessard, line 105: Need to fix these tests before merging

karllessard (changes requested):
Thanks @rnett , I like where this is going. I'm concerned a bit about the support in the saved model though, for which reason the resource ownership of a `ConcreteFunction` was added for (and removed in this PR). I can give you more context about this if you want.

Also, I have a feeling that the size of this PR has increased dramatically (did it?), if that the case, is everything in there required to add this new functionality for native functions or there is a mix of other things as well? Reducing the size of the PRs helps a lot with the review process. Thanks!

rnett:
> Thanks @rnett , I like where this is going. I'm concerned a bit about the support in the saved model though, for which reason the resource ownership of a `ConcreteFunction` was added for (and removed in this PR). I can give you more context about this if you want.
> 
> Also, I have a feeling that the size of this PR has increased dramatically (did it?), if that the case, is everything in there required to add this new functionality for native functions or there is a mix of other things as well? Reducing the size of the PRs helps a lot with the review process. Thanks!

I didn't think you'd get to reviewing it yet, parts are still WIP (tests and saving).  I'll mark it as draft until they are done.

The size hasn't increased much, I think it's mostly bad diffs and perhaps some rebase issues, plus auto-formatting files that weren't previously (Graph mostly, it looks like).  I can pull some parts out into an earlier PR.  The only large part I added recently was tree-walking functions on `GraphOperation`, which is necessary for building the new functions (since they may be a subset of a graph).

Context on saving would be appreciated, I'm going to take a look at how python does it.  If I understand it right, both the function native instances and the resource variables are not associated with any specific session or graph (although I need to test that for variables), so I can use a graph and session for loading and then discard them.  We're going to need some kind of variable lifting anyways for `@tf.function` style stuff, so if the variables are session local I'd like to design and use something like that here, rather than duplicating it.  Worst case scenario I can add a session back to `ConcreteFunction` with ownership, but I don't understand how that works with attaching the function to the graph (need to test that too).  I.e. if I make a function and attach it to two different graphs, do they share variables.  That's also probably different for resource and non-resource variables.

How much do we need to support non-resource variables, anyways?  It seems like resource variables were introduced largely to make the functional stuff easier.  Of course, we don't have access to them yet, but we should soon-ish (my gradient PR is sitting in their review queue).

rnett:
Replaced by #233.

PR #212, opened 2021-02-10
Upgrade for TensorFlow 2.4.1

saudet:
This draft PR includes initial changes required just to get the builds going with TF 2.4.1. Some tests are crashing, probably because we still need to update the handling of `TF_String` as per the breaking change below, but there's probably a bit more to be done.
>  * The byte layout for string tensors across the C-API has been updated to match TF Core/C++; i.e., a contiguous array of `tensorflow::tstring`/`TF_TString`s.

https://github.com/tensorflow/tensorflow/releases/tag/v2.4.0

It's quite a big release, and a lot of ops have been added, so even after getting everything fixed up, there's probably more to be done before we can get this merged. @karllessard Please take a look and feel free to push any changes to my fork and/or merge this into a branch of the main repository.

On the flip side, most toolchain workarounds have been moved from `configure.py` to `.bazelrc` (so we can clean up `build.sh` a little), we finally have support for CUDA 11 and cuDNN 8, MKL-DNN 1.x now works on all platforms (Linux, Mac, and Windows) without huge external binaries, and I was able to find what the issue was with CUDA on CentOS. Apparently, it doesn't like its new version of glibc, and we can prevent it from getting installed by calling `yum` this way:
```bash
          GLIBC="glibc glibc-common glibc-devel glibc-headers"
          yum --disablerepo updates -y install $GLIBC
          yum -x "$GLIBC" -y update ...
```
https://github.com/saudet/tensorflow-java/blob/upgrade-tensorflow-241/.github/workflows/ci.yml#L53-L58
@karllessard Feel free to merge this bit right away for TF 2.3.1!


karllessard:
Thanks a lot @saudet , let's prioritize this merge and let 2.3.1 broken for now. I'll pull your branch and see what I can do to fix the string issues + classify the new ops

karllessard:
Finally, while I'm building your branch, let's give it a try on 2.3.1... https://github.com/tensorflow/java/commit/8b36d7e89c541e1e9d9823df0359fd5d39d393f6

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/global/tensorflow.java
```diff
@@ -11,6 +11,264 @@
...
+
+// #ifndef TENSORFLOW_CORE_PLATFORM_CTSTRING_H_
+// #define TENSORFLOW_CORE_PLATFORM_CTSTRING_H_
+
+// #include <stdint.h>
+// #include <stdlib.h>
+
+// #include "tensorflow/core/platform/ctstring_internal.h"
+
+// Initialize a new tstring.  This must be called before using any function
```
karllessard, line 190: Any clue why the documentation of these new string endpoints get all messed up by being aggregated together?

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_generator.cc
```diff
@@ -349,6 +349,10 @@ void RenderGettersAndSetters(const OpSpec& op, SourceWriter* writer) {
         .EndMethod();
   }
   for (const ArgumentSpec& output : op.outputs()) {
+    if ("size" == output.var().name()) {
+        // reserved name
+        continue;
+    }
```
karllessard, line 355: We should not skip the getter of an output that have a reserved name but just rename it in the API def, though I'm not sure why `size` would cause any problem here. Can you give more context to this change?

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/global/tensorflow.java
saudet: The TF_TString type isn't defined in ctstring.h, so we need to include ctstring_internal.h as well, which contains the definitions of those functions too. JavaCPP simply skips the redundant declarations that follow. We can hack something up, but this is something they should fix upstream, and eventually will I assume, so probably not something worth spending time on here.

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_generator.cc
saudet: There are new ops with outputs named "size". Do whatever you want with them, I just did that to get them to compile.

saudet:
Everything builds and all tests pass, so I believe this is ready to merge, unless we should refine all changes to the ops here as well? Also, the GPU builds on Linux are timing out:
https://github.com/saudet/tensorflow-java/actions/runs/585339990
For the past few days, the normal CPU builds have been taking 4 hours instead of 3, and that would explain why the GPU builds don't finish in 6 hours. Either something's changed with the VMs, but more likely this will pass.

karllessard:
I'm actually working on classifying the new ops, which should be completed before merging. I'll make a PR to your branch probably today. 

I'm almost done, I just hit a backward-compatibility issue with some of them which I'll duplicate for now but that's something I'd like to bring as a topic for our next session to decide how to handle this properly next time.

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_generator.cc
karllessard: Ok, I'll take care of it

karllessard:
Just pushed the changes, let's see how it goes with the build: https://github.com/tensorflow/java/actions/runs/589500056

karllessard:
@googlebot I consent

PR #214, opened 2021-02-11
Boolean mask ops

rnett:
Adds custom ops for boolean masking (`tensor[mask]` and `tensor[mask] = update`).

Note that there is no Python equivalent of `BooleanMaskUpdate`, but there is in Numpy and PyTorch, so I included it.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
@@ -126,6 +126,25 @@ public final class Scope {
   }
 
   /**
+   * Returns a new scope where added operations will be prefixed by this scope's op name
+   * (set by {@link #withName(String)}), or the given default if it is unset.  This is intended to be used for
+   * composite ops.
+   *
+   * <p>Ops created with this scope will have {@code name/opName/} as the prefix. The actual
+   * name will be unique in the returned scope. All other properties are inherited from the current
+   * scope.
+   *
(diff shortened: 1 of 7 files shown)


rnett:
I can't run tests locally b/c `jnitensorflow in java.library.path` errors, I'm guessing because snapshot builds were just fixed.  We'll see if the CI runs them.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -125,6 +125,25 @@ public Scope withName(String opName) {
...
+   * <p>The default child scope name must match the regular expression {@code [A-Za-z0-9.][A-Za-z0-9_.\-]*}
+   *
+   * @param defaultName name of the sub scope if this scope's name hasn't been set.
+   * @return a new subscope
+   * @throws IllegalArgumentException if the name is invalid
+   */
+  public Scope withNameAsSubScope(String defaultName){
+    return new Scope(env, nameScope.withSubScope(nameScope.makeOpName(defaultName)), controlDependencies, deviceSpec);
+  }
+
```
karllessard, line 146: It could be interesting to update our other composite ops to make use of this new method, like this one: https://github.com/tensorflow/java/blob/ef6ece686a36809a186abd7db32b0a919829b48f/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Ones.java#L53

karllessard (approved):
Thanks @rnett !

PR #225, opened 2021-02-24
Adds `Session.restore` to allow loading of checkpoints saved by `Session.save`

Craigacp:
This allows the easy restoration of checkpoints during training. A fresh graph object loaded from a `GraphDef` will not have a `SaverDef` defined, so the method that creates a `SaverDef` was modified to check if the graph already had restore operations, in which case the names of the operations are filled in rather than trying to mutate the graph twice (which will throw an exception). The names are set based on the defaults used in Saver.py, though it's not identical as that adds an extra placeholder called "save/Const" for unspecified backwards compatibility, and when I mirror that in Java calling session.run gives TFIllegalArgumentException complaining that placeholder is unfilled.

It's not quite the same as the TF 2 checkpoint restore mechanism, but that seems tightly woven into the Python structure and we don't quite have the infrastructure yet.

I also bumped the version of `maven-surefire-plugin` as otherwise I get error messages out of Maven when running the tests. Not sure when this started, but it's happening for me on Ubuntu 20.04 x86_64.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/pom.xml
+++ b/tensorflow-core/tensorflow-core-api/pom.xml
@@ -373,7 +373,7 @@
       </plugin>
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.22.0</version>
+        <version>3.0.0-M5</version>
         <executions>
           <execution>
             <!--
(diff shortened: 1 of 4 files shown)


karllessard (approved):
Thanks @Craigacp !

PR #227, opened 2021-02-26
Add dependency policy to CONTRIBUTING.md

rnett:
Add dependency policy to CONTRIBUTING.md, noting the Google-provided list and that we want to keep dependencies minimal.


karllessard (approved):
Thanks @rnett !

PR #232, opened 2021-03-04
Add graph walking functions to Graph and GraphOperation

rnett:
Extracted from #211, adds graph walking and a minimal subgraph function.

Also fixes a bug in Constant where the scope wasn't applied (including device).

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
@@ -228,11 +228,13 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
    *
    * @param inputs the inputs of the subgraph.  Must be from single output ops.
    * @param outputs the outputs of the subgraph
+   * @param allowConstants whether to allow constants in the function body (that aren't in inputs)
    * @return the set of operations needed to calculate outputs from inputs, including outputs and inputs
    * @throws IllegalStateException if outputs depends on ops outside of the subgraph (i.e. is not calculable based
    * solely on inputs)
    */
-  public synchronized Set<GraphOperation> completeSubgraph(Set<Operand<?>> inputs, Set<Operand<?>> outputs) {
+  public synchronized Set<GraphOperation> completeSubgraph(Set<Operand<?>> inputs, Set<Operand<?>> outputs,
+          boolean allowConstants) {
     Queue<GraphOperation> currents = new LinkedList<>();
     Set<GraphOperation> seen = new LinkedHashSet<>(outputs.size());
     Set<GraphOperation> inputOps = new LinkedHashSet<>(inputs.size());
@@ -260,9 +262,11 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
       }
 
       if (op.numControlInputs() + op.numInputs() == 0) {
-        throw new IllegalStateException("Operation " + op
-                + " has no inputs, but is not set as an input.  "
-                + "It is impossible to calculate the specified outputs with the given inputs.");
+        if (!(allowConstants && op.type().equals(Constant.OP_TYPE))) {
+          throw new IllegalStateException("Operation " + op
+                  + " has no inputs, but is not set as an input.  "
+                  + "It is impossible to calculate the specified outputs with the given inputs.");
+        }
       }
 
       for (GraphOperation control : op.controlInputs()) {
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
@@ -31,7 +31,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
+import java.util.Set;
 import org.bytedeco.javacpp.BytePointer;
 import org.bytedeco.javacpp.Pointer;
 import org.bytedeco.javacpp.PointerScope;
@@ -69,12 +73,16 @@ import org.tensorflow.types.family.TType;
  */
 public final class Graph implements ExecutionEnvironment, AutoCloseable {
 
-  /** Create an empty Graph. */
+  /**
+   * Create an empty Graph.
+   */
   public Graph() {
     nativeHandle = allocate();
   }
 
-  /** Create a Graph from an existing handle (takes ownership). */
+  /**
+   * Create a Graph from an existing handle (takes ownership).
+   */
   Graph(TF_Graph nativeHandle) {
     this.nativeHandle = nativeHandle;
   }
@@ -129,6 +137,77 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
   }
 
   /**
+   * Returns the operation (node in the Graph) with the provided name, or throws {@link IllegalArgumentException} if
+   * there isn't one.
+   *
+   * @param name name of the operation to look for
+   * @return operation in the graph with this name
+   * @throws IllegalArgumentException if no such operation exists in the Graph
+   */
+  public GraphOperation operationOrError(String name) {
+    GraphOperation op = operation(name);
+    if (op == null) {
+      throw new IllegalArgumentException("No Operation named [" + name + "] in the Graph");
+    }
+    return op;
+  }
+
+  /**
+   * Returns the output with the provided name, or {@code null} if there is no such output.
+   * <p>Names should be of the
+   * format {@code /scope/op}, with an optional index: {@code /scope/op:1}. {@code 0} is used if the index is not
+   * specified.
+   *
+   * @param output the output to get
+   * @return the output with this name, or null if there isn't one
+   */
+  @SuppressWarnings("rawtypes")
+  public Output<?> output(String output) {
+    int colon = output.lastIndexOf(':');
+    if (colon == -1 || colon == output.length() - 1) {
+      GraphOperation op = operation(output);
+      if (op == null) {
+        return null;
+      }
+      return new Output(op, 0);
+    }
+    try {
+      String op = output.substring(0, colon);
+      int index = Integer.parseInt(output.substring(colon + 1));
+      GraphOperation operation = operation(op);
+      if (operation == null) {
+        return null;
+      }
+      return new Output(operation, index);
+    } catch (NumberFormatException e) {
+      GraphOperation op = operation(output);
+      if (op == null) {
+        return null;
+      }
+      return new Output(op, 0);
+    }
+  }
+
+  /**
+   * Returns the output with the provided name, or throws {@link IllegalArgumentException} if there isn't one.
+   * <p>Names should be of the
+   * format {@code /scope/op}, with an optional index: {@code /scope/op:1}. {@code 0} is used if the index is not
+   * specified.
+   *
+   * @param output the output to get
+   * @return the output with this name
+   * @throws IllegalArgumentException if no such output exists in the Graph
+   * @see #output(String)
+   */
+  public Output<?> outputOrError(String output) {
+    Output<?> op = output(output);
+    if (op == null) {
+      throw new IllegalArgumentException("No Operation named [" + output + "] in the Graph");
+    }
+    return op;
+  }
+
+  /**
    * Iterator over all the {@link Operation}s in the graph.
    *
    * <p>The order of iteration is unspecified. Consumers of the iterator will receive no
@@ -138,14 +217,80 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
     return new OperationIterator(this);
   }
 
+  private GraphOperation graphOp(Operand<?> operand) {
+    checkInput(operand);
+    return (GraphOperation) operand.op();
+  }
+
+  /**
+   * Finds the operations used to produce {@code outputs} from {@code inputs}, or throws if that is not possible.
+   * Respects control dependencies.
+   *
+   * @param inputs the inputs of the subgraph.  Must be from single output ops.
+   * @param outputs the outputs of the subgraph
+   * @return the set of operations needed to calculate outputs from inputs, including outputs and inputs
+   * @throws IllegalStateException if outputs depends on ops outside of the subgraph (i.e. is not calculable based
+   * solely on inputs)
+   */
+  public synchronized Set<GraphOperation> completeSubgraph(Set<Operand<?>> inputs, Set<Operand<?>> outputs) {
+    Queue<GraphOperation> currents = new LinkedList<>();
+    Set<GraphOperation> seen = new LinkedHashSet<>(outputs.size());
+    Set<GraphOperation> inputOps = new LinkedHashSet<>(inputs.size());
+
+    for (Operand<?> input : inputs) {
+      if (input.op().numOutputs() > 1) {
+        throw new IllegalStateException("Only ops with one output are supported as subgraph inputs");
+      }
+      GraphOperation op = graphOp(input);
+      inputOps.add(op);
+      seen.add(op);
+    }
+
+    for (Operand<?> operand : outputs) {
+      GraphOperation op = graphOp(operand);
+      currents.add(op);
+    }
+
+    while (!currents.isEmpty()) {
+      GraphOperation op = currents.poll();
+
+      // skip if already present
+      if (!seen.add(op)) {
+        continue;
+      }
+
+      if (op.numControlInputs() + op.numInputs() == 0) {
+        throw new IllegalStateException("Operation " + op
+                + " has no inputs, but is not set as an input.  "
+                + "It is impossible to calculate the specified outputs with the given inputs.");
+      }
+
+      for (GraphOperation control : op.controlInputs()) {
+        if (!inputOps.contains(control)) {
+          currents.add(control);
+        }
+      }
+
+      for (Operand<?> input : op.inputs()) {
+        GraphOperation inputOp = graphOp(input);
+        if (!inputOps.contains(inputOp)) {
+          currents.add(inputOp);
+        }
+      }
+
+    }
+
+    return seen;
+  }
+
   /**
    * Returns a builder to add {@link Operation}s to the Graph.
    *
    * @param type of the Operation (i.e., identifies the computation to be performed)
    * @param name to refer to the created Operation in the graph.
    * @return an {@link OperationBuilder}, which will add the Operation to the graph when {@link
-   *     OperationBuilder#build()} is invoked. If {@link OperationBuilder#build()} is not invoked,
-   *     then some resources may leak.
+   * OperationBuilder#build()} is invoked. If {@link OperationBuilder#build()} is not invoked, then some resources may
+   * leak.
    */
   @Override
   public GraphOperationBuilder opBuilder(String type, String name) {
@@ -216,6 +361,7 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
 
   /**
    * Adds an initializer to the graph initializer list.
+   *
    * @param initializer An initializer to add to the list.
    */
   public synchronized void addInitializer(Op initializer) {
@@ -230,12 +376,11 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
   }
 
   /**
-   * Adds operations to compute the partial derivatives of sum of {@code y}s w.r.t {@code x}s, i.e.,
-   * {@code d(y_1 + y_2 + ...)/dx_1, d(y_1 + y_2 + ...)/dx_2...}
+   * Adds operations to compute the partial derivatives of sum of {@code y}s w.r.t {@code x}s, i.e., {@code d(y_1 + y_2
+   * + ...)/dx_1, d(y_1 + y_2 + ...)/dx_2...}
    *
    * <p>{@code dx} are used as initial gradients (which represent the symbolic partial derivatives
-   * of some loss function {@code L} w.r.t. {@code y}). {@code dx} must be null or have size of
-   * {@code y}.
+   * of some loss function {@code L} w.r.t. {@code y}). {@code dx} must be null or have size of {@code y}.
    *
    * <p>If {@code dx} is null, the implementation will use dx of {@link
    * org.tensorflow.op.core.OnesLike OnesLike} for all shapes in {@code y}.
@@ -245,8 +390,8 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
    *
    * <p>If {@code prefix} is null, then one will be chosen automatically.
    *
-   * @param prefix unique string prefix applied before the names of nodes added to the graph to
-   *     compute gradients. If null, a default one will be chosen.
+   * @param prefix unique string prefix applied before the names of nodes added to the graph to compute gradients. If
+   * null, a default one will be chosen.
    * @param y output of the function to derive
    * @param x inputs of the function for which partial derivatives are computed
    * @param dx if not null, the partial derivatives of some loss function {@code L} w.r.t. {@code y}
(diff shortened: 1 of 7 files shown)


rnett:
I may need to add a flag to include variables, too.

rnett:
The CI seems to be broken again, I'm not having issues running `install` for the whole project locally, but there's javadoc resolution errors from framework.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -138,14 +217,101 @@ public GraphOperation operation(String name) {
...
+   * @param outputs the outputs of the subgraph
+   * @param allowedBodyOps types of ops to allow as 0-input ops in the body.  Allows all (except {@code
+   * forbiddenBodyOps}) if null.
+   * @param forbiddenBodyOps types of ops to never allow as 0-input ops in the body.  Forbids all (except {@code
+   * allowedBodyOps}) if null.
+   * @return the set of operations needed to calculate outputs from inputs, including outputs and inputs
+   * @throws IllegalStateException if outputs depends on ops outside of the subgraph (i.e. is not calculable based
+   * solely on inputs)
+   */
+  public synchronized Set<GraphOperation> completeSubgraph(Set<Operand<?>> inputs, Set<Operand<?>> outputs,
```
Craigacp, line 241: Should we have an overload for this that just accepts the first two arguments and calls this with `Collections.emptySet()` for those arguments? Or document that all the arguments must be non-null and expect our users to suppled `Collections.emptySet()` on their own?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -138,14 +217,101 @@ public GraphOperation operation(String name) {
...
+    if (forbiddenBodyOps != null && allowedBodyOps != null) {
+      for (String t : forbiddenBodyOps) {
+        if (allowedBodyOps.contains(t)) {
+          throw new IllegalArgumentException("Can't allow and forbid op type " + t + ".");
+        }
+      }
+    }
+
+    Queue<GraphOperation> currents = new LinkedList<>();
+    Set<GraphOperation> seen = new LinkedHashSet<>(outputs.size());
```
Craigacp, line 253: `seen` is allocated as `outputs.size()`, but we immediately add each element of `inputs` to it. Should it be sized as `inputs.size()` (or even `inputs.size() + outputs.size()`)?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -138,14 +217,101 @@ public GraphOperation operation(String name) {
...
+    while (!currents.isEmpty()) {
+      GraphOperation op = currents.poll();
+
+      // skip if already present
+      if (!seen.add(op)) {
+        continue;
+      }
+
+      if (op.numControlInputs() + op.numInputs() == 0) {
+        if ((forbiddenBodyOps != null && forbiddenBodyOps.contains(op.type()))
```
Craigacp, line 279: Initialising these to `Collections.emptySet()` internally even if we don't make users do it will make this if a lot less ugly.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -170,6 +184,121 @@ Tensor tensor(int outputIdx) {
...
+
+      TF_OperationAllInputs(getUnsafeNativeHandle(), handles, numInputs);
+
+      List<Operand<?>> operands = new ArrayList<>(numInputs);
+      for (int i = 0; i < numInputs; i++) {
+        TF_Output atPos = handles.position(i);
+        TF_Operation op = atPos.oper();
+        int index = atPos.index();
+        String opName = TF_OperationName(op).getString();
+        operands.add(graph.operationOrError(opName).output(index));
```
Craigacp, line 207: In what situations could this throw?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -170,6 +184,121 @@ Tensor tensor(int outputIdx) {
     throw new IllegalStateException("Graph tensors must be fetched by running a session");
   }
 
+
+  public int numInputs() {
```
Craigacp, line 188: Javadoc for these new methods please.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -170,6 +184,121 @@ Tensor tensor(int outputIdx) {
...
+      TF_Input handles = new TF_Input(numConsumers);
+
+      TF_OperationOutputConsumers(output, handles, numConsumers);
+
+      Set<GraphOperation> operands = new LinkedHashSet<>(numConsumers);
+      for (int i = 0; i < numConsumers; i++) {
+        TF_Input atPos = handles.position(i);
+        TF_Operation op = atPos.oper();
+        String opName = TF_OperationName(op).getString();
+        operands.add(graph.operationOrError(opName));
```
Craigacp, line 233: When could this one throw?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Constant.java
```diff
@@ -1355,13 +1356,15 @@
...
   @Endpoint(name = "constantOf")
   public static <T extends TType> Constant<T> create(Scope scope, T tensor) {
-    return new Constant<>(
-        scope
-            .env()
-            .opBuilder("Const", scope.makeOpName("Const"))
-            .setAttr("value", tensor)
-            .setAttr("dtype", tensor.dataType())
-            .build());
+    OperationBuilder builder = scope
```
Craigacp, line 1359: Why is this changed?

Craigacp (commented):
There are a lot of unnecessary formatting changes in this PR. Could you back those out?

Craigacp:
It might be useful to have two extra graph walking functions one which traces all the ops which depend on a set of inputs, and one which traces all the ops which are depended on by a set of outputs. That way you can take any graph and figure out what you need to feed into it for a given output. That's not likely to be in a production system, but it would be super useful when poking a model in jshell. I've had to write ad-hoc things which do that when I was trying to understand the TF 1.x graph structure for saving, and having a supported method would be nice.

rnett:
> It might be useful to have two extra graph walking functions one which traces all the ops which depend on a set of inputs, and one which traces all the ops which are depended on by a set of outputs. That way you can take any graph and figure out what you need to feed into it for a given output. That's not likely to be in a production system, but it would be super useful when poking a model in jshell. I've had to write ad-hoc things which do that when I was trying to understand the TF 1.x graph structure for saving, and having a supported method would be nice.

Yeah, I can do that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Constant.java
rnett: Because it wasn't applying the scope correctly (missed the `scope.apply` call), control dependencies and the device were missed.  The control deps are what caused the issue here, but I ran into the device elsewhere as well.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -128,6 +136,77 @@ public GraphOperation operation(String name) {
...
 
+  /**
+   * Returns the operation (node in the Graph) with the provided name, or throws {@link IllegalArgumentException} if
+   * there isn't one.
+   *
+   * @param name name of the operation to look for
+   * @return operation in the graph with this name
+   * @throws IllegalArgumentException if no such operation exists in the Graph
+   */
+  public GraphOperation operationOrError(String name) {
```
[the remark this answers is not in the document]
rnett: Cause that's a lot of boilerplate every time you want to get a for-sure existing op.  There's multiple places where it's used currently (graph walking and session, mostly), so we'd have utility functions for it anyways.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: The reason I made it take nulls in the first place is because calling with both as emptySet errors whenever something is in the body (since we're allowing the empty set of ops).  I think an overload is a good idea though, and specifying the first two args must be non-null.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: See one of the other comments, nulls are necessary to distinguish from "only allow the empty set" which would error whenever something is in the body.  Optionally we could do that behavior (i.e. allow everything) for the empty set, but imo nulls is clearer (and what if that's what you want?).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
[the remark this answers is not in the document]
rnett: A lot of the uses cases are places where we require the op to exist, and it should always exist, but if it somehow doesn't throwing the IllegalStateException is clearer and easier to debug than an NPE.

rnett:
> There are a lot of unnecessary formatting changes in this PR. Could you back those out?

This is due to my auto format on commit, so not really, at least not without it being a big pain.  It's google-java-format, though, so it should really have already been done.  Imo we should add a format check for PRs in the CI (and format everything existing to match), at least once it's not having time issues.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
Craigacp: Why would we want the behaviour of only allowing the empty set? So you could catch the runtime exception and have different behaviour if there were intermediate nodes?

Anything that implies users have to catch a subclass of `RuntimeException` isn't good.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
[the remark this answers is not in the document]
rnett: As for the name, imo `OrError` sounded/read better than `OrException`, although now that I think about it `OrThrow` seems fine.  Or we could make these the default methods and use `OrNull` for the nullable ones.  Thoughts?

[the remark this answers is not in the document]
Craigacp: I think I prefer `orThrow`, as that's more similar to the methods in the JDK - https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/Optional.html#orElseThrow(). It's worth noting that if we know the exception can't happen, then the one that returns null will likely be better optimised by the JIT compiler because exceptions get in the way.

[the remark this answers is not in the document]
rnett: That makes sense.  Currently the only uses of the throwing variants are with user input, or in the graph walking functions.  The only case where the graph walking functions would have issues is if the graph is changed in another thread and the inputs were removed (which we don't even expose APIs for) so I'll go ahead and convert them to the non-throwing variant.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: It's for 0 input intermediate nodes specifically, i.e. things that don't depend on the inputs.  Generally `Constant` is ok and the others (`Placeholder`, `Variable`, etc) are ok, but since I'm making it a parameter "none" is an option.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -224,33 +226,62 @@ private GraphOperation graphOp(Operand<?> operand) {
...
 
+    if (inputs == null) {
+      throw new IllegalArgumentException("Inputs can't be null.");
+    }
+
+    if (outputs == null) {
+      throw new IllegalArgumentException("Outputs can't be null.");
+    }
+
     Queue<GraphOperation> currents = new LinkedList<>();
```
Craigacp, line 283: Given this is an `ArrayDeque` in the other new methods, maybe this one should be an `ArrayDeque` not a `LinkedList`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -304,6 +336,80 @@ private GraphOperation graphOp(Operand<?> operand) {
...
   }
 
+  /**
+   * Get all ops directly or indirectly required to calculate {@code outputs} (not including {@code outputs}), including
+   * control dependencies.
+   *
+   * @param outputs the starting points of the traversal.
+   * @return the ops needed to calculate {@code outputs}, not including {@code outputs}
+   */
+  public synchronized Set<GraphOperation> upstreamOps(Set<GraphOperation> outputs) {
```
Craigacp, line 346: I missed this earlier, but why are these methods `synchronized`? It seems like they don't mutate the state of the Graph.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -304,6 +336,80 @@ private GraphOperation graphOp(Operand<?> operand) {
...
+
+  /**
+   * Get all ops directly or indirectly required to calculate {@code outputs} (not including {@code outputs}), including
+   * control dependencies.
+   *
+   * @param outputs the starting points of the traversal.
+   * @return the ops needed to calculate {@code outputs}, not including {@code outputs}
+   */
+  public synchronized Set<GraphOperation> upstream(Set<Operand<?>> outputs) {
+    return upstreamOps(outputs.stream().map(this::graphOp).collect(Collectors.toSet()));
```
Craigacp, line 391: What does this do if you hand it a set of `Operand`s from `EagerOperation`s?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
Craigacp: Ok. I'm still not sure that the semantics of this is quite right, but given we've not stabilised the API yet lets add it in and use it a bit first before making further decisions.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: It doesn't, but if you mutate the graph while they are running the results could be incorrect.  More for downstream (with consumers) though, like if you added a new op using one of the inputs that had already been processed.  I can remove that from upstream.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: They should be LinkedList too, I think, for performance.  All I'm doing is front removal and back append, no indexing.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: `graphOp` will throw (with a nice message), since it calls `checkInput`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
Craigacp: Ok.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
Craigacp: The performance of `LinkedList` is bad due to having to do two pointer chases to find any given object and thus likely induce two cache misses (one for the node and one for the object itself). I'd expect `ArrayDeque` to be faster given it is based on a circular array, so it's likely to have the array always in cache when it's being traversed.

However performance isn't likely to be an issue in this code, so provided they are all the same I don't care.

Craigacp: Polling takes about the same amount of dereferences, either `LinkedList -> Node -> Object` or `ArrayDeque -> backing array -> Object`.

Craigacp (dismissed):
LGTM.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -170,6 +184,155 @@ Tensor tensor(int outputIdx) {
     throw new IllegalStateException("Graph tensors must be fetched by running a session");
   }
 
+  /**
+   * Get the number of inputs to the op, not including control inputs.
+   */
+  public int numInputs() {
+    try (PointerScope scope = new PointerScope()) {
+      return TF_OperationNumInputs(getUnsafeNativeHandle());
+    }
```
karllessard, line 193: We don't need a pointer scope here since we call straight a native endpoint without allocating anything

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -138,14 +219,205 @@ public GraphOperation operation(String name) {
...
+   * forbiddenNoInputBodyOps}) if null.
+   * @param forbiddenNoInputBodyOps types of ops to never allow as 0-input ops in the body.  Forbids all (except {@code
+   * allowedNoInputBodyOps}) if null.
+   * @return the set of operations needed to calculate outputs from inputs, including outputs and inputs
+   * @throws IllegalStateException if outputs depends on ops outside of the subgraph (i.e. is not calculable based
+   * solely on inputs)
+   * @see #completeSubgraph(Set, Set, boolean)
+   */
+  public synchronized Set<GraphOperation> completeSubgraph(Set<Operand<?>> inputs, Set<Operand<?>> outputs,
+      Set<String> allowedNoInputBodyOps, Set<String> forbiddenNoInputBodyOps) {
```
karllessard, line 265: These two last extra parameters obfuscate a bit the overall solution, which would be very simple without them. Why can't we simply include all ops we encounter without this logic, whether they have inputs available or not, and let the caller filter them out after if he needs to?

rnett: Yeah, that's fine.  I don't like how easy it is to end up including things upstream from the inputs in the body then, but I don't see much of a way around it, since we can't just forbid it altogether.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -128,6 +138,77 @@ public GraphOperation operation(String name) {
...
+      if (operation == null) {
+        return null;
+      }
+      return new Output(operation, index);
+    } catch (NumberFormatException e) {
+      GraphOperation op = operation(output);
+      if (op == null) {
+        return null;
+      }
+      return new Output(op, 0);
```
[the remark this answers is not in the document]
rnett: I'm not.  I think it was to handle cases where you had a `:` in the name but no output index, but that seems forbidden in the front end at least.  Tests pass though, so I think we're fine.

karllessard:
@rnett looks good, are you done with the changes? If so, can you please check why the quick build has failed?

rnett:
It's the javadoc issues in `framework`.  I think what's happening is `core-api` is set to ignore Javadoc failures, so Javadoc fails but the build succeeds, but then `framework`, which isn't set to ignore Javadoc failures, fails when it tries to depend on them.  I'm pretty sure this is an issue on master as well, I didn't add anything that should fail.

karllessard:
Ok, I see. Indeed, the javadoc validation [is disabled](https://github.com/tensorflow/java/blob/0f0c638b691907e954e3f3bbff465a96c354a71d/tensorflow-core/tensorflow-core-api/pom.xml#L429) in the `tensorflow-core-api` since this generated documentation from Markdown is not 100% legit right now. 

Now what you are saying is that the framework documentation with links to the generated classes fail to pass this validation because it tries to validate the documentation it links to? 

I was not aware of that but I guess it's possible, @JimClarke5 can we clean up the framework documentation to at least fix the CI build, even if it means to disable these links in it? 

I will merge this PR then, thanks @rnett !

karllessard (approved):

JimClarke5:
There is a C library for commonmark, but it would need to be extended with a JavaDoc renderer based on the HtmlRenderer. I have done this with Python and am now working on the Java version.  I could look at the C version and determine what needs to be done.

karllessard:
Ok @JimClarke5 I'll let you judge how long it would take but for now the priority is really to fix the CI build so that snapshots get refreshed on Sonatype after merging a PR. Do you think we can simply fix the framework quickly by removing links to the generated ops?

JimClarke5:
@karllessard  Let me experiment with the [CommonMark-cmark](https://github.com/commonmark/cmark) library. That would be the quickest fix and incorporate it into SourceWriter.cc

rnett:
Couldn't we, for now, just let framework build w/ failing javadocs?  W/ a note to fix before the next release or something.

JimClarke5:
@karllessard please explain exactly what you mean by “ quickly by removing links to the generated ops?”

JimClarke5:
@karllessard As a test, I have been able to transform the markup from api_def_UnsortedSegmentMax.pbtxt

```
  description: <<END
Read
[the section on segmentation](https://tensorflow.org/api_docs/python/tf/math#Segmentation)
for an explanation of segments.

This operator is similar to the unsorted segment sum operator found
[(here)](../../../api_docs/python/math_ops.md#UnsortedSegmentSum).
Instead of computing the sum over segments, it computes the maximum such that:

\\(output_i = \max_{j...} data[j...]\\) where max is over tuples `j...` such
that `segment_ids[j...] == i`.

If the maximum is empty for a given segment ID `i`, it outputs the smallest
possible value for the specific numeric type,
`output[i] = numeric_limits<T>::lowest()`.

If the given segment ID `i` is negative, then the corresponding value is
dropped, and will not be included in the result.
```

to

```
/**
 * Read <a href="https://tensorflow.org/api_docs/python/tf/math#Segmentation">the section on segmentation</a> for an  
 * explanation of segments.This operator is similar to the unsorted segment sum operator found        
 * <a href="../../../api_docs/python/math_ops.md#UnsortedSegmentSum">(here)</a>.Instead of computing the sum over
 *  segments, it computes the maximum such that:\(output_i = \max_{j...} data[j...]\) where max is over tuples {@code j... } 
 * such that {@code segment_ids[j...] == i }.If the maximum is empty for a given segment ID {@code i }, it outputs the 
 * smallestpossible value for the specific numeric type,{@code output[i] = numeric_limits<T>::lowest() }.If the given segment 
 * ID {@code i } is negative, then the corresponding value isdropped, and will not be included in the result.
 * /

```
The only issue so far is the relative URL `../../../api_docs/python/math_ops.md#UnsortedSegmentSum` in the second link.
With a little brute force, I think I can change that to `{@link MathOps#unsupportedSegmentSum}`.

I can change SourceWriter.cc to call out to new C methods to handle the JavaDoc rendering and in turn link in the cmark C library. I will also have to add the external cmark source to` bazel_bin-external` ( I think ). Is external pulled down from GitHub somewhere? Also, there may be some edge cases I have to handle as I expand out to the full api_def directory.

karllessard:
@JimClarke5 , I would say don't put too much effort of this C++ markdown converter because @rnett is working on porting the op generator to Java anyway (#244). Unless Ryan you think that we could use it to convert the documentation as we generate the new ops.pb file?

rnett:
I mean, we probably could, but given this is a Java library originally (if I understood things right) it's probably easier to wait.

PR #235, opened 2021-03-09
Add easy to use configuration scripts

### LINKED ISSUE #206: README should include an example of a build command with explicit component paths
**System information**
- TensorFlow version: 2.3.1
- Doc Link: https://github.com/tensorflow/java (the README.md file)

**Describe the documentation issue**

The README.md file should provide a build example for an arbitrary environment, which contains all of the required components, but not necessarily configured as the default choices. This is often the case with ML development workstations, which have multiple versions of CUDA, cuDNN, gcc, and friends, since different projects require the developer to use different versions for reproducibility reasons.

This example uses Bazelisk to manage multiple Bazel versions.

`TF_CUDA_COMPUTE_CAPABILITIES` will require #195 to be merged.

```
TF_CUDA_COMPUTE_CAPABILITIES=3.5,3.7,6.1,7.0,7.5 \
GCC_HOST_COMPILER_PATH=/usr/bin/gcc-8 \
CC=/usr/bin/gcc-8 \
CXX=/usr/bin/g++-8 \
TF_CUDA_PATHS=/usr/local/cuda-10.2.89,/usr/local/cudnn-10.2-7.6.5.32 \
CUDA_TOOLKIT_PATH=/usr/local/cuda-10.2.89 \
CUDNN_INSTALL_PATH=/usr/local/cudnn-10.2-7.6.5.32 \
USE_BAZEL_VERSION=3.1.0 \
TMP=/tmp \
mvn install -Dmaven.test.skip=true -Djavacpp.platform.extension=-gpu
```

**We welcome contributions by users. Will you be able to update submit a PR (use the [doc style guide](https://www.tensorflow.org/community/documentation)) to fix the doc Issue?**

No.

rnett:
Adds bazel configuration scripts that download the tensorflow binary and call its configure scripts.  The `bazel info output_base` use is because in some cases (such as switching OSs), the symlinks become corrupted and bazel doesn't generate new ones.

Fixes #206


tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -2,6 +2,10 @@
 # Script to build native TensorFlow libraries
 set -eu
 
+if [ ! -f .tf_configure.bazelrc ]; then
+    echo "WARNING: No .tf_configure.bazelrc, using CI settings.  For local builds, run the ./configure script." >&2
```
karllessard, line 6: I would remove the "WARNING:" prefix, it is legitimate and common to use the CI settings for a local build (e.g. on a different Linux distribution).

tensorflow-core/tensorflow-core-api/configure
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+#
+# /* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
```
karllessard, line 4: Unnecessary `/*`

karllessard (approved):
Looks good, just left two small comments there but it's approved for merging

tensorflow-core/tensorflow-core-api/configure
Craigacp: And a `*/` at the end of the license block

rnett:
Your approval got removed when I pushed changes, and I don't think I can override the CI.

tensorflow-core/tensorflow-core-api/.bazelversion
```diff
@@ -0,0 +1 @@
+3.7.2
```
karllessard, line 1: Is this file required? The version does not match the one we currently use in the CI (3.1.0) so I'm afraid it can cause some sort of conflicts?

Anyway I think I'll merge this after releasing 0.3.0 since that feature is only useful if you compile from sources so no need to have it now.

rnett: We could (and probably should) change it to what the CI uses.  It's necessary for anyone using bazelisk, which afaik is the recommended way to use bazel, since otherwise it automatically uses the latest bazel version (4.0.0) which isn't supported by `tensorflow` yet.

Optionally we could just gitignore it and leave it to users, but imo matching the CI is better.

PR #244, opened 2021-03-14
Move op generation to Java

rnett:
This is the start of moving the op class generators to Java.  It works fine on Linux, which is where I'm assuming the old op generator is normally run.  

However, there are some output differences on Windows:
 * Lots of parameter descriptions are missing.
 * A few ops are missing, a few new ones are added.
 * Some class names don't line up.  For example, we're currently generating `BroadcastArgs` with the class name `BroadcastDynamicShape`, and `All` as `ReduceAll`.  The class name comes from the `ApiDef` endpoints on both generators, I am unsure why they differ.

I don't think there's really anything we can do about this at the moment, it's just a bit odd to have a java program that you can't run on all platforms.


rnett:
Not sure what's up with the Javadoc generation failing on the CI.  I saw it a few times locally, but a clean fixed it.

rnett:
There's also a bunch of ops that start with `_` that seem to not be detected by C++.  They aren't in `ops.pbtxt`, so I'm assuming they are added in code rather than statically, after the point where the C++ generator gets it's defs.  There's some interesting ones there like `_XlaCompile`, but I'm not sure if we should include them.

rnett:
For people working on windows, this can be run (correctly) via WSL.  You need to get the api defs from somewhere, but a clone of `tensorflow` on the proper release branch (i.e. `r2.4`) will do.

I set up the exec plugin to allow for this, although since it doesn't use the working dir when running java the command gets a bit long.  I'm currently using something like `mvn exec:java -Dexec.arguments=../tensorflow-core-api/src/gen/java,$clonedTensorflow/tensorflow/core/api_def/base_api,../tensorflow-core-api/src/bazel/api_def -f pom.xml -Pdef -Pjdk11`.  Intellij can run this on WSL as part of the run configuration with the latest EAP, see [this](https://blog.jetbrains.com/idea/2021/01/run-targets-run-and-debug-your-app-in-the-desired-environment/).

rnett:
Alright, I fixed the build issues, they are only present when including `tensorflow-core-platform` as a compile time dependency, and I can include it as a runtime dependency instead.  However, we should make sure this doesn't affect 3rd party libraries depending on `tensorflow-core-platform` (they should probably use `-api`, but in some cases they may need the runtime).

rnett:
Since we don't need javadocs for the generator, can we skip them?  Or will that cause issues with central?  I don't think it would be published with the way it is configured atm, and we certainly don't need it to be.

saudet:
BTW, TF Core needs MSYS2 to build, it doesn't work with WSL...
https://www.tensorflow.org/install/source_windows

Craigacp:
> Since we don't need javadocs for the generator, can we skip them? Or will that cause issues with central? I don't think it would be published with the way it is configured atm, and we certainly don't need it to be.
> 
> `tensorflow-core-api` doesn't fail on javadoc errors, so for now I'm going to do the same.

Javadoc isn't just for users, it's for us in 6 months time. I think all new code in this repo should have it whether we expect people to use the class or not (excluding test code).

rnett:
Yeah, that's one of the advantages of doing the op gen in Java, it works fine with `dev`, we don't need to build TF Core.

rnett:
> > Since we don't need javadocs for the generator, can we skip them? Or will that cause issues with central? I don't think it would be published with the way it is configured atm, and we certainly don't need it to be.
> > `tensorflow-core-api` doesn't fail on javadoc errors, so for now I'm going to do the same.
> 
> Javadoc isn't just for users, it's for us in 6 months time. I think all new code in this repo should have it whether we expect people to use the class or not (excluding test code).

Oh yeah, the code will have it.  I just mean building the javadoc jars.

saudet:
> Yeah, that's one of the advantages of doing the op gen in Java, it works fine with `dev`, we don't need to build TF Core.

Right, but if we need WSL, it's going to cause problems for CI. I imagine this can eventually be adjusted to work with MSYS2 as well. Just wanted to point it out for now.

rnett:
> > Yeah, that's one of the advantages of doing the op gen in Java, it works fine with `dev`, we don't need to build TF Core.
> 
> Right, but if we need WSL, it's going to cause problems for CI. I imagine this can eventually be adjusted to work with MSYS2 as well. Just wanted to point it out for now.

Well, it turns out it works fine on Windows if I set the api docs dir correctly...  

The larger issue here is that it depends on `tensorflow-core-platform`, so we would need a 2 phase build.  We might be able to move some of the stuff out, the only thing it really needs is `TF_NewApiDefMap` and `TensorFlow.registeredOpList()`.

Craigacp:
> > > Since we don't need javadocs for the generator, can we skip them? Or will that cause issues with central? I don't think it would be published with the way it is configured atm, and we certainly don't need it to be.
> > > `tensorflow-core-api` doesn't fail on javadoc errors, so for now I'm going to do the same.
> > 
> > 
> > Javadoc isn't just for users, it's for us in 6 months time. I think all new code in this repo should have it whether we expect people to use the class or not (excluding test code).
> 
> Oh yeah, the code will have it. I just mean building the javadoc jars. `core-api` allows the build to proceed if there's errors, so that's what I'm doing here. As this isn't being published though, we could disable it altogether.

Looks like something is currently published - https://search.maven.org/artifact/org.tensorflow/tensorflow-core-generator/0.2.0/jar, and so if we still want to publish it then it will need javadoc and sources jars.

Does this only happen on Windows? I've not had chance to look through the PR yet, is there an error message somewhere from javadoc I could look at?

saudet:
> Well, it turns out it works fine on Windows if I set the api docs dir correctly...

Oh? Weird.

> The larger issue here is that it depends on `tensorflow-core-platform`, so we would need a 2 phase build. I might be able to get it on it's own if that's an issue, the only thing it really needs is `TF_NewApiDefMap` and `TensorFlow.registeredOpList()`. But I would think with the new caching that shouldn't be too much overhead.

That's not an issue, we just need to run the build with a specific platform like `mvn -Djavacpp.platform=windows-x86_64 ...`, see https://github.com/bytedeco/javacpp-presets/wiki/Reducing-the-Number-of-Dependencies

rnett:
> Looks like something is currently published - https://search.maven.org/artifact/org.tensorflow/tensorflow-core-generator/0.2.0/jar, and so if we still want to publish it then it will need javadoc and sources jars.
> 
> Does this only happen on Windows? I've not had chance to look through the PR yet, is there an error message somewhere from javadoc I could look at?

Huh, that's the annotation processor but if it's published this will be too.

It doesn't happen locally for me, just in the CI (quick build is on linux).  I suspect what is happening is that `core-api` has `failOnError=false` for javadocs, it fails but the build succeeds, and then when `op-generator` tries to build it's javadocs, it errors because it can't find the core javadocs.  To see it happen you'll need to remove the `failOnError=false` from `op-generator`.

rnett:
> > Well, it turns out it works fine on Windows if I set the api docs dir correctly...
> 
> Oh? Weird.

There's one or two differences, but I think it's just because I'm building off a cloned tensorflow instead of the downloaded source archive.

> > The larger issue here is that it depends on `tensorflow-core-platform`, so we would need a 2 phase build. I might be able to get it on it's own if that's an issue, the only thing it really needs is `TF_NewApiDefMap` and `TensorFlow.registeredOpList()`. But I would think with the new caching that shouldn't be too much overhead.
> 
> That's not an issue, we just need to run the build with a specific platform like `mvn -Djavacpp.platform=windows-x86_64 ...`, see https://github.com/bytedeco/javacpp-presets/wiki/Reducing-the-Number-of-Dependencies

I assume that's done by the CI.  My issue is more that the op generation is done in `build.sh`, which is used to build `core-api`, but the new generation depends on `core-api`.  So we'd need to build the core artifacts with the current generation (what was pushed), regenerate the ops, then build again.  Probably as part of the CI rather than in `build.sh`.

rnett:
@Craigacp `framework` fails as well now, so I'm pretty sure I'm right about the reason.  Not sure why that doesn't happen locally, I may have a javadoc disable setting floating around.

This may be because the newly generated op javadocs are slightly different, but I'm pretty sure this happens on `master` too, it was failing even before I pushed the newly generated ops and is doing so on other PRs as well.  I'm not sure why the release builds don't fail.

Craigacp:
It looks like a lot of the changes are because the `@return` javadoc param moved in the generated ops. Is it possible to move it back before we start to review this in earnest? I see there are a bunch of other small javadoc changes, but those appear to actually change the generated output and so we should review them. Given how big the PR is anything that you can do to reduce the code movement noise will make it a lot easier to review.

rnett:
That's due to the way JavaPoet handles parameter javadocs.  I should be able to work around it, but it may not be pretty.

rnett:
I have a better solution to the two phase builds.  We can still use the native code for op detection, but output the OpDef and ApiDef protos to the Java generator to do the generation.  I'd like to it via pipe but I need to look into protobuf a little more.  The only thing is that we need to include the proto defs in the generator.  Since they are auto-generated anyways I may just be able to add the dir as a source file, otherwise we can either copy them or extract them to another module.  Any preferences there?

saudet:
We could potentially just bundle them in the JAR file. It's really easy to
bundle and extract things like that as "native resources" with JavaCPP.
Check what it looks like for files from CPython here:
https://github.com/bytedeco/javacpp-presets/blob/master/cpython/src/main/java/org/bytedeco/cpython/presets/python.java#L192
https://github.com/bytedeco/javacpp-presets/blob/master/cpython/src/main/java/org/bytedeco/cpython/presets/python.java#L213

karllessard:
Thanks @rnett , the size of that PR is quite impressive and would take time to review. I remember in the days I've written the C++ version, my Google reviewer asked me to break it down into smaller pieces and I think I understand why ;) Any chance you can do the same?

On a different note, here's just a few thoughts:
* the `tensorflow-core-generator` was created having in mind to contain both the annotation processor for the `Ops` and the converted Java generator for the wrappers in the same JAR, since they are pretty related. Is it possible to do it?
* what are the real gains of migrating this code to Java, other that facilitate its maintenance?
* the cyclic dependency between the `tensorflow-core-api` (which contains the native artifact) and `tensorflow-core-generator` was an issue, indeed. Some solutions are:
    * move the Bazel build directly in the `tensorflow-core` parent so every submodule can access the native libraries and files, would require some additional tweaks in the `tensorflow-core-api` pom.
    * move all [Java API def protos](https://github.com/tensorflow/java/tree/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def) to the generator module instead
    * run the generator not on the native library but by consuming the [`ops proto file`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt), which should be pretty accurate.

rnett:
> Thanks @rnett , the size of that PR is quite impressive and would take time to review. I remember in the days I've written the C++ version, my Google reviewer asked me to break it down into smaller pieces and I think I understand why ;) Any chance you can do the same?

Yeah, I can leave the actual generation for another PR, just set it up so that we could use it.  I'll leave it in for now so you can see what the generated code looks like, and pull it out sometime in the next few days.  Depending on how you do the review, I've tried (and I think been successful at) keeping the generated code in separate commits, so you can always just drop them locally.

> On a different note, here's just a few thoughts:
> 
The `tensorflow-core-generator` was created having in mind to contain both the annotation processor for the `Ops` and the converted Java generator for the wrappers in the same JAR, since they are pretty related. Is it possible to do it?

It should be.  The reason I didn't is because I was using `tensorflow-core-api` as a dependency, but I'm going to move off of that anyways.  Depending on how that goes I think it would be good to include a version that depends and runs based off of the `core-api` dependency instead of the native build so that people can work on the generator in `dev` mode, but how exactly I'll do that is still up in the air.

What are the real gains of migrating this code to Java, other that facilitate its maintenance?

It's mostly just maintenance.  I know @JimClarke5 mentioned writing a better markdown converter.  The original reason I did it is I was running into segfaults when trying to add function generation, but it turns out that's because I was running the wrong thing.

I also want to try some more complicated stuff with functions.  What exactly that looks like depends on what we do with the function API, but ideally I'd like to be able to pass lambdas to ops that take functions.  There's also capture handling, where ops like `If` take inputs that they feed to their branch functions, which enables the functions to use captures (once that's supported, anyways).  Additionally, the optionally stateful ops can be disambiguated when called, i.e. we have a single `If` op at the Java level that calls `StatelessIf` or `If` depending on if the passed functions are stateful.  All of that sort of stuff is stuff I would much rather do in Java.

> * the cyclic dependency between the `tensorflow-core-api` (which contains the native artifact) and `tensorflow-core-generator` was an issue, indeed. Some solutions are:
>   
>   * move the Bazel build directly in the `tensorflow-core` parent so every submodule can access the native libraries and files, would require some additional tweaks in the `tensorflow-core-api` pom.
>   * move all [Java API def protos](https://github.com/tensorflow/java/tree/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def) to the generator module instead
>   * run the generator not on the native library but by consuming the [`ops proto file`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt), which should be pretty accurate.

I'll update this PR as I do this, but my current approach is doing something like @saudet mentioned, and creating what is essentially our own `ops.pbtxt` file during the native build (but with API defs included), and then doing the generation based on that.  I want to see how big it is before including it in the jar, but even if we don't include it we can get the same info using the native library.  You'd just end up with two generator artifacts, one that depends on `core-platform` and gets it's info from there, and one that accepts the file as input (streamed, if it's big).

I need access to the protos either way, do you have any preference between:
1. a separate `tensorflow-core-proto` artifact that both generator and core depend on
2. duplicating them, copying the protos to both generator and core during native build
3. putting the protos in generator and having core depend on it.  I don't like this since the main methods need to be public.
4. adding the generated proto dir in `core` (possibly moving it from `gen` to `proto` or similar, if supported) as an external source root for `generator` in maven

karllessard:
> I'll update this PR as I do this, but my current approach is doing something like @saudet mentioned, and creating what is essentially our own ops.pbtxt file during the native build (but with API defs included)

I cannot figure why we would prefer to build our own `ops.pbtxt` file if it is already available there, in the core sources. Only maintaining our own API defs, like we do now, sounds better. 

> I need access to the protos either way, do you have any preference between:

Ideally my preference would be 5. :) which is to run the Bazel build at the `tensorflow-core` level and let it's child modules access the `bazel-*` folders to do whatever they need. The annoying part is that the child modules will depend on external (parent)  directories but it is still manageable. 

In other words, I'm dreaming since the first day of a file structure like this:

* tensorflow-core
    * bazel-bin (created by Bazel)
    * bazel-out (created by Bazel)
    * bazel-tensorflow (created by Bazel)
    * BUILD
    * WORKSPACE
    * build.sh
    * patches
    * api-defs
    * tensorflow-core-api (access parent bazel output directories to generated JavaCPP bindings and creating native jars)
        * src/gen
        * src/main
    * tensorflow-core-generator
        * src/main/org/tensorflow/generator/ops/annotation
        * src/main/org/tensorflow/generator/ops/wrappers

When invoking the `tensorflow-core-generator` jar to generate the ops wrappers, the Maven task for `tensorflow-core-api` can pass in parameter the path to the `api-defs` and the `${project.parent.basedir}/bazel-out/.../ops.pbtxt` file. Then, in non-dev mode, JavaCPP can also retrieve the native libraries from `${project.parent.basedir}/bazel-out/...` directory the generate the Java C API bindings and packages them up with our native `tensorflow-core-api-*-*` artifacts.

saudet:
> Note that we can still do all of the above while keeping the actual file structure, that would work too, but I always got that feeling that the native libraries/files resulting from the Bazel build should be accessible to more than one module (`tensorflow-core-api` in this case).

All files are available to all other modules with the approach I'm talking about above with CPython. What's wrong with packaging those development files in JARs?

karllessard:
> All files are available to all other modules with the approach I'm talking about above with CPython. What's wrong with packaging those development files in JARs?

I'm not saying it's wrong but I would prefer avoiding having an additional JAR/module like `tensorflow-core-protos` and the copies of protos if we can. Why not simply have a small JNI file in the `tensorflow-core-generator` module that invokes `TF_GetAllOpsList` like it was done in the old days (https://github.com/tensorflow/tensorflow/blob/14dcf6d00e4b0cfdf8c60cc22787f1b0f954bbd8/tensorflow/java/src/main/native/tensorflow_jni.cc#L28)? 

The `tensorflow-core-generator` could be a Maven plugin that we pass a dependency to the `tensorflow-core-api-*-*` jar so it can call the underlying runtime library?

rnett:
A couple things:
* Tensorflow ApiDefs are not valid protos, they don't parse (at least in Java, last time I tried it).  It's the `<<<END` text block syntax they use.  You have to use the native ApiDefMap methods.
* The bazel dirs don't get created in `dev` mode.  Their creation is also pretty unreliable in general.  You can work around this in scripts (like the configure scripts) but it's not ideal.  When I run native builds on linux and then switch to windows, it won't replace the symlinks when I run stuff from windows.  Plus fetching the whole tensorflow lib takes a while.
  * This is why I wanted to bundle our ops def file as a resource, so it ran be ran properly in dev mode.
* There's ops in the runtime that aren't in ops.pbtxt.  If we want to continue to ignore them that's fine though.

JimClarke5:
<<END is called a here document and a similar syntax is also used in Unix Shell.

Also, I have found an open source Java project to parse Markdown similar to the python code I was using. [commonmark-java](https://github.com/commonmark/commonmark-java) . This should make it easier to covert the Markdown in the apidef to JavaDoc.

saudet:
> I'm not saying it's wrong but I would prefer avoiding having an additional JAR/module like `tensorflow-core-protos` and the copies of protos if we can. Why not simply have a small JNI file in the `tensorflow-core-generator` module that invokes `TF_GetAllOpsList` like it was done in the old days (https://github.com/tensorflow/tensorflow/blob/14dcf6d00e4b0cfdf8c60cc22787f1b0f954bbd8/tensorflow/java/src/main/native/tensorflow_jni.cc#L28)?
> 
> The `tensorflow-core-generator` could be a Maven plugin that we pass a dependency to the `tensorflow-core-api-*-*` jar so it can call the underlying runtime library?

It doesn't need to be a plugin, but in any case, we can already call TF_GetAllOpList() from Java, yes, that's not a problem:
https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/global/tensorflow.java#L2926
Along with all those functions for TF_ApiDefMap.

> If you have a solution for providing the native library in dev mode this changes, but my thinking was if we need to produce an ApiDef map using the native library anyways, we might as well just add the op defs rather than trying to get bazel to fetch `ops.pbtxt`. And since we want to support dev mode, we need to save that file in the artifact. I don't even think we'd need JavaCPP, we could just put in `src/main/resources` (in core, and provide an accessor method).

We don't *need* JavaCPP, but it's 1 line to copy all resources at build time, and 1 line to load them in the cache. There are other libraries that do that too, it can be (re)done manually too, sure, but what's the point.

> I was thinking we'd probably end up with two generator modules. One for native builds that doesn't depend on `core` and takes the file as a parameter, and one that does depend on core and gets the file from core's resources (for dev mode). I don't see a way of avoiding two artifacts like that since they need different dependencies, but the dev mode one would be very light.

It doesn't need to be in the same JAR file. We can have a "dev" JAR where we just dump everything that wouldn't be needed by end users, just like Linux distributions do it. That way, the JAR file can also be downloaded with Maven automatically by developers that don't need to rebuild TF Core without having to deal with Bazel at all. With Maven profiles, the same artifact can have different dependencies, that's not a problem either.

karllessard:
@rnett  ok here's another idea then, why not bundling the `ops.pbtxt` file (generated or static, let's ignore this for now) under `/src/main/resources/` of the `tensorflow-core-api-*-*` JARs, and then add this artifact as a dependency to `tensorflow-core-generator` so it can retrieve the resources from it? The sequence would be:

Full build:
1) Build native library and run other Bazel goals 
2) Package native library, JavaCPP JNI code and `ops.pbtxt` in a `tensorflow-core-api-*-*` artifact
3) Run `tensorflow-core-generator` as an executable JAR (or Maven plugin?) with a dependency to `tensorflow-core-api-*-*` to generate the ops wrappers
4) Build and package the `tensorflow-core-api` JAR

Dev build:
1) Fetch the `tensorflow-core-java-*-*` JAR for the current platform
2) Run `tensorflow-core-generator` as an executable JAR (or Maven plugin?) with a dependency to `tensorflow-core-api-*-*` to generate the ops wrappers
3) Build and package the `tensorflow-core-api` JAR

This should be quite simple to achieve and is very close that we are already doing, without the need of reshuffling the modules at all.

It would have been interesting too to be able to call the JNI code that is already part of the `tensorflow-core-api-*-*` artifacts but I don't how we can do this without having its generated Java classes as well.

rnett:
> It doesn't need to be in the same JAR file. We can have a "dev" JAR where we just dump everything that wouldn't be needed by end users, just like Linux distributions do it. That way, the JAR file can also be downloaded with Maven automatically by developers that don't need to rebuild TF Core without having to deal with Bazel at all. With Maven profiles, the same artifact can have different dependencies, that's not a problem either.

Right, forgot about that, that's much better.  You're saying something like adding an execution to the jar plugin in `core` that creates a `dev` jar with only the op map resources?

The trouble then is getting the resource from the jar, I'd need to either pass the jar as a parameter when `dev` is set and somehow get the file, or do something with `Class.forName` or ServiceLoader, which should really go in a separate artifact.

rnett:
> @karllessard ...

When you're saying `tensorflow-core-java-*-*` you mean a new artifact?  That seems fine, it's along the lines of what I was thinking anyways.  That way I can add it as a resource and add accessor code instead of trying to add another jar to core somehow.

karllessard:
> When you're saying tensorflow-core-java-*-* you mean a new artifact?

No I mean for example the `tensorflow-core-api-0.3.0-linux-x86_64.jar`, we already build these first in before building the `tensorflow-core-api` itself (all done in the `tensorflow-core-api` module) and that's the one we are already fetching from Sonatype in dev model, so there won't be much to change. The `ops.pbtxt` file is almost 1M though but compared to that 59M the artifact already has, it might be negligible ;)

saudet:
> It would have been interesting too to be able to call the JNI code that is already part of the `tensorflow-core-api-*-*` artifacts but I don't how we can do this without having its generated Java classes as well.

I'm not following, why is that a problem exactly? We can just do it, right? What am I missing?

> Right, forgot about that, that's much better. You're saying something like adding an execution to the jar plugin in `core` that creates a `dev` jar with only the op map resources?

For creating another JAR, yes, that's correct.

> The trouble then is getting the resource from the jar, I'd need to either pass the jar as a parameter when `dev` is set and somehow get the file, or do something with `Class.forName` or ServiceLoader, which should really go in a separate artifact.

We just need to add it as a dependency in the pom.xml file, potentially in a dev profile like the one here already:
https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/pom.xml#L88-L102
And then it magically becomes available on the class path, from which it can easily be extracted to cache with Loader.load() or something. Give it a try with LLVM and CPython here to see how it works:
https://github.com/bytedeco/javacpp-presets/blob/master/tvm/samples/DeployBERT.java

karllessard:
> I'm not following, why is that a problem exactly? We can just do it, right? What am I missing?

Because the generator would run before the `tensorflow-core-api.jar` is built, where resides the Java classes, but after the native JARs are packaged, which only contain the JNI code, the TF runtime and now the `ops.pbtxt` file.

saudet:
> Because the generator would run before the `tensorflow-core-api.jar` is built, where resides the Java classes, but after the native JARs are packaged, which only contain the JNI code, the TF runtime and now the `ops.pbtxt` file.

That's how it works currently, but isn't the whole point of this PR to remove tensorflow-core-api's dependency on tensorflow-core-generator?

rnett:
> > Because the generator would run before the `tensorflow-core-api.jar` is built, where resides the Java classes, but after the native JARs are packaged, which only contain the JNI code, the TF runtime and now the `ops.pbtxt` file.
> 
> That's how it works currently, but isn't the whole point of this PR to remove tensorflow-core-api's dependency on tensorflow-core-generator?

No, it's going to depend on it anyways for annotation processing.  What we're trying to figure out is how to do the op class generation in `generator` without depending on `core-api`, since the generator needs to be ran first.

rnett:
> We just need to add it as a dependency in the pom.xml file, potentially in a dev profile like the one here already:
> https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/pom.xml#L88-L102
> And then it magically becomes available on the class path, from which it can easily be extracted to cache with Loader.load() or something. Give it a try with LLVM and CPython here to see how it works:
> https://github.com/bytedeco/javacpp-presets/blob/master/tvm/samples/DeployBERT.java

If I'm understanding that right, you generate a class for the native resource.  That still wouldn't work, since the generated class would be in `core-api`, which is what we're trying to remove the dependency on.

rnett:
> No I mean for example the `tensorflow-core-api-0.3.0-linux-x86_64.jar`, we already build these first before building the `tensorflow-core-api-0.3.0.jar` itself (all done in the `tensorflow-core-api` module) and that's also the JAR we are already fetching from Sonatype in dev mode, so there won't be much to change. The `ops.pbtxt` file is almost 1M though but compared to that 59M the artifact already has, it might be negligible ;)

Wait, you can access resources from dependencies' Jars?  I've been looking at [this](https://stackoverflow.com/questions/5292283/use-a-dependencys-resources) which implies you can't.  I'll test once the native build is done.

saudet:
> If I'm understanding that right, you generate a class for the native resource. That still wouldn't work, since the generated class would be in `core-api`, which is what we're trying to remove the dependency on.

So, backtracking a bit here, isn't there a version of this that could be sorted topologically?
tensorflow-core-generator <- tensorflow-core-api and/or maybe tensorflow-core-api:dev <- tensorflow-core-op-generator
Sounds like this could work. What would be wrong with that?

> Wait, you can access resources from dependencies' Jars? I've been looking at [this](https://stackoverflow.com/questions/5292283/use-a-dependencys-resources) which implies you can't. I'll test once the native build is done.

Yes, of course we can. That's what the class path is.

rnett:
> > If I'm understanding that right, you generate a class for the native resource. That still wouldn't work, since the generated class would be in `core-api`, which is what we're trying to remove the dependency on.
> 
> So, backtracking a bit here, isn't there a version of this that could be sorted topologically?
> tensorflow-core-generator <- tensorflow-core-api and/or maybe tensorflow-core-api:dev <- tensorflow-core-op-generator
> Sounds like this could work. What would be wrong with that?

`tensorflow-core-op-generator` (the op class generator in this PR, that needs the file/`core-api`) is being merged into `tensorflow-core-generator` (the existing annotation processor, that doesn't need the deps).

karllessard:
Looks like we are all confused now 😄 @rnett , I think you understood that with my proposal, we wont need a `tensorflow-core-op-generator`, right? These classes can simply be part of `tensorflow-core-generator`, which is either used as an executable JAR (operations wrappers) or as an annotation processor (`Ops` classes).

By the way when I meant having `tensorflow-core-api-*-*` as a dependency to `tensorflow-core-generator`, it's not by adding it in the generator's pom but simply by passing it in the classpath when executing the JAR.

saudet:
What do we gain by merging everything into tensorflow-core-generator? I don't see why we're trying to do that.

saudet:
BTW, initially I envisioned `tensorflow-core-api` building whatever we needed for C/C++ _only_ with Bazel, with nothing related to Java except JavaCPP. And then you'd have other modules like `tensorflow-core-generator` depend on it. What is preventing us from moving all the Java classes from `tensorflow-core-api` to some other module like `tensorflow-core-generator`?

rnett:
> Looks like we are all confused now  @rnett , I think you understood that with my proposal, we wont need a `tensorflow-core-op-generator`, right? These classes can simply be part of `tensorflow-core-generator`, which is either used as an executable JAR (operations wrappers) or as an annotation processor (`Ops` classes).

Yeah, that's what I ment, `op-generator` is being merged into `generator`, I just haven't actually done it yet.

> By the way when I meant having `tensorflow-core-api-*-*` as a dependency to `tensorflow-core-generator`, it's not by adding it in the generator's pom but simply by passing it in the classpath when executing the JAR.

Is there a reason to not just add it to the pom?  Or would that add `tensorflow-core-api` as a dependency?

The resource generation is easy enough to do in `build.sh`, but to run the generator I need a phase after the native jars are built but before `compile`, while the native jars are currently packaged in `package` like normal.  It might be possible to move all the JavaCPP tasks before that, but it seems like `javacpp-compiler` depends on the compiled class files, so maybe not.  I'm thinking it would just be easier to pass the generated resource file to the generator directly when building from native, and only use the dependency in `dev` mode.

saudet:
> BTW, initially I envisioned `tensorflow-core-api` building whatever we needed for C/C++ _only_ with Bazel, with nothing related to Java except JavaCPP. And then you'd have other modules like `tensorflow-core-generator` depend on it. What is preventing us from moving all the Java classes from `tensorflow-core-api` to some other module like `tensorflow-core-generator`?

Or actually, we'd want to move all the Java classes to something like `tensorflow-core-ops` such that end users wouldn't need to touch `tensorflow-core-generator`. The dependency graph would look like this with everything related to ops generation in `tensorflow-core-generator`:
 * tensorflow-core-api and/or maybe tensorflow-core-api:dev <- tensorflow-core-generator <- tensorflow-core-ops

What's wrong with this one?

karllessard:
> Is there a reason to not just add it to the pom? Or would that add tensorflow-core-api as a dependency?

It's because in a full build, the `tensorflow-core-api-*-*` artifacts won't be installed in your local repo since you are in the process of doing it when we'll run the generator, therefore Maven won't resolve it. I think it's also preferable not having this cyclic dependency. So basically the design would be simply to pass an `ops.pbtxt` to the classpath of the generator to make it run, no matter how it should work.

karllessard:
> What do we gain by merging everything into tensorflow-core-generator? I don't see why we're trying to do that.

I think it make perfect sense since the operations wrappers and the annotation processor are tightly coupled together, so I revert the question: why we should not do that? Especially if it is possible to avoid this cyclic dependency with the core-api the way I'm suggesting it.

saudet:
I think I understand what you guys are trying to do. Yes, we can run whatever we need to run after everything related to JavaCPP within the same `pom.xml` file, that works, I'm just not sure it's worth it though. That way, as you found out, it makes it hard for people to work on ops without having to go through the Bazel build. If we separate those into `tensorflow-core-api` and `tensorflow-core-ops`, or some other names like `tensorflow-core-native` for Bazel+JavaCPP and `tensorflow-core-api` for the ops, we would just need to run the build on the latter with `mvn -Pdev ...`, that would download everything we need from the former. In other words, we wouldn't need to give it anything else to run `tensorflow-core-generator` with precompiled resources!

rnett:
@saudet What's the proper way to add a resource to the native jar?  I tried adding it to the JavaCPP plugin under `resourcePath` or `buildResources` (`${project.basedir}/src/gen/resources/`), but neither worked.

saudet:
The `resourcePath` is like the `includePath`, `linkPath`, and `preloadPath`: It's not going to do anything on its down. It's used as the search path for the resources of modules, which we can specify on their classes with `@Platform(resource=...` like this:
https://github.com/bytedeco/javacpp-presets/blob/master/cpython/src/main/java/org/bytedeco/cpython/presets/python.java#L192

`buildResource` is for the other way around, when we want to use those resources that were previously bundled. JavaCPP extracts all the files from those resources in its cache for native build tools to use. If you're going to consume all those resources from Java, you probably don't need that.

rnett:
That works, but it puts it in the jar under `org/tensorflow/internal/c_api/linux-x86_64`, where I want it under `org/tensorflow/op` (or at a minimum, somewhere that isn't dependent on the platform).

rnett:
I cherry-picked the op exporting part to #247 so that it will make it into snapshots.  This PR will then just be the op generation classes usable w/ `dev`, and then will be another PR afterwards to wire it up to the CI build.

rnett:
I've also removed the generated ops and rebased, so this should be reviewable now (although I still need to update to use the exported op defs once #247 is merged, and move things into `generator`).

karllessard:
> That works, but it puts it in the jar under org/tensorflow/internal/c_api/linux-x86_64, where I want it under org/tensorflow/op

What will happen then if you have in your classpath more than one native artifacts (like those imported by `tensorflow-core-platform`? You'll end up having a conflict exporting the same file in different JARs. I don't know how it will react since it's not a class but I would feel more comfortable if you stick moving them in the OS package, aside the native libraries. It should not be a problem to refer to that file from the pom.xml since you are always building relatively to a platform anyway.

rnett:
> > That works, but it puts it in the jar under org/tensorflow/internal/c_api/linux-x86_64, where I want it under org/tensorflow/op
> 
> What will happen then if you have in your classpath more than one native artifacts (like those imported by `tensorflow-core-platform`)? You'll end up having a conflict exporting the same file in different JARs. I don't know how it will react since it's not a class but I would feel more comfortable if you stick moving them in the OS package, aside the native libraries. It should not be a problem to refer to that file from the pom.xml since you are always building relatively to a platform anyway.

Which one you get is undefined (although there's ways to get all and check), but shouldn't matter since they should all be identical.  There's no errors on conflicts.

The way I'm planning on handling passing it to the generator is to accept it as `args[1]` (used in `build.sh`), and if it's not specified there look for `org/tensorflow/op/ops.pb` with `getResourceAsStream` or the like (used when running w/ `dev`, via an `exec` plugin in `generator`).  Passing it as an argument is a little easier in `build.sh` than adding `src/gen/resources` to the classpath (although it's not much trouble), and allows for specifying a custom file if you need to test it with some other setup.  

Actually, I just realized we don't need to put it in the jar at all.  It's in the git repo, so I can just always add it to the classpath in `generator` or pass it as an argument.  When building with `dev` it will be the one from the git repo, when doing the native build it will have been replaced.  I prefer passing it as an argument over using the classpath since it's easier to do custom runs, but I don't have a strong preference if you really want to use the classpath.

Craigacp:
If we can put it in the git repo why doesn't it just live there permanently? Sure a 1.5MB text file is a bit of a pain, but it's significantly less painful than juggling this file and storing it in maven if we don't need to.

rnett:
That's basically what I'm thinking.  We still need to re-generate it during the native build and pass the updated version to the op generator, but that's pretty easy to do, just have to make sure it happens in the right order.

karllessard:
If we are generating the file, then yes it make sense to persist it in the repo and therefore we don't need to store it in the JAR, sgtm!

rnett:
Currently rebased on #247, the changelist got big because I moved the proto files, I'll extract that once #247 is merged.

JimClarke5:
In` org.tensorflow.op.generator.GeneratorUtils`, if you use [commonmark-java ](https://github.com/commonmark/commonmark-java/blob/main/README.md) you can create a `JavaDocRender` based on the `HtmlRenderer` and then the code would look like this.

    ```
 private static final Parser parser = Parser.builder().build();

    /**
     * Convert markdown descriptions to JavaDocs.
     */
    static String parseDocumentation(String docs) {
        Node document = parser.parse(docs);
        JavaDocRenderer renderer = JavaDocRenderer.builder().build();
        return renderer.render(document);
    }
```

This will handle all the markdown and I wrote custom code to change the links from `../../../api_docs/python/math_ops.md#UnsortedSegmentSum` to 
{@link org.tensorflow.op.MathOps#unsortedSegmentSum}

**Comment by JimClarke5** [comment]
FYI, OpGenerator.java line 215, `Files.writeString` is a Java 11 only method. This fails when source is set to 8.

```
Files.writeString(outputFile.toPath(), builder.toString(), StandardOpenOption.WRITE,
              StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
```

**Comment by rnett** [comment]
Ugh, I have the source level set to 8, that should have been caught.  There's probably a few others, I'll check.

**Comment by rnett** [comment]
Ok, what's getting me here is that the `jdk11` profile sets the source version to 11.  Is there a reason we did that (as opposed to just target and release)?  Since the sources need to be compatible with 8 anyways.  I could see it if we end up using split Jars but it would be nice to do it selectively rather than on the whole project (i.e. in a `java11` folder, which I think is how it's normally done?).

**Comment by JimClarke5** [comment]
I have finished my branch with the` commonmark-java` library.  In the process, I have fixed a number of issues with the actual JavaDoc generation, mainly adding default descriptions when the pbxt files were missing such items.. Now, I can produce all the classes with no `javadoc` errors or warnings.  Next step is to sample the various generated HTML pages to see if there are any obvious things I did wrong.

Let me know when you want me to push it.

**Comment by rnett** [comment]
Can you PR against my branch in my fork?  I'm fairly sure that's allowed.

**Comment by JimClarke5** [comment]
@rnett I rebased commommark-java and just pushed it. Let me know if it looks good.

**Comment by rnett** [comment]
The git tree got worse somehow, I'm just going to cherry pick your commits from there.

**Comment by rnett** [comment]
@JimClarke5 I think I got everything docs-related, can you check it over (the last 11 commits, specifically).  Assuming I did, it looks good to me.

**Comment by JimClarke5** [comment]
@googlebot I consent.

**Comment by JimClarke5** [comment]
@rnett I am not seeing my changes yet in the source directly from your branch. Maybe it is not updated yet.

In method `buildClass()`, you should have something like this code.


  ```
  String summary = parseDocumentation(apiDef.getSummary());
    if (!summary.isEmpty()) {
      builder.addJavadoc("$L", summary + "\n");
    } else {
      builder.addJavadoc("$L", String.format("The %s operation\n", apiDef.getGraphOpName()));
    }
```

rnett:
The closest I see is [this](https://github.com/rnett/java/blob/rn_java_op_gen/tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/ClassGenerator.java#L230), it might have gotten clobbered in the merge.

JimClarke5:
I can sent you my version and your can compare. It's just the one class.

rnett:
Ok, looking at the diff, I think I got everything.

karllessard:
Ok, so it happens that the javadoc issue that was preventing the quick-build to pass was unrelated to this. GitHub Action container has updated their default JDK from 1.8.0_282 to 11.0.10 right where it started to failed.

I'm gonna fix that first, probably by forcing a specific JDK version and then start to review and merge other PRs, including this one.

tensorflow-core/tensorflow-core-generator/pom.xml
```diff
@@ -28,15 +28,69 @@
...
+          -->
+          <execution>
+            <id>add-gen-sources</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <configuration>
+              <sources>
+                <source>${project.basedir}/src/gen/protoc</source>
```
karllessard, line 69: I was under the impression that only the protos we need for generating the ops would be copied to the generator source tree (directly with other java sources, under `src/main/java`). AFAIK, there should be only a few needed.

We won't need then this `build-helper-maven-plugin`.

tensorflow-core/tensorflow-core-generator/pom.xml
```diff
@@ -28,15 +28,69 @@
...
+            <goals>
+              <goal>java</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <mainClass>org.tensorflow.op.generator.OpGenerator</mainClass>
+          <arguments>
+            <argument>${project.basedir}/../tensorflow-core-api/src/gen/java</argument>
+            <argument>${project.basedir}/../tensorflow-core-api/src/gen/resources/org/tensorflow/op/ops.pb</argument>
```
karllessard, line 90: Shouldn't the execution of the generator be added to the core's api POM.xml instead? We can build the generator, and we execute it with this maven plugin, passing it the list of arguments.

karllessard (changes requested):
Instead of reviewing all the code of the generator, is there a dump of the resulting generated ops somewhere where I can compare with the actual version? You can create a temporary draft PR with the update ops in it if you want.

karllessard:
oh also can you do a quick rebase now that the javadoc build [has been fixed](https://github.com/tensorflow/java/pull/279)?

tensorflow-core/tensorflow-core-generator/pom.xml
rnett: I can do that, I just kept them in another folder for seperation.

tensorflow-core/tensorflow-core-generator/pom.xml
rnett: This is as much for manual running/debugging/testing as for the CI.  For the CI, since this is done as part of `build.sh`, I figured I'd just do a `cd ../tensorflow-core-generator && mvn exec`, adding it as part of the POM is complicated (if it's even possible) and unnecessary.

For your top level comment, you can just run the `exec` goal to generate the new ops once you've checked out this branch.  I can make a temp PR with that if you want.

karllessard: The reason why I'd prefer having the exec goal in the core-api pom is that we won't cross anymore module boundaries. You don't need to run this goal from `build.sh`, it can be simply configured to be automatically executed run at the right phase when building the core-api (probably just before we run the annotation processor).

rnett: Ok, this is done, I left the phase commented out for now so it doesn't auto-run, but it is easy to enable.  We should do that in another PR after this one, with CI Build for testing.

rnett:
I made a separate PR with all the generated op changes: https://github.com/rnett/java/pull/3

rnett:
The memory leak test is failing @karllessard @saudet, I didn't change anything around that so I'm 90% sure it's not me.

tensorflow-core/tensorflow-core-api/pom.xml
```diff
@@ -328,6 +328,36 @@
           </dependency>
         </dependencies>
       </plugin>
+      <plugin>
```
karllessard, line 331: That's perfect @rnett , thanks. I'll suggest that you add a comment what this plugin/goal is doing, like the other goals in the same `pom.xml`, it helps to know what is going on since there are a lot of steps involved. 

We can do that when we'll uncomment the phase as well.

karllessard (dismissed):
Good job @rnett and @JimClarke5 ! We can merge that one now (unless @rnett you want to add [that comment](https://github.com/tensorflow/java/pull/244/files#r609110569) before) but before enabling this Java generator, I would like to configure the lint checks (as I was suppose to do in the last two years...) to double-check that the code generated with it is still compliant (like the actual one generated by the C++ generator is).

saudet:
> The memory leak test is failing @karllessard @saudet, it's failing on `master` too.

Which platform?

karllessard:
> > The memory leak test is failing @karllessard @saudet, it's failing on `master` too.
> 
> 
> 
> Which platform?

That's on the latest ubuntu container in GitHub Actions, check config for the quick-build

saudet:
> > > The memory leak test is failing @karllessard @saudet, it's failing on `master` too.
> > 
> > 
> > Which platform?
> 
> That's on the latest ubuntu container in GitHub Actions, check config for the quick-build

Seems a bit flaky on all platforms actually yes. I'll try to make it more reliable...

rnett:
Also, what do you think about only automatically running the generator when `dev` **isn't** enabled?  Since the op def file is only generated during native builds.

karllessard (approved):

karllessard:
> @karllessard yeah I'll add the comment.
> 
> There's a slight issue wrt running the generator during the `core-api` build: it will never build the generator, it will always get the latest version in it's repository (downloaded or maven local). This isn't an issue if you do an `install` in one of the parent projects (which is what I think the CI does), but could trip people up. I think the annotation processor works the same way, so imo it's fine to just document this in the contributors doc since I don't know of a way around it.

Oops, sorry, I’ve noticed your comments just after merging! But that’s fine, it’s the normal way Maven works with interdependencies inside the same project: modules added as dependencies to others needs to be build/installed first and that’s what we do.

> Also, what do you think about only automatically running the generator when dev isn't enabled? Since the op def file is only generated during native builds. The generator itself could be changed, but I would think anyone doing that would run it manually anyways.

I will also leave it like that, it might be counterintuitive that you need to run it manually when you’re in dev mode. Is there any value for not running it?

rnett:
Only speed, really, and not having to worry about any nondeterminism or extra changes (like with the annotation processor).  It's pretty fast, but I'm not sure if it will stay that way if we have to format it.

karllessard:
> Only speed, really, and not having to worry about any nondeterminism or extra changes (like with the annotation processor). It's pretty fast, but I'm not sure if it will stay that way if we have to format it. Probably not worth pulling it out though.

yeah so let’s be reactive and only readjust if needed 👍

PR #247, opened 2021-03-19
Generate op specs and package as resource

rnett:
Extracts the op export part of #244 so that it can make its way into the snapshots.

Adds ~1.5MB to the 70MB native jars, so I don't think the size is an issue.

@karllessard can you add the CI build tag?


rnett:
The CI's having issues w/ `framework`'s javadocs again.

karllessard:
Thanks @rnett , interesting approach, let's give it a try with a full build

rnett:
The build's failing b/c linux-mkl-gpu is running out of disk space.

rnett:
I'm going to move the proto sources in this PR instead of #244, since we already have native builds going.

tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -85,7 +86,17 @@ $BAZEL_BIN/java_op_generator \
...
 
+GEN_RESOURCE_DIR=src/gen/resources/org/tensorflow/op
+mkdir -p $GEN_RESOURCE_DIR
+
+# Generate Java operator wrappers
+$BAZEL_BIN/java_op_exporter \
+    --api_dirs=$BAZEL_SRCS/external/org_tensorflow/tensorflow/core/api_def/base_api,src/bazel/api_def \
+    $TENSORFLOW_LIB > $GEN_RESOURCE_DIR/ops.pb
+
+PROTO_DIR=src/gen/protoc
```
karllessard, line 97: Why changing the directory where the generated Java protos are written? It was fine having .java source files under src/gen/java, no?

rnett: I need to add it as a source directory for `generator`, which I can't do in `src/gen/java` since there's other stuff in there.  There's no way to filter subfolders/packages when attaching it as a source dir.

karllessard: Ok I see. So are we still heading for a two-phase build then?

karllessard: You only need the OpDef and ApiDef proto, right? At this point, I'm wondering if we shouldn't just simply make a one-shot copy of the java protos we need to the generator folder. The protos are under TF stability guarantee and should remain backward-compatible if they happen to change (but we saw what can happen to that guarantee with string tensors...). Anyhow, for the sake of simplifying the whole architecture, it might be worth that the generator has its own hard-copies of these files.

rnett: Shouldn't be two phases, you can run native generation + protos, then generation, then compile core.  The generation will happen in build.sh, right after the proto copy.

What I'm doing is essentially a copy, and we can do it that way, I'm just using maven to attach to the source dir from both projects rather than copying it.

rnett: #244 is set up that way if you want to take a look.  I don't have the actual op gen code in there yet, but it would just be a `cd ../tensorflow-core-generator ; mvn compile exec:java` line at the end of `build.sh`.

karllessard: yeah.. I'm personally not a big fan of this cross-module setup, I'm starting to think why not moving the source code of `tensorflow-core-generator` directly into `tensorflow-core-api` (maybe under `/src/generator/java`?) and build that jar as an additional artifact of `tensorflow-core-api` itself. I guess you've already proposed that but now I see better the reasons for doing it. 

Anyway, we don't use the generator outside the core API and, while initially it was conceived so that users can run it to create their own Ops bindings, it seems that won't be a real use case.

rnett: > Anyway, we don't use the generator outside the core API and, while initially it was conceived so that users can run it to create their own Ops bindings, it seems that won't be a real use case.

I do for the Kotlin API though.  It has it's own generator artifact, which depends on the Java one (just the annotation processor though, not the op gen code).

I'm not a huge fan of crossing the module boundaries like that either, but since it's just protots, we could copy them like you mentioned earlier and just have some duplication.  Or put them in `tensorflow-core` and depend on them from `core-api` and `core-generator`.  It's easier than trying to build multiple jars, and I have no idea if I could get the multiple jars to work with `exec` (which doesn't even need the jars, just `compile`), plus we'd need the generator jar available before the main jar's `compile` phase which I don't think is possible.

karllessard: I've been looking at this and I think you should be fine adding a classifier to [this path](https://github.com/tensorflow/java/blob/5ca958d855307b19f420288f90cbe6dea164411c/tensorflow-core/tensorflow-core-api/pom.xml#L162) in the core API pom. So just to explain a little bit the sequence I'm proposing is:

(0. Move all `tensorflow-core-generator` sources to `tensorflow-core-api/src/generator`)
1. Right after the native build (or after we've downloaded the native JARs in dev mode), compile the generator sources by adding a goal to the `maven-compiler-plugin` 
2. Package the generator jar by adding a goal to the `maven-jar-plugin`
3. You might need to install prematurely the generator into your local maven repository before executing the `default-compile`, that I don't know...
4. Continue Maven build as usual

Please let me know what you think about this

rnett: The only missing step is actually running the (non-annotation processor) generator, which needs to happen before `default-compile`.  I'm also using the `exec` plugin to run it, and I have no idea if that supports classifiers.

It seems possible, and I'll try it, but it's a lot of build cruft to avoid just making a copy of the protos (which would live in git anyways), I'm not sure it's worth it.  Plus, it would be easy to cause errors by depending on other core api sources in the generator, and I'm not sure Intellij (or other IDEs) would be smart enough to warn about that.

karllessard: > Plus, it would be easy to cause errors by depending on other core api sources

Actually why not leveraging that and let the generator use the JavaCPP C API bindings so it can retrieve the list of ops from the native library instead of relying on a generated `ops.pbtxt` file?

We wouldn't have to maintain C++ code anymore.

rnett: Then we're back to essentially a two phase build: building native, building generator (which depends on everything in core) without the annotation processor, generating the ops, then building core.  It's probably possible, but I don't see why it's worth it, when the alternative is just copying the protos.  

We'd have c++ code anyways for the ApiDef importing, and the export main is only ~100 lines.  And it makes an already messy maven build much worse.  Plus there's other ops that get added at runtime that end up in the OpDefs when calling from Java that aren't there when called from the native build (and which depend on the OS).  Not necessarily a bad thing, but they are undocumented and probably unstable, and not something we should depend on without marking it.

karllessard: > building generator (which depends on everything in core) 

I was about to tell you that you a wrong because the C API bindings are hermetic to the rest of the core's code but now thinking of it, some of the abstract classes extends from those found under `src/main/java/org/tensorflow/internal/c_api`... so yeah, that can be a problem (do these abstract classes depend on anything else in the core though?).

As for the ApiDef importer, this is a secondary utility that should not influence the decisions we are taking for designing the main architecture.

Finally about the "missing ops" in some native libraries, I'm also not sure what is best to do with them. Which platform do we elect to be responsible for generating the `op.pbtxt`, isn't the problem the same?

rnett: > Yeah, for the complexity of the Maven build I understand your point but I'm just trying to see if we can avoid crossing module boundaries here, which sounds even worst.

Agreed, but the boundary crossing is only necessary if we don't copy the protos.  It's easy enough to just make two copies, one in core and one in generator, like you suggested earlier.  They'll both be in git anyways and have tensorflow stability gaurentees.  I think that's much nicer than the build stuff.

> do these abstract classes depend on anything else in the core though?

Not that I know of, but I meant it more in the "the IDE won't be smart enough to stop you from using Graph in the generator" sense.  Regardless of whether we allow using the bindings or not that will be an issue, I think.

> Finally about the "missing ops" in some native libraries, I'm also not sure what is best to do with them. Which platform do we elect to be responsible for generating the op.pbtxt, isn't the problem the same?

I generated the one we have on Linux, but I don't think that problem exists for the native exporter (although I can't get Windows native builds working to check).  If it is a problem it's one that exists now, and since we haven't ran into anything I assumed things were fine.  Most if not all (I didn't check) of the additional ops that were present in the loaded library started with `_` and I think they were added at runtime.  Things like `_Arg` or `_XlaCompile` that I think are supposed to be "private" ops anyways (they aren't in tensorflow's `ops.pbtxt`).

It would be good to check out ops.pb against tensorflow's for the op defs at least.  It should be possible to do that in the generator and fail if they don't match, it's just a pretty big download.  Maybe better to make a separate `main` for it.


Also, the whole reason core-api runs it's tests in `integration-test` instead of `test` is that the native libraries aren't available for `test` (I think), so I'm not actually sure if we could get them before `compile`.

karllessard: > It's easy enough to just make two copies, one in core and one in generator, like you suggested earlier. 

Ok I agree, let's go back with this idea and play simple. 

Just to recapitulate and make sure we've understood each other: while the protos in`tensorflow-core-api` are generated at build, I was suggesting doing only a static copy of the protos we need in the generator, presuming that they won't change (much) anyway.

rnett: 👍 works for me.  The protos used for exporting the ops from C++ are from the native tensorflow version, so worst case scenario we get errors during native build if something changes.

saudet: BTW, anything that gets executed after the JavaCPP plugin executions has access to its classes and all its resources, that is to say the native libraries, in other words, we have access to the C API from Java at that point. That's just how Maven works. Anything that gets outputted by a plugin automatically gets added to the class path of the following plugins to be executed.

karllessard (approved):
Thanks @rnett , let's see how it goes with a full CI build but without waiting, I'll merge it now and adjust if needed

PR #250, opened 2021-03-21
Javadoc framework fixes

JimClarke5:
This PR cleans up the JavaDoc in the `org.tensorflow.framework.data`  package. Many classes and methods had JavaDoc missing and this was causing errors during the javadoc build phase. As a few classes were missing JavaDoc all together, please double-check that my JavaDoc is accurate.

Also, fixed a missing `@param <T>` error in the `org.tensorflow.framework.constraint.Constraint` class

No code was changed in the PR, just JavaDoc addition and fixes.


rnett:
Looks like you're still getting build issues because the core docs aren't built.  I've said this elsewhere, but imo the best solution for now is to enable ignoring javadoc failures in framework (like core) until we can update the markdown converter.

JimClarke5:
FYI I have the Java version of the markdown converter done. I just need to wire it into the op generator when that is done.

Craigacp:
Should we temporarily turn off doclint on the core docs? https://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

karllessard:
One thing I really don't understand is why we cannot reproduce this error locally? Running a `mvn install -Pdev` locally is successful, and looking at the error it sounds like the issue is that the `tensorflow-core-api` JAR is not found when validating the framework javadoc. This is all very confusing...

karllessard:
I'm relaunching the [initial workflow](https://github.com/tensorflow/java/actions/runs/654488253) for deploying a first 0.4.0-SNAPSHOT to Sonatype, looks like it was kind of "hanging" there without being completed. I have no clue if that can fix anything but such obscure days, anything is possible...

JimClarke5:
I have had some issues on rebuilding locally (`mvn install`) after a` mvn clean`. One or more times, I have had to do the `mvn clean` again, finally it takes. Usually, when this happens I get errors complaining about missing `Op` classes.

karllessard:
OK @JimClarke5 I did a clean but only in the `framework`, let me try with a full clean then...

rnett:
Make sure you don't have a `-Dmaven.javadoc.skip=true` for speed reasons like I did.  Without that though, I can reproduce it locally w/ `mvn clean install -Pdev -Pjdk11` ( I don't think the `jdk11` is necessary).

Also, are we even publishing javadocs for core-api?  I don't see any artifacts in https://oss.sonatype.org/content/repositories/snapshots/org/tensorflow/tensorflow-core-api/0.4.0-SNAPSHOT/ for them.

And https://oss.sonatype.org/content/repositories/snapshots/org/tensorflow/tensorflow-framework/0.4.0-SNAPSHOT/ is empty, although I suppose that could be because you restarted the action.

PR #253, opened 2021-03-26
Fix memory leak occuring with new layout of string tensors

### LINKED ISSUE #251: Possible memory leak with version 0.3.0
When upgrading from version 0.2.0 to 0.3.0 we are seeing a possible memory leak. I have attached a small sample project that demonstrates the issue. It has a unit test that has an infinite loop but which should be properly closing the resources it uses. If you run `mvn clean install` and observe the memory usage of the process, you should observe that the process continues to consume more and more memory. We did not observe this behavior with this code using version 0.2.0.
[example.zip](https://github.com/tensorflow/java/files/6193420/example.zip)

saudet:
Also fix and actually use JavaCPP deallocator for TF_Session as well

Fixes #251 and we should probably add a regression test for that somewhere.
What would be a good place?

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
@@ -604,23 +604,16 @@ public final class Session implements AutoCloseable {
         status.throwExceptionIfNotOK();
       }
 
-      TF_Session session = TF_NewSession(graphHandle, opts, status);
+      TF_Session session = TF_Session.newSession(graphHandle, opts, status);
       status.throwExceptionIfNotOK();
 
-      return session;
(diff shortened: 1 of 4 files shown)


karllessard:
Thanks @saudet , can you please rebase your changes on the `r0.3` branch instead? We'll merge it in that branch to quickly release a version 0.3.1 with these fixes.

For the regression test, what about [here](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/types/TStringTest.java)?

saudet:
I've changed the base to r0.3 and it seems like it can still merge this, so go ahead!

karllessard (approved):
All right looks good @saudet , thanks! I'll merge it when I'll be ready to release 0.3.1

Issue #260, opened 2021-03-30
Exception in thread "main" java.lang.IllegalArgumentException: No tensor type has been registered for data type DT_RESOURCE


micosacak:
Now I had another problem. I had the following issue (https://github.com/tensorflow/java/issues/258) which have been solved. if I use float image;

`Exception in thread "main" org.tensorflow.exceptions.TFInvalidArgumentException: Expects arg[0] to be double but float is provided
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:87)
	at org.tensorflow.Session.run(Session.java:691)
	at org.tensorflow.Session.access$100(Session.java:72)
	at org.tensorflow.Session$Runner.runHelper(Session.java:381)
	at org.tensorflow.Session$Runner.run(Session.java:329)
	at TfDataTypeIssue.main(TfDataTypeIssue.java:45)`

then I converted all buffered image to ´TFloat64´ which is double now I get the following error:

`Exception in thread "main" java.lang.IllegalArgumentException: No tensor type has been registered for data type DT_RESOURCE
	at org.tensorflow.internal.types.registry.TensorTypeRegistry.find(TensorTypeRegistry.java:50)
	at org.tensorflow.RawTensor.fromHandle(RawTensor.java:147)
	at org.tensorflow.Session.run(Session.java:695)
	at org.tensorflow.Session.access$100(Session.java:72)
	at org.tensorflow.Session$Runner.runHelper(Session.java:381)
	at org.tensorflow.Session$Runner.run(Session.java:329)
	at TfDataTypeIssue.main(TfDataTypeIssue.java:45)`

here is the part of my Java code;

`SavedModelBundle theModel = SavedModelBundle.load("CreateModelWithPythonTensorflow/model", "serve");`
´double[][][][] floatTensor = getTensoredImage("test_images/test_00001.png");´
´TFloat64 tensorImage = TFloat64.tensorOf(StdArrays.ndCopyOf(floatTensor));´
        
`Tensor res = sess.runner().feed("serving_default_inputTensor_input", tensorImage).fetch("outputTensor/kernel").run().get(0);`

or

`TFloat64 res = (TFloat64) sess.runner().feed("serving_default_inputTensor_input", tensorImage).fetch("outputTensor/kernel").run().get(0);`

rnett:
[...]

PR #261, opened 2021-03-30
Add fetchVariable method to Session to get value of resource variable

rnett:
Small PR to add a method to Session to allow for the fetching of the values of resource variables.  Fetches the resource tensor then uses an eager session to read the value.


rnett:
CI errors are from Javadoc, tests are passing.

rnett:
Users can tell if their operand is a variable by checking the data type, maybe we should add a `Operand.isResourceVariable()`?  But yeah, since we don't have a tensor type for resources there's no good way to make it typesafe.  Adding it is a possibility, and probably good to do at some point, but beyond me at the moment (it's not mappable to NDArrays, so it doesn't fit well with the existing tensor type setup).

Fetching a variable is almost always an error anyways, since it's not usually an output, especially if you don't realize what you're fetching is a variable (i.e. #260).  This is intended for the special use cases where you want to get a variable, in which case you'll know ahead of time it's a variable.

To read a value from a resource tensor, you need to use the `readVariableOp` Op, either in the graph or eager session.  I could add the ops to the graph and fetch them instead, using reads if they are already present, but that adds extra ops to the graph and it's not easy to detect if you already have a read, unless we limit it to ones added by this (attributes need to match too (i.e. not being on another device) and we don't have a way to check those).  It's worth considering, but creating eager sessions doesn't seem heavy, so I went with that.

Craigacp:
I think fetching variables is entirely reasonable as you can fetch everything else in the graph by name. Having a set of things that you can't fetch and the only way to find out is by trying it and catching a runtime exception is bad.

I'd expect TF python to add `readVariableOp` during variable creation or when the graph is saved, so shouldn't those ops exist in any graph that has resource tensors?

rnett:
> I think fetching variables is entirely reasonable as you can fetch everything else in the graph by name. Having a set of things that you can't fetch and the only way to find out is by trying it and catching a runtime exception is bad.

Agreed, but I don't think it's possible, at least for now.  The reason I did this in a separate method is because we can't fetch variables normally.  There's no Resource TType, so tensor creation fails.  Even if we did add one, it doesn't fit with the existing TType setup as the only method we would want to provide is `read(Class<? extends TType>)`.  We can't even auto-read variables in `fetch` since they don't specify their value's dtype (at least that I can find), so it needs the class parameter.

> I'd expect TF python to add readVariableOp during variable creation or when the graph is saved, so shouldn't those ops exist in any graph that has resource tensors?

I think it does, but it's possible for people to use the raw op rather than `tf.Variable`.  Checking for a read op is probably the better route though.

Craigacp:
> > I think fetching variables is entirely reasonable as you can fetch everything else in the graph by name. Having a set of things that you can't fetch and the only way to find out is by trying it and catching a runtime exception is bad.
> 
> Agreed, but I don't think it's possible, at least for now. The reason I did this in a separate method is because we can't fetch variables normally. There's no Resource TType, so tensor creation fails. Even if we did add one, it doesn't fit with the existing TType setup as the only method we would want to provide is `read(Class<? extends TType>)`. We can't even auto-read variables in `fetch` since they don't specify their value's dtype (at least that I can find), so it needs the class parameter.
> 
> > I'd expect TF python to add readVariableOp during variable creation or when the graph is saved, so shouldn't those ops exist in any graph that has resource tensors?
> 
> I think it does, but it's possible for people to use the raw op rather than `tf.Variable`. Checking for a read op is probably the better route though.

Why not have it check for the read op and then silently redirect the fetch to that read op. That will give us back a non-resource tensor right? Otherwise unpicking the resource and exporting it into a regular tensor seems fine too. I'm wary of spinning up and throwing away eager sessions to access a variable. There's a bunch of stuff that happens in [TFE_NewContext](https://github.com/tensorflow/tensorflow/blob/f6079ec87f01feaa9cc7a5134a8064d05944ea2a/tensorflow/c/eager/c_api.cc#L115).

rnett:
It looks like it is possible to get the variable dtype from the attribute, so I can do everything using graph based ops, and automatically.  However, eventually, once we have support for resource tensors, fetching a resource tensor will be a legitimate operation.  So there still needs to be some way to differentiate between fetch w/ conversion and fetch without.  Maybe make fetch auto-convert for now, and once we have resource tensor support add `fetchVariable` that gets the resource tensor?

rnett:
Ok, it automatically wraps variables in a read in the graph in `fetch` now.

karllessard:
@rnett, did you try simply using the [`TF_TensorData`](https://github.com/tensorflow/java/blob/2e8f00c2f95fee75f38614ccacb665f0eb5f3250/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/global/tensorflow.java#L711) endpoint of the C API to get a pointer to the variable native buffer directly? That's what the [1.x client](https://github.com/tensorflow/tensorflow/blob/04f05ec9b031bad26e858d70cb90674b1b59dc26/tensorflow/java/src/main/native/tensor_jni.cc#L500) was using before, making a copy first before returning it though. And I'm not sure if that was supporting variables but it could be interesting to try?

rnett:
I didn't try that, but I assume the 1.x client was using the old variables, not resource variables.  I'm not sure how to get a tensor from that buffer, either.

rnett:
> But anyway, if Python does it with the `readVariableOp`, then let's do that as well

👍 They are already added to the graph during the variable's creation in Python, and I plan on doing something similar for our version, so it should be fine.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -225,24 +235,59 @@ public Runner fetch(String operation) {
...
+          TF_OperationGetAttrType(graphOp.getUnsafeNativeHandle(), "dtype", rawDt, status);
+          status.throwExceptionIfNotOK();
+        }
+
+        DataType valueDt = DataType.forNumber(rawDt[0]);
+
+        Operand<?> read = null;
+        for (GraphOperation op : graphOp.consumers()) {
+          if (op.dtype(0) == valueDt && op.type().equals(ReadVariableOp.OP_NAME)) {
+            read = op.output(0);
```
Craigacp, line 270: Should it break out of this loop when the right op is found?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -225,24 +235,59 @@ public Runner fetch(String operation) {
...
+          }
+        }
+
+        if (read == null) {
+          read = Ops.create(graph).withSubScope("session_reads").withName(output.op().name() + "_read")
```
Craigacp, line 275: Might be nice to have some logging here that we can turn on to see if it's being hit a lot (as that means our assumptions about graphs are being violated). Maybe at `logger.debug` level?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -269,13 +312,11 @@ public Runner addTarget(String operation) {
...
+     * @throws IllegalStateException if the operation is not from the session's graph.
      */
     public Runner addTarget(Operation operation) {
-      if (!(operation instanceof GraphOperation)) {
-        throw new IllegalArgumentException(
-            "Operation of type "
-                + operation.getClass().getName()
-                + " is not supported in graph sessions");
+      if (operation.env() != graph) {
+        throw new IllegalStateException("Can't fetch operation " + operation + ", it is from a different graph.");
```
Craigacp, line 319: This exception message is a little different, and could be confusing when debugging if you pass in an eager op somehow as it won't come from a graph and the message might lead you up the wrong path. Maybe "it is either an eager operation or from a different graph"? Similarly for the other places with this exception message that accept an `Operation`.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/SessionTest.java
```diff
@@ -244,9 +249,54 @@ public void saveAndRestore() throws IOException  {
...
-            .sorted(Comparator.reverseOrder())
-            .map(Path::toFile)
-            .forEach(File::delete);
+        .sorted(Comparator.reverseOrder())
+        .map(Path::toFile)
+        .forEach(File::delete);
+  }
+
+  @Test
+  public static void testFetchVariable() {
```
Craigacp, line 258: Might be an idea to have a test that loads in a GraphDef from python which has a resource variable in it. That way we're definitely testing the most common way to exercise that codepath.

Craigacp (changes requested):
A couple of small things about logging/messages, and I think a test loading a python graph is important to have.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
rnett: I don't think we have a logging library at the moment, do we?  But yeah, it would be nice.

Craigacp: Oh yeah, probably not. I tend to use `java.util.logging` for everything but I understand that most people use `slf4j`.

rnett: There's been a couple times that having one would be nice, we should decide on something.  Next core meeting maybe?

Craigacp (approved):
LGTM

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
Craigacp: Yeah, let's discuss it then.

PR #266, opened 2021-03-31
Add regression test for memory leaks in string tensors

saudet:
From #251

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/types/TStringTest.java
+++ b/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/types/TStringTest.java
@@ -20,8 +20,10 @@ package org.tensorflow.types;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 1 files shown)


karllessard (commented):
Thanks @saudet , would you mind explaining quickly what the test is measuring (e.g. where comes the 10_000_000 and why running GC if we are closing explicitly the tensors?)

It could be by adding some comments in the test?

saudet:
It's just some arbitrary value that's probably enough to account for the overhead of the JVM and stuff that it does in the background. There isn't much to explain! I just came up with something that seems to work well enough for a unit test. If you feel that we shouldn't put tests that are not 100% in there, well, let's not merge this. This is the best I could come up with without spending the whole day on it.

saudet:
It looks like there's supposed to be a "CppMemoryChecker" that tracks allocations as per this file in the Python API:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/memory_checker.py#L27
However, that module is nowhere to be found. I assume this is something that Google hasn't open sourced...
The only thing available seems to be the "PythonMemoryChecker" that works very much like JavaCPP and is only able to track objects that are registered with its GC:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/python_memory_checker.py#L44-L50

karllessard (approved):

Issue #268, opened 2021-04-01
Ability to convert Tensor to String representation


cowwoc:
Per our discussion on Gitter, here is a possible implementation for converting Tensors to a String representation. It is still missing some important features, like collapsing long arrays using ellipses, but this can serve as a stepping stone. The functionality is meant to ease troubleshooting/debugging so performance should not be an issue.

```
import org.tensorflow.Session;
import org.tensorflow.ndarray.Shape;
import org.tensorflow.ndarray.buffer.DataBuffer;
import org.tensorflow.ndarray.buffer.DoubleDataBuffer;
import org.tensorflow.ndarray.buffer.FloatDataBuffer;
import org.tensorflow.ndarray.buffer.IntDataBuffer;
import org.tensorflow.ndarray.buffer.LongDataBuffer;
import org.tensorflow.ndarray.buffer.ShortDataBuffer;
import org.tensorflow.types.TFloat16;
import org.tensorflow.types.TFloat32;
import org.tensorflow.types.TFloat64;
import org.tensorflow.types.TInt32;
import org.tensorflow.types.TInt64;
import org.tensorflow.types.TUint8;

import java.util.StringJoiner;

public final class Tensors
{
	private final Session session;

	/**
	 * @param session the session used by all operations
	 */
	public Tensors(Session session)
	{
		this.session = session;
	}

	/**
	 * @param tensor a tensor
	 * @return the String representation of the tensor
	 */
	public String toString(TFloat64 tensor)
	{
		Shape shape = tensor.shape();
		DoubleDataBuffer doubles = tensor.asRawTensor().data().asDoubles();
		return toString(doubles, shape, 0, 0, tensor.rank()).text;
	}

	/**
	 * @param tensor a tensor
	 * @return the String representation of the tensor
	 */
	public String toString(TFloat32 tensor)
	{
		Shape shape = tensor.shape();
		FloatDataBuffer doubles = tensor.asRawTensor().data().asFloats();
		return toString(doubles, shape, 0, 0, tensor.rank()).text;
	}

	/**
	 * @param tensor a tensor
	 * @return the String representation of the tensor
	 */
	public String toString(TFloat16 tensor)
	{
		Shape shape = tensor.shape();
		FloatDataBuffer doubles = tensor.asRawTensor().data().asFloats();
		return toString(doubles, shape, 0, 0, tensor.rank()).text;
	}

	/**
	 * @param tensor a tensor
	 * @return the String representation of the tensor
	 */
[...]

PR #272, opened 2021-04-03
Added Tensor.dataToString()

cowwoc:


cowwoc:
@googlebot I signed it!

cowwoc:
@rnett Please review.

Apologies in advance. The implementation got hairy once I added `maxWidth` support. Please refactor this as you see fit to improve readability.

cowwoc:
The build failure doesn't seem to be related to my changes. Please confirm.

Craigacp:
The quick build failing in javadoc generation is a known issue we're working on in other PRs.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * @param tensor   a tensor
+   * @param maxWidth the maximum width of the output in characters ({@code null} if unlimited). This
+   *                 limit may surpassed if the first or last element are too long.
+   * @return the String representation of the tensor
+   */
+  public static String toString(Tensor tensor, Integer maxWidth) {
+    if (!(tensor instanceof NdArray)) {
```
rnett, line 39: This will error on inputs like RawTensor, which should be supported.  Maybe take a look at `RawTensor.asTypedTensor`.  You'd have to move packages, but it seems like it would be useful.  @karllessard are there better ways to convert a tensor to an ndarray?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
```diff
@@ -0,0 +1,158 @@
...
+      String element = iterator.next().getObject().toString();
+      joiner.add(element);
+      int addedLength = joiner.length() - lengthBefore;
+      lengths.add(addedLength);
+      lengthBefore += addedLength;
+    }
+    return truncateWidth(joiner.toString(), maxWidth, lengths);
+  }
+
+  /**
```
rnett, line 94: Can you add a basic description of what this does?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * @param tensor a tensor
+   * @return the String representation of the tensor
+   */
+  public static String toString(Tensor tensor) {
+    return toString(tensor, null);
+  }
+
+  /**
+   * @param tensor   a tensor
```
rnett, line 33: Can you add some description?  I.e. how will maxWidth change the output?  I assume it controls the use of the `..`.s?

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/TensorTest.java
```diff
@@ -541,6 +541,33 @@ public void gracefullyFailCreationFromNullArrayForStringTensor() {
     }
   }
 
+  @Test
+  public void dataToString() {
```
rnett, line 545: Add a multidimensional test maybe?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -212,4 +246,23 @@
    */
   @Override
   void close();
+
+  public static class ToStringOptions {
```
rnett, line 250: What do you think about adding a float format option?  Either just the decimal places or a `String.format`/printf` style format.  Maybe use `String.format` internally?

rnett (changes requested):
Did you consider adding an `Operand.dataToString()` default method that calls `asTensor().dataToString()` (or similar)?

cowwoc:
> Looks good, I have a few documentation requests and a request for float formatting. Did you consider adding an `Operand.dataToString()` default method that calls `asTensor().dataToString()` (or similar)?

Yes, I could do that. Are you saying that you would leave the main implementation on `Tensor` and simply add a convenience method on `Operand` that redirects to it?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
cowwoc: I'm confused. The `truncateWidth()` methods is already documented. What did you mean?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
cowwoc: Good idea, but I'm going to leave this for someone else to add as a future PR.

cowwoc:
FYI, I decided to remove trailing commas after closing brackets. I think it looks better this way and I think that `tf.print` does the same. In order words:

```
[
  [1, 2],
  [3, 4],
  [5, 6]
]
```

Will now show up as:

```
[
  [1, 2]
  [3, 4]
  [5, 6]
]
```

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/TensorTest.java
cowwoc: Done. Please review.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
cowwoc: I've added a description. Please review.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
cowwoc: I've added a description. Please review.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/Tensors.java
cowwoc: I think I've now added support for `RawTensor`. Please review.

rnett:
> > Looks good, I have a few documentation requests and a request for float formatting. Did you consider adding an `Operand.dataToString()` default method that calls `asTensor().dataToString()` (or similar)?
> 
> Yes, I could do that. Are you saying that you would leave the main implementation on `Tensor` and simply add a convenience method on `Operand` that redirects to it?

Yeah, that's exactly what I ment.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -30,12 +30,20 @@ public static String toString(Tensor tensor) {
...
    *                 limit may surpassed if the first or last element are too long.
    * @return the String representation of the tensor
    */
   public static String toString(Tensor tensor, Integer maxWidth) {
+    if (tensor instanceof RawTensor) {
+      System.out.println("Got rawTensor: " + tensor);
+      tensor = ((RawTensor) tensor).asTypedTensor();
+    }
     if (!(tensor instanceof NdArray)) {
       throw new AssertionError("Expected tensor to extend NdArray");
```
rnett, line 48: Can you make this message more descriptive?  Mention the tensor's dtype, the typed class (i.e. `tensor.getClass()`), and that it isn't mapped to an ndarray and thus can't be printed.  Maybe mention that you can use `tf.print`, too.

rnett (changes requested):
Other than the small comments, I made a PR against your branch with additional data type tests, and (because of said tests) wrapping string elements in quotes.  Use the PR if you want, but regardless, both of those would be good to have.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -30,12 +30,20 @@ public static String toString(Tensor tensor) {
...
+   * data.
+   *
    * @param tensor   a tensor
    * @param maxWidth the maximum width of the output in characters ({@code null} if unlimited). This
    *                 limit may surpassed if the first or last element are too long.
    * @return the String representation of the tensor
    */
   public static String toString(Tensor tensor, Integer maxWidth) {
+    if (tensor instanceof RawTensor) {
+      System.out.println("Got rawTensor: " + tensor);
```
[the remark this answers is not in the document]
saudet: This might be a good opportunity to decide which framework we want to use, if not SLF4J itself, Logback I suppose. @karllessard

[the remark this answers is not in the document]
rnett: Yeah, I ran into this in another PR as well, I plan on bringing it up at the next core meeting.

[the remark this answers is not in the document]
cowwoc: For what it's worth, I simply forgot to remove this line before committing :) I'll clean this up now.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
cowwoc: I will add more debugging information but keep in mind that this error is not meant to be user facing (hence `AssertionError` versus a normal exception). If we see this error it is because someone added a new type of tensor and forgot to update `Tensors` to support it. This corresponds to an internal error on our part. At least, that was my intent.

cowwoc:
> Other than the small comments, I made a PR ([cowwoc#1](https://github.com/cowwoc/java/pull/1)) against your branch with additional data type tests, and (because of said tests) wrapping string elements in quotes. Use the PR if you want, but regardless, both of those would be good to have.

Merged. Thank you.

Let me know if there is anything else. Otherwise, how do we fix the quick-build so I can merge this PR?

Craigacp:
The test failure will be fixed in #276. The Javadoc failures we're still working on. Once #276 is merged in then I'll rerun the action and check that the tests complete.

rnett:
@googlebot I consent.

rnett (approved):
LGTM

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -26,9 +26,9 @@
  * A statically typed multi-dimensional array.
  *
  * <p>There are two categories of tensors in TensorFlow Java: {@link TType typed tensors} and
- * {@link RawTensor raw tensors}. The former maps the tensor native memory to an
- * n-dimensional typed data space, allowing direct I/O operations from the JVM, while the latter
- * is only a reference to a native tensor allowing basic operations and flat data access.</p>
+ * {@link RawTensor raw tensors}. The former maps the tensor native memory to an n-dimensional typed
+ * data space, allowing direct I/O operations from the JVM, while the latter is only a reference to
+ * a native tensor allowing basic operations and flat data access.</p>
```
karllessard, line 31: There is a lot of (undesirable?) format change in this PR, can you please revert those and only preserve changes related to the `dataToString` feature?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -191,6 +196,33 @@
...
+      }
+    }
+    return Tensors.toString(this, maxWidth);
+  }
+
+  /**
+   * @param maxWidth the maximum width of the output in characters ({@code null} if unlimited). This
+   *                 limit may surpassed if the first or last element are too long.
+   */
+  static ToStringOptions maxWidth(Integer maxWidth) {
```
karllessard, line 222: I'm not sure this method is in the right place. Maybe move it in `ToStringOptions`? Also you need to describe the returned value in the javadoc or the checks might complain.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
```diff
@@ -191,6 +196,33 @@
...
   long numBytes();
 
+  /**
+   * Returns the String representation of elements stored in the tensor.
+   *
+   * @param options overrides the default configuration
+   * @return the String representation of the tensor elements
+   * @throws IllegalStateException if this is an operand of a graph
+   */
+  default String dataToString(ToStringOptions... options) {
```
karllessard, line 206: I don't think the use of a vararg to handle the optional presence of `options` is a good idea. Having a second method accepting no parameter would be better. 

We use vararg options in the op wrappers because we want to limit the number of methods that ending up in the `*Ops` classes, which is already more than a thousand. But here it's fine "duplicating" it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -0,0 +1,186 @@
...
+   * Returns a String representation of a tensor's data. If the output is wider than {@code
+   * maxWidth} characters, it is truncated and "{@code ...}" is inserted in place of the removed
+   * data.
+   *
+   * @param tensor   a tensor
+   * @param maxWidth the maximum width of the output in characters ({@code null} if unlimited). This
+   *                 limit may surpassed if the first or last element are too long.
+   * @return the String representation of the tensor
+   */
+  public static String toString(Tensor tensor, Integer maxWidth) {
```
karllessard, line 42: I'm not sure about having that many ways to print a tensors (if we can call `tensor.toString`, why having `Tensors.toString(tensor)`?) It's a bit confusing and I find that that feature maybe spreads out a bit in multiple parts of the code.

I have a new design to propose, let me know what you think. But what about having this `toString` logic in another called, let say, `TensorPrinter`, which can be returned or directly invoked in the `Tensor` class? i.e. 

```java
class TensorPrinter {

    TensorPrinter(Tensor tensor, int maxWidth) { ... }

    TensorPrinter withMaxWidth(int maxWidth) {
        return new TensorPrinter(this.tensor, maxWidth);
    }

    String print() {
        return .... (this logic here)
    }
}

interface Tensor {

    String print() {
        return new TensorPrinter(this, null).print();
    }

    TensorPrinter printer() {
        return new TensorPrinter(this, null);
    }
}

Tensor t = TFloat32.scalarOf(10.0f);
t.print();
t.printer().maxWidth(10).anotherOption(234).print();
```

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
```diff
@@ -0,0 +1,186 @@
...
+    if (tensor instanceof RawTensor) {
+      tensor = ((RawTensor) tensor).asTypedTensor();
+    }
+    if (!(tensor instanceof NdArray)) {
+      throw new AssertionError("Expected tensor to extend NdArray.\n" +
+          "actual  : " + tensor + "\n" +
+          "dataType: " + tensor.dataType() + "\n" +
+          "class   : " + tensor.getClass());
+    }
```
karllessard, line 51: Since you only accept typed tensors for this method, the endpoint should probably be added to `TType` instead of `Tensor`.

karllessard (commented):
Thanks a lot @cowwoc , I'm not done with the review but I'll wait to see you have to say about my comments so far.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
cowwoc: I prefer your proposal. It is exactly what I was hoping to implement in the long term but I ended up pushing what you see above as a stepping stone in the right direction.

Also, consider the relationship (if any) between `Ops.print()` and this functionality. I may be wrong, but I believe the C++ definition of `tf.print()` maps to what I'm trying to implement in this PR. Does it make sense to have `Ops.print()` delegate to this as well?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
cowwoc: @karllessard Is there a way for me to handle all tensors (not just `TType`)? It's okay if not. I'm just asking.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
JimClarke5: I agree with @karllessard

JimClarke5 (commented):
I have quite a bit of Tensor print helper classes that I use in `test/java/org.tensorflow.framework.utils.TestSession`, `EagerTestSession`, and `GraphTestSession`.

I would like to use this PR in those classes to make sure nothing is missed.
I'll test that and let you know how it goes.

cowwoc:
@JimClarke5 Any chance you could pick up the torch (apply the changes that Karl asked for) on my behalf? I won't have time to pick this up for another week or two.

JimClarke5:
@cowwoc OK, one problem I have is I need to work with my latest version of the xxxSession classes that I am using for Models and Layers, I assume you could cherry pick the Tensor related changes.

cowwoc:
> @cowwoc OK, one problem I have is I need to work with my latest version of the xxxSession classes that I am using for Models and Layers, I assume you could cherry pick the Tensor related changes.

You should be able to fork my PR, merge in your changes and push a new PR. I will approve it which should merge the changes back into this initial PR.

JimClarke5:
I understand that, but I need to test with my latest code first.

JimClarke5:
Will `Operand.dataToString()` work in Graph mode with out first doing a try catch?

```
try (Tensor tensor = session.runner().fetch(input).run().get(0)) {
     tensor.dataToString()
}
```

JimClarke5:
Would we want the printed values enclosed in braces `( {} )` rather than square brackets` ( [] `}. Braces make it easier to copy and past into Java code.

JimClarke5:
I think we should pattern the output after` java.util.Arrays.deepToString()`, which takes us back to brackets with commas.

```
float[][][] f = new float[][][] {
            {{1,2}, {3,4}},
            {{5, 6}, {7, 8}},
            {{9, 10}, {11, 12}}
};   
System.out.println(Arrays.deepToString(f));
// [[[1.0, 2.0], [3.0, 4.0]], [[5.0, 6.0], [7.0, 8.0]], [[9.0, 10.0], [11.0, 12.0]]]
```

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensor.java
JimClarke5: I do it similar to this:
`Layers.Options.create().inputShape(Shape.of(2,2))`
without the ellipsis in the CTORs.

cowwoc:
> Will `Operand.dataToString()` work in Graph mode with out first doing a try catch?
> 
> ```
> try (Tensor tensor = session.runner().fetch(input).run().get(0)) {
>      tensor.dataToString()
> }
> ```

Per https://github.com/tensorflow/java/issues/268#issuecomment-811663700 I was asked to only support eager mode. I suggest discussing this part with @rnett.

> I think we should pattern the output after` java.util.Arrays.deepToString()`, which takes us back to brackets with commas.
> 
> ```
> float[][][] f = new float[][][] {
>             {{1,2}, {3,4}},
>             {{5, 6}, {7, 8}},
>             {{9, 10}, {11, 12}}
> };   
> System.out.println(Arrays.deepToString(f));
> // [[[1.0, 2.0], [3.0, 4.0]], [[5.0, 6.0], [7.0, 8.0]], [[9.0, 10.0], [11.0, 12.0]]]
> ```

My 2 cents: Make the default output human-friendly (i.e. square brackets as `TF.print()` and others do) and provide a Builder option that will request Java-style output. I find the indentation I provided above easier to read (for humans) than Java code, but I agree that the latter is useful functionality.

rnett:
> Per [#268 (comment)](https://github.com/tensorflow/java/issues/268#issuecomment-811663700) I was asked to only support eager mode. I suggest discussing this part with @rnett.

That's for Operands, `Tensor`s should work fine regardless (which is what was in Jim's example).

> My 2 cents: Make the default output human-friendly (i.e. square brackets as `TF.print()` and others do) and provide a Builder option that will request Java-style output. I find the indentation I provided above easier to read (for humans) than Java code, but I agree that the latter is useful functionality.

I'd agree with this: default to `tf.print` style but have options.  I don't find the default Java style to be very readable for large tensors.

JimClarke5:
@rnett @cowwoc FYI, in Graph mode, accessing` Operand.asTensor()` will fail without calling the session `fetch` first.

rnett:
> @rnett @cowwoc FYI, in Graph mode, accessing` Operand.asTensor()` will fail without calling the session `fetch` first.

Yeah, that's why we didn't have `dataToString()` do anything in Graph mode.  It just calls `asTensor().dataToString()`, so it will do the same thing as `asTensor()`.

JimClarke5:
@cowwoc Your branch is behind master and that is causing me build issues.

cowwoc:
@JimClarke5 One sec, I'll merge master into it.

cowwoc:
@JimClarke5 Try now.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
JimClarke5: Do we want to call it `print()` and not `toString()`?

cowwoc: If we do that, then I would then expect it to work the same as `Ops.print()` since the two share the same name. Are we planning to merge the two?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
JimClarke5: @karllessard  How do you create a `RawTensor` that doesn't inherit from `NdArray`? Or more precisely how do you create a plain `Tensor` object that doesn't inherit from `NdArray` without using one of the `org.tensorflow.types` classes?

`RawTensor rawTensor = RawTensor.allocate(TFloat32.class, Shape.of(2, 2), -1);`
Still inherits from `NdArray`.

rnett: You can't user-side, but all of the `Tensor.fromHandle` methods don't inherit, afaik.  And in places where that's used (Session, somewhere in Operand.asTensor()) if a `asTypedTensor()` is missed somewhere it's nice to not crash when printing.  However, afaik, that's covered by the `tensor = ((RawTensor) tensor).asTypedTensor();` line, so the method should support all Tensors not just TType.

JimClarke5:
Actually the print code is keying on inheriting from NDArray. It has nothing to do specifically with TType.  That is why @karllessard ’s comment is throwing me off.

karllessard:
@JimClarke5 yeah maybe my comment wasn't clear. I was just pointing out that the `dataToString()` method could be added in `TType` instead of `Tensor` since only typed tensors are supported anyway. The `dataToString()` version for raw tensors is converting to a typed tensor before printing it, something the user could do directly if needed.

But I'm also fine leaving it there, in `Tensor`, if that makes things easier or if you think it make more sense.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Tensors.java
karllessard: I think `Ops.print()` simply writes a given string to an output stream, like the console, when the graph is being executed. 

I picked the "print" expression in this example just to show the logic but it can for sure be named otherwise to avoid the confusion. Maybe `TensorStringifier`, `tensor.stringify()` and `tensor.stringifier()`? Or the methods could remain `dataToString()` as well.

ajs1998:
Any updates on this? I am very new to Tensorflow and I have no idea how to visualize simple output. This seems perfect but it's been a year an a half since the last comment.

karllessard:
@ajs1998 you can print the value of any operand (`Operand`) using TF debug operations, such as `tf.print(tf.strings.stringFormat(tensors...))` for example. 

But if you want to print the values of a `Tensor` object itself without iterating through it, then that is still lacking afaik and should probably be added to the [ndarray](https://github.com/tensorflow/java-ndarray) module instead. Could be an interesting and easy first PR for anyone interested to contribute to the project!

ajs1998:
@karllessard Excellent! Didn't know I could print an operand like that, thank you.

PR #276, opened 2021-04-06
Fetch resource variable fix

Craigacp:
The test added in #261 was using a malformed protobuf as TF functions have to have at least one input. This caused the test to always fail. This PR fixes the python generation function, along with the protobuf file and the test itself. The python file was slightly reformatted to make it a little more PEP 8 friendly.

I also fixed a few small Javadoc issues in the hand written code of tensorflow-core-api, presumably the generated ones will be fixed by the PR that rewrites that.

cc @rnett


karllessard (approved):

rnett:
Odd, I ran them locally just fine.  Something may have been cached somewhere, but I'm pretty sure I cleaned first.

Craigacp:
The TF function code barfed on it because it didn't have any inputs. I'm not sure what it actually wrote out, but the TF native library did not like it. Maybe you've got an older version of Python tf which does something different? I tried it on 2.4.1 and it did not like the protobuf that was there.

rnett:
Yeah, I've got 2.3.1 in Python, but the Java library is still using 2.4.1.  Odd, but whatever, it works now.

PR #279, opened 2021-04-07
Changing quick build over to use actions/setup-java and Java 8

Craigacp:
actions/setup-java is the preferred way of specifying the JDK in a Github action. This PR changes the quick-build ci pipeline over to use setup-java and specifies Java 8 as the build JDK. It doesn't modify any of the long builds or the packaging pipeline, we can experiment with introducing it there later.

This fixes our javadoc build failures. The javadoc still has errors in, but at least it doesn't fail the whole CI and ignore test results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


karllessard:
Multiple thanks @Craigacp !

karllessard (approved):

PR #287, opened 2021-04-11
Use the Java Op generator

### LINKED ISSUE #286: The protobuf version doesn't match the underlying tensorflow version
A comment in one of the project pom files states the protobuf version needs to match the underlying native tensorflow api. (see:  https://github.com/tensorflow/java/blob/v0.3.1/tensorflow-core/tensorflow-core-api/pom.xml#L18 ). This is setting the dependent protobuf version to 3.8.0.

The table in the tensorflow java readme says the underlying tensorflow version for tensorflow java 0.3.1 is 2.4.1. tensorflow 2.4.1 uses protobuf 3.9.2 (see https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/workspace.bzl#L584 ) not 3.8.0.

rnett:
Enables the Java op generator during `install` and removes the C++ generator.  Builds fine locally, although there are still Javadoc errors in core with openjdk14 for non-opgen classes (i.e. `scope` param tags are used in the `Ops` class).  I'll leave them to another PR.

@karllessard can you flag this for CI build?

It also fixes #286.


karllessard:
Ok so it passes a few platforms in the full build but timed out eventually and Bazel-caching is not enabled for PRs so we will now give it a shot in master. I did not went through all the 1000+ regenerated ops to find issues but we'll fix them as we find them, if any.

Great work @rnett ! I might miss that old C++ code of mine but that is just me being sentimental :)

karllessard:
@rnett can you please check why the Windows build is [failing](https://github.com/tensorflow/java/runs/2396351397?check_suite_focus=true) after merging this PR?

karllessard:
@rnett , we've noticed that on Windows, looks like there are only 603 files being compiled while on Linux, there are 1807. If that can give some cue...

rnett:
Odd, that seems like it's probably a problem with the op export code rather than the op generation.  Will look.

PR #301, opened 2021-04-26
Move ndarray library to its own repository

karllessard:
The Java NdArray library now have its own repository, to emphasis its independence from TensorFlow itself and allow it to have its own development/release cycle.

As part of the review of this PR, please take also a look at the new repository : https://github.com/tensorflow/java-ndarray and let me know if you see anything that should be changed there as well.

A note on this, I've moved the library as a sub-module since we might end up having other modules there as well, e.g. `ndarray-kotlin`. Also, I've picked `tensorflow-java-ndarray` as the name of the parent module so it matches the name of the repo (like `tensorflow-java` matches this one) but I guess this is also debatable.


karllessard:
CC\ @saudet

saudet:
Sounds alright and looks fine, but did you want me to look at something in particular?

Just one thing, we probably want to set the version to 0.4.0-SNAPSHOT instead of 0.3.1.

karllessard:
> Sounds alright and looks fine, but did you want me to look at something in particular?
> 
> Just one thing, we probably want to set the version to 0.4.0-SNAPSHOT instead of 0.3.1.

Nothing in particular, was just seeking for a general feedback like this one, thanks. The version is 0.4.0-SNAPSHOT btw, it’s just the example in the README that shows the usage of the last released version.

saudet:
Since we're probably going to end up with more modules, maybe we could rename the "ndarray" one to "ndarray-core" or something?

karllessard:
I think I still prefer the simplicity of `org.tensorflow:ndarray` for importing the main artifact of that library, then append a `-something` to any of its upcoming extensions. 

Ideally we would have done the same with tensorflow but `org.tensorflow:tensorflow` was already being used by TF 1.x

karllessard:
Ok I’ll take this as a go, thanks!

PR #302, opened 2021-04-27
Generatr op input and attribute accessors

rnett:
This adds a nested class to the generated op classes that contains typesafe (and not quite type safe) accessors for the op inputs and attributes.  It's only supported for graph ops at the moment, but it will eventually support `ForwardOperation`s as well (from the eager gradient support).  The main use for this is defining custom gradients, whether it's the legacy graph gradients or the new gradient API.

I haven't committed the generated outputs to make it easier to review, I'll do so after approval or in another PR.


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -411,12 +643,273 @@ private static int dtype(TF_Graph graphHandle, TF_Operation opHandle, int output
...
+          + ") for an operation that has " + numOutputs + " outputs");
     }
 
     try (PointerScope scope = new PointerScope()) {
       return TF_OperationOutputType(new TF_Output().oper(opHandle).index(outputIndex));
     }
   }
+
+  private static AttributeMetadata getAttrMetadata(TF_Operation handle, String name) {
+    try (PointerScope scope = new PointerScope()) {
```
Craigacp, line 656: This one is missing `requireHandle(handle)` but it still has a public accessor.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -330,6 +399,166 @@ public int numControlConsumers() {
...
+   * @param name the name of the attribute
+   * @return the value of the attribute
+   */
+  public Shape[] getAttrShapeList(String name) {
+    return getAttrShapeList(unsafeNativeHandle, name);
+  }
+
+
+  /**
+   * Get the metadata of a n attribute of this operation.
```
Craigacp, line 544: Typo `a n`

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -411,12 +643,273 @@ private static int dtype(TF_Graph graphHandle, TF_Operation opHandle, int output
...
+      long totalSize = metadata.totalSize;
+
+      PointerPointer<BytePointer> values = new PointerPointer<>(listSize);
+      SizeTPointer lengths = new SizeTPointer(listSize);
+      BytePointer storage = new BytePointer(totalSize);
+
+      TF_Status status = TF_Status.newStatus();
+      TF_OperationGetAttrStringList(handle, new BytePointer(name), values, lengths, listSize,
+          storage, totalSize,
+          status);
```
Craigacp, line 690: Formatting?

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/AttributeType.java
```diff
@@ -0,0 +1,31 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================
+ */
+package org.tensorflow.op.generator;
+
+public enum AttributeType {
+  String, Int, Float, Bool, Shape, Tensor, Type;
```
Craigacp, line 20: All caps for the enum constants?

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/ClassGenerator.java
```diff
@@ -63,7 +64,8 @@
   static boolean canGenerateOp(OpDef op, ApiDef apiDef) {
     return apiDef.getVisibility() != Visibility.SKIP
         && !op.getAttrList().stream().anyMatch(x -> x.getType().contains("func"))
-        && !op.getName().startsWith("_"); //TODO do I want this?  Some interesting ops like _XlaCompile
+        && !op.getName()
```
Craigacp, line 67: Why has this file been reformatted? Didn't you just add this?

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/TypeResolver.java
```diff
@@ -21,19 +21,18 @@
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeVariableName;
 import com.squareup.javapoet.WildcardTypeName;
+import java.util.ArrayList;
```
Craigacp, line 24: Why'd the imports move?

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/AttributeType.java
rnett: I'm using the names to resolve the methods, so I wanted to avoid having a redundant constructor parameter.  I can do that if it's a big deal.

Craigacp: You can put the parameter in the enum and get it back out of the object. I think directly using the enum constant names is probably a bad idea.

rnett:
The formatting issues were since my IDE somehow got off google-java-format.  I have the plugin now, which added some things that weren't in the XML.

karllessard:
Looks like the javadoc reformatting is still present

rnett:
Yeah, that happened when I applied the google-java-format plugin, and it matches the spec.  Idk why the XML format didn't do it, I can revert to that.  

We really should set up some kind of CI formatter, I keep getting burnt by this.  I haven't had time to look at it much but one of the plugins you linked in the lint PR looked like it did ktlint too (although it had a bug), that would be nice since we could to the Kotlin API in the same plugin.

rnett:
Reminder for myself: run generation before merging!  I'm going to work on a CI test for that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
```diff
@@ -330,6 +384,165 @@ public int numControlConsumers() {
...
+
+  /**
+   * Get the value of an attribute of this operation as an {@link AttrValue} proto.
+   *
+   * @param name the name of the attribute
+   * @return the value of the attribute as an {@link AttrValue} proto
+   */
+  public AttrValue getAttrValueProto(String name) {
+    return getAttrValueProto(unsafeNativeHandle, name);
+  }
```
karllessard, line 545: Just little suggestion, maybe instead of "polluting" the `GraphOperation` public namespace with all these new endpoints that are basically only used for internal purposes (by the new `Inputs` classes), we should move them all to some kind of `GraphOperationInspector` that is located somewhere in the `op.tensorflow.internal` package and add a `public GraphOperationInspector inspector()` function here?

rnett:
FYI @karllessard I'm going to be out of town most of the next 2 weeks or so.  I might have time to follow up on these PRs, but I'm not sure.  Feel free to take things over.  I should at least be able to reply to comments.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperation.java
rnett: Where do you think I should put the demarcation though?  I could do just attributes, but then I should probably do inputs too, and then it's close enough to consumers and outputs, and then there's nothing left in `GraphOperation`.

karllessard: True, I'll leave the existing methods where they are and I think it is find as well to have `inputs` at the `GraphOperation` level since there are already public endpoints to retrieve its outputs. So let's stick to attributes then.

karllessard (dismissed):
Thanks @rnett !

rnett:
Let me generate the op files first, too.

rnett:
Yeah, good point, I'll look at that first.

karllessard:
Looks good, I've just merged another PR (#388) though so I'll wait that one to complete building before merging this one.

karllessard (approved):

karllessard:
@rnett , I just realized that we forgot to a CI build round on this PR before merging it... :( 

Failing on Windows: https://github.com/tensorflow/java/runs/3908491275?check_suite_focus=true

It fails on Linux too but it looks more like a bug we are having with the Bazel cache and that was happening before.

PR #303, opened 2021-04-27
Upgrade for TensorFlow 2.5.0

saudet:
I was thinking of doing this as part of pull #283, but there are too many changes to the ops, so let's do that separately. The tests pass so it's kind of working, but if we remove the `src/gen` directory, a lot of ops don't get regenerated. Also, as you can see from the second commit, there is a lot of new stuff that doesn't look right.

BTW, this is still only a RC upstream, so if we find issues fast enough, we should be able to have them fixed before the release!
@karllessard @rnett


rnett:
The best way to track op differences is to use #305.  I compared the files and there doesn't seem to be anything too out of the ordinary.  Just standard op changes.

FYI the op generator deletes the base directory (`gen/java/org/tensorflow/op`) when you run it, so I don't know what's up with that.  It's deterministic based on `ops.pb`.

saudet:
Ok, so why doesn't everything gets regenerated when purging the `src/gen` directory?

rnett:
Try just purging the `org.tensorflow.op` package, I think the rest is just protobuf?  I'll clone your branch and check, but I can't think of a reason why ops wouldn't be generated.

rnett:
Deleting and re-generating `src/gen/java/org/tensorflow/op` works fine for me, it's the exact same thing.

saudet:
Ah, you're right, it looks like I simply forgot to rerun the whole build. It didn't work with the old generator though. Now, it's actually just a couple of proto classes that are gone. So, this looks alright after all.

saudet:
Ok, so, well, this looks good to me. Please take a look and feel free to merge! @karllessard

Craigacp:
2.5.0 came out yesterday (https://github.com/tensorflow/tensorflow/releases/tag/v2.5.0), could you bump this PR to use that? Then we can get it merged in.

karllessard:
As usual, when upgrading the version of TF, we need to run the [API importer](https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/bazel/api_def/import/api_import.cc) to classify properly the new ops (which are all added to the root package by default). 

The process is not fully straightforward and even a bit arbitrary. I'll try to find some time soon to run it but I guess I should document what I'm doing so that others can also do it (and maybe even better!). I'll push my changes to your branch @saudet , can you please grant me write access to it if I don't have them already?

saudet:
You should already have write access, yes. Let me know if you don't see the "Add more commits by pushing to the..." message below the comments here though.

karllessard:
@googlebot I consent.

karllessard:
@saudet I'm done classifying the new ops and protos in TF2.5 but we are hitting an error in the MacOSX CI build, any chance you can take a look at it? I think it happens here:

```
Usage: /Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool [-change old new] ... [-rpath old new] ... [-add_rpath new] ... [-delete_rpath old] ... [-id name] input
```

karllessard:
I've just disabled (maybe temporarily) this [previous path fix](https://github.com/tensorflow/java/blob/7b5a1ca635a6e45cb091e0abae8fafc061553b02/tensorflow-core/tensorflow-core-api/build.sh#L70), which seems to cause the error. We'll see how the CI build goes.

karllessard:
As the SIG has already discussed during our last session, we'll disable for now MKL builds because they are unstable and causing us too much trouble to build, while seemingly not providing any improvement in the performances (and actually, quite the opposite).

PR #305, opened 2021-04-27
Add pbtxt op converter

rnett:
Adds a utility `main` to convert the exported ops to protobuf text, for diffing (i.e. #303).

I'm very tempted to have the op exporter do this, it will cause op differences to show up in the git diffs (and doesn't increase the Jar size, since it's not included).  Thoughts?


karllessard:
Can't the C++ bazel job create both?

rnett:
Yeah, it can.  I just had to write this up anyways to check #303.  Don't merge this, I'll add it to the exporter.

karllessard:
@rnett , closing this one now that the pbtxt is generated by the C++ generator

PR #308, opened 2021-04-29
Use spotless plugin for formating

rnett:
Uses the [spotless plugin](https://github.com/diffplug/spotless/tree/master/plugin-maven) for formatting.  Will check formatting in the `initialize` phase (so its before codegen & native builds), and auto-format can be ran with `mvn spotless:apply`.  We can use it for the Kotlin API as well once diffplug/spotless#142 is fixed.

I have it set to ratchet from `master` (i.e. only check changed files), do we want to do a big "apply formatting everywhere" PR (it's ~500 files)?

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,5 @@
+--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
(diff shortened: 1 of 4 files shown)


karllessard:
I'm not convinced about reformatting source code with Maven, any reason why simply using [Google Java Style formatting](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) isn't enough?

rnett:
It differs a bit from the [actual plugin](https://plugins.jetbrains.com/plugin/8527-google-java-format), at least for me (things like line length and single line comments).  Plus, it's easy to forget and have a bunch of improperly formatted files end up in the repo, like we have now.  This doesn't actually do any formatting, it just checks it as part of `install`, like the lint checks.

karllessard (approved):

karllessard:
@rnett I should have labeled that PR with CI build before merging... can you please check the multiple errors we are facing on many platforms from GitHub Actions? I might need
to revert that PR unless you find a quick fix for it, thanks

rnett:
Ugh, it seems to be that `--add-exports` isn't recognized for JDK 8, but is required for 16.  I'll remove and ignore that file for now.

rnett:
Do you or @Craigacp know if there's a way to have a version-dependent `.mvn/jvm.config` file?

rnett:
See #324.

PR #306, opened 2021-04-29
Export ops.pbtxt in addition to ops.pb

rnett:
#305, but done properly.  The ops.pbtxt isn't used by the generator, but it's good to have as documentation and for git diffs.

PR #310, opened 2021-04-30
Setting all the optimizers to have useLocking = True

Craigacp:
I'm running into determinism issues when training models using TF-Java. This is one area which could be causing it, as in TF 2 all the optimizers have `useLocking=true`. We don't currently set this in TF-Java, and I'm worried the code path is degrading (as it says the behaviour may be undefined but faster with `useLocking=false`).

This doesn't resolve my non-determinism issue completely, but it seems a little better.

I've added a test to GradientDescentTest which checks that the models produced are identical. This test fails randomly, for two reasons, both of which are confusing. First it seems like when we fetch the weights from a model, we get a pointer to the weights, not a copy of them. This means that when they are trained the "copies" I'm saving out as the initialized ones are being updated. Second, and this is the real issue, the gradient updates can be different on identical models for identical data, for no apparent reason. I think this is happening somewhere in the C API, as I can't see where it could be happening in our code and the models use *identical GraphDefs*.

### DIFF
--- a/tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/AdaGrad.java
+++ b/tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/AdaGrad.java
@@ -19,6 +19,7 @@ import org.tensorflow.Graph;
 import org.tensorflow.Operand;
 import org.tensorflow.Output;
 import org.tensorflow.op.Op;
+import org.tensorflow.op.train.ApplyAdagrad;
 import org.tensorflow.op.core.Variable;
 import org.tensorflow.types.family.TType;
 
@@ -41,6 +42,9 @@ public class AdaGrad extends Optimizer {
   public static final String ACCUMULATOR = "accumulator";
   public static final float LEARNING_RATE_DEFAULT = 0.001f;
   public static final float INITIAL_ACCUMULATOR_DEFAULT = 0.01f;
+
+  private static final ApplyAdagrad.Options[] opts = new ApplyAdagrad.Options[]{
+          ApplyAdagrad.updateSlots(true),ApplyAdagrad.useLocking(true)};
 
   private final float learningRate;
 
@@ -140,7 +144,7 @@ public class AdaGrad extends Optimizer {
   protected <T extends TType> Op applyDense(Output<T> gradient, Output<T> variable) {
     Variable<T> slot = getSlot(variable, ACCUMULATOR).get();
     return tf.train.applyAdagrad(
-        variable, slot, tf.dtypes.cast(tf.constant(learningRate), gradient.type()), gradient);
+        variable, slot, tf.dtypes.cast(tf.constant(learningRate), gradient.type()), gradient, opts);
   }
 
   /** {@inheritDoc} */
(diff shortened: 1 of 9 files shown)


karllessard:
@rnett , looks like the quick build is failing after merging #306 , can you please take a look at it?

karllessard:
@Craigacp can you please rebase once more so we can give the quick-build another try?

Craigacp:
Done.

Craigacp:
Looks like everything passed this time.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/AdaGrad.java
```diff
@@ -42,6 +43,9 @@
   public static final float LEARNING_RATE_DEFAULT = 0.001f;
   public static final float INITIAL_ACCUMULATOR_DEFAULT = 0.01f;
 
+  private static final ApplyAdagrad.Options[] opts = new ApplyAdagrad.Options[]{
+          ApplyAdagrad.updateSlots(true),ApplyAdagrad.useLocking(true)};
```
karllessard, line 47: Super nit: any formatter will probably complain about the missing space after a comma.

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/GradientDescentTest.java
```diff
@@ -97,4 +112,158 @@ public void testBasic() {
...
+                .feed("output", targetTensor)
+                .fetch(lossName)
+                .run().get(0);
+        postTrainingLoss[i] = lossVal.getFloat();
+        lossVal.close();
+      }
+    }
+
+    for (int i = 1; i < numRuns; i++) {
+      assertEquals(initialLoss[0],initialLoss[i]);
```
karllessard, line 240: Super nit: spaces after commas.

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/GradientDescentTest.java
```diff
@@ -97,4 +112,158 @@ public void testBasic() {
...
+        s.run(initName);
+
+        initialized.add(
+            s.runner()
+                .fetch(fcWeightName)
+                .fetch(fcBiasName)
+                .fetch(outputWeightName)
+                .fetch(outputBiasName)
+                .run());
+        System.out.println("Initialized - " + ndArrToString((TFloat32)initialized.get(i).get(3)));
```
karllessard, line 207: Can we avoid the verbosity in the unit test, aren't the equality check enough to validate? I'm fine just commenting out these `println`

karllessard (approved):
Thanks @Craigacp , I've approved it but left a few minor comments here and there, if you want to take a look

Craigacp:
I'll clean up the test on Monday, and split it into two. There should be one test for the outputs returning a reference rather than a copy of the weights (as this might allow you to mutate the weights directly, which seems bad), and that test which just checks the gradient behaviour.

It's all conflated in that single test as I spent hours trying to figure out what was going on, so it has got all the print statements and other stuff necessary for me to track it down.

I'll clean up the formatting issues at the same time.

karllessard (approved):

PR #322, opened 2021-05-13
Load TF library before computing TString size

### LINKED ISSUE #320: TString.scalarOf core dumps on OSX
**System information** 
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): **No**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): **OSX 11.3.1 (intel)**
- TensorFlow installed from (source or binary): binary **TF Java 0.3.1 and 0.4.0-SNAPSHOT**
- TensorFlow version (use command below): **2.4.1 as built with TF Java 0.3.1**
- CUDA/cuDNN version: **No**
- GPU model and memory: **No**

**Describe the current behavior**

Calling `TString.scalarOf("hello");` will core dump on OSX. It will actually succeed 1 out of 10 times.

**Describe the expected behavior**

I want to be able to create a `TString` without a core dump.

**Code to reproduce the issue**

JUnit test

```
    @Test
    public void testTString() throws Exception {
        TString.scalarOf("hello");
    }
```

**Other info / logs**

Running same test case through docker linux vm succeeds.

One core dump from a run:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00000001422d724c, pid=16660, tid=5635
#
# JRE version: OpenJDK Runtime Environment (11.0.2+9) (build 11.0.2+9)
# Java VM: OpenJDK 64-Bit Server VM (11.0.2+9, mixed mode, tiered, compressed oops, g1 gc, bsd-amd64)
# Problematic frame:
# C  [libtensorflow_cc.2.dylib+0x822924c]  tensorflow::TensorInterface::ByteSize() const+0xc
#
# Core dump will be written. Default location: /cores/core.16660
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

---------------  S U M M A R Y ------------

Command Line: /Users/jakob/proj/tfjava031/target/surefire/surefirebooter11868128427691651602.jar /Users/jakob/proj/tfjava031/target/surefire/surefire16140580774728156639tmp /Users/jakob/proj/tfjava031/target/surefire/surefire_012592081837142384420tmp

Host: MacBookPro13,2 x86_64 3100 MHz, 4 cores, 16G, Darwin 20.4.0
Time: Tue May 11 11:47:43 2021 CEST elapsed time: 3 seconds (0d 0h 0m 3s)

---------------  T H R E A D  ---------------

Current thread (0x00007fb48f809000):  JavaThread "main" [_thread_in_native, id=5635, stack(0x00007000053f2000,0x00007000054f2000)]

Stack: [0x00007000053f2000,0x00007000054f2000],  sp=0x00007000054eeea8,  free space=1011k
Native frames: (J=compiled Java code, A=aot compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libtensorflow_cc.2.dylib+0x822924c]  tensorflow::TensorInterface::ByteSize() const+0xc
j  org.tensorflow.internal.buffer.TensorBuffers.tensorMemory(Lorg/tensorflow/internal/c_api/TF_Tensor;)Lorg/bytedeco/javacpp/Pointer;+5
j  org.tensorflow.internal.buffer.TensorBuffers.toStrings(Lorg/tensorflow/internal/c_api/TF_Tensor;J)Lorg/tensorflow/internal/buffer/ByteSequenceTensorBuffer;+1
j  org.tensorflow.internal.types.TStringMapper.mapDense(Lorg/tensorflow/RawTensor;)Lorg/tensorflow/types/TString;+11
j  org.tensorflow.internal.types.TStringMapper.mapDense(Lorg/tensorflow/RawTensor;)Lorg/tensorflow/types/family/TType;+2
j  org.tensorflow.RawTensor.asTypedTensor()Lorg/tensorflow/types/family/TType;+8
j  org.tensorflow.Tensor.of(Ljava/lang/Class;Lorg/tensorflow/ndarray/Shape;J)Lorg/tensorflow/types/family/TType;+10
j  org.tensorflow.Tensor.of(Ljava/lang/Class;Lorg/tensorflow/ndarray/Shape;JLjava/util/function/Consumer;)Lorg/tensorflow/types/family/TType;+3
j  org.tensorflow.types.TString.tensorOf(Ljava/nio/charset/Charset;Lorg/tensorflow/ndarray/NdArray;)Lorg/tensorflow/types/TString;+28
j  org.tensorflow.types.TString.tensorOf(Lorg/tensorflow/ndarray/NdArray;)Lorg/tensorflow/types/TString;+4
j  org.tensorflow.types.TString.scalarOf(Ljava/lang/String;)Lorg/tensorflow/types/TString;+4
j  com.peltarion.tf.test.JakobTest.testTString()V+2
v  ~StubRoutines::call_stub
V  [libjvm.dylib+0x39b736]  JavaCalls::call_helper(JavaValue*, methodHandle const&, JavaCallArguments*, Thread*)+0x21a
V  [libjvm.dylib+0x64c917]  invoke(InstanceKlass*, methodHandle const&, Handle, bool, objArrayHandle, BasicType, objArrayHandle, bool, Thread*)+0x4e6
V  [libjvm.dylib+0x64c3e7]  Reflection::invoke_method(oopDesc*, Handle, objArrayHandle, Thread*)+0xf5
V  [libjvm.dylib+0x42236d]  JVM_InvokeMethod+0x20a
...
```

karllessard:
Fixes #320 

Make sure the TF native libraries have been loaded properly before instantiating a String tensor, since we need to know in advance the size of the `TF_TString` native class to compute the required size of the whole buffer.

Also cleaned up a few unrelated unit tests... (@rnett if you want to take a look, especially at `BooleanMaskTest`)

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
@@ -32,6 +32,12 @@ import org.tensorflow.proto.framework.OpList;
 
 /** Static utility methods describing the TensorFlow runtime. */
 public final class TensorFlow {
+
+  /** Make sure all TensorFlow native libraries have been loaded properly */
+  public static void init() {
+    // Do nothing, we'll let the class static initializer load the native library if needed
+  }
+
   /** Returns the version of the underlying TensorFlow runtime. */
   public static String version() {
     return TF_Version().getString();
@@ -106,7 +112,7 @@ public final class TensorFlow {
   private TensorFlow() {}
 
   /** Load the TensorFlow runtime C library. */
-  static void init() {
+  private static void initTensorFlow() {
     try {
       NativeLibrary.load();
     } catch (Exception e) {
@@ -123,6 +129,6 @@ public final class TensorFlow {
   }
 
   static {
-    init();
+    initTensorFlow();
   }
 }
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
@@ -33,11 +33,6 @@ import org.tensorflow.proto.framework.OpList;
 /** Static utility methods describing the TensorFlow runtime. */
 public final class TensorFlow {
 
-  /** Make sure all TensorFlow native libraries have been loaded properly */
-  public static void init() {
-    // Do nothing, we'll let the class static initializer load the native library if needed
-  }
-
   /** Returns the version of the underlying TensorFlow runtime. */
   public static String version() {
     return TF_Version().getString();
@@ -112,7 +107,7 @@ public final class TensorFlow {
   private TensorFlow() {}
 
   /** Load the TensorFlow runtime C library. */
-  private static void initTensorFlow() {
+  static {
     try {
       NativeLibrary.load();
     } catch (Exception e) {
@@ -126,9 +121,5 @@ public final class TensorFlow {
       e.printStackTrace();
       throw e;
     }
-  }
-
-  static {
-    initTensorFlow();
   }
 }
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
@@ -404,6 +404,11 @@ public final class EagerSession implements ExecutionEnvironment, AutoCloseable {
   }
 
   static {
-    TensorFlow.init();
+    try {
+      // Ensure that TensorFlow native library and classes are ready to be used
+      Class.forName("org.tensorflow.TensorFlow");
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException(e);
+    }
   }
 }
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
@@ -1070,6 +1070,11 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
   }
 
   static {
-    TensorFlow.init();
+    try {
+      // Ensure that TensorFlow native library and classes are ready to be used
+      Class.forName("org.tensorflow.TensorFlow");
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException(e);
+    }
   }
 }
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/RawTensor.java
@@ -228,6 +228,11 @@ public final class RawTensor implements Tensor {
   private ByteDataBuffer buffer = null;
 
   static {
-    TensorFlow.init();
+    try {
+      // Ensure that TensorFlow native library and classes are ready to be used
+      Class.forName("org.tensorflow.TensorFlow");
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException(e);
+    }
   }
 }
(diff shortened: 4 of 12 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
```diff
@@ -32,6 +32,12 @@
 
 /** Static utility methods describing the TensorFlow runtime. */
 public final class TensorFlow {
+
+  /** Make sure all TensorFlow native libraries have been loaded properly */
+  public static void init() {
+    // Do nothing, we'll let the class static initializer load the native library if needed
```
Craigacp, line 38: Maybe we should use `Class.forName("org.tensorflow.TensorFlow")` to trigger initialization? Having an empty method that we call to trigger the class's static initializer seems unpleasant (and likely to make static analysis or linters grumpy).

saudet: It should already have been working correctly with `init()` in the static initialization block, and having that one call  `NativeLibrary.load()`. Why do we need to change this?

Craigacp: `ByteSequenceTensorBuffer` can't see the TensorFlow.init() method as it's package private, and nothing on the codepath for `TString.scalarOf` triggered TensorFlow.init() by hitting a static initialiser in other classes in `org.tensorflow`.

saudet: Yeah, we can make init() public or something, but that should be able it.

karllessard: I changed it because right now, everytime `TensorFlow.init()` is called (which is done in several places), [the JavaCPP loader](https://github.com/tensorflow/java/blob/a24b8cae2481b4ed75d4444b344c385bcf1b2a15/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/NativeLibrary.java#L48) is invoked to load TF classes, while it should basically be done only once. There is probably no issue calling it multiple times but no reasons neither afaik. So moving this call exclusively to the class static initializer of `TensorFlow` ensure that we try to load the library only once.

Now @Craigacp , you seem to foresee potential problems doing it that way, can you explain a bit more?

Craigacp: I think an empty method that we call to ensure a class is initialised is a code smell. `Class.forName` is documented as triggering the class initialization (and it's mentioned in the language spec - https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4), so that's always going to work, and we can document at the use site that it's to trigger library loading.

Samuel's suggestion of making init public also seems fine, though my preference in that case would be to have a `private static boolean isLoaded`, guard the `NativeLibrary.load` call by that, and also make the block synchronized on TensorFlow.class to ensure there aren't races in native library loading. @saudet everything in `NativeLibrary.load` after the call to `Loader.load(org.tensorflow.internal.c_api.global.tensorflow.class)` and the `isLoaded` check is irrelevant now right? It looks like that's all related to the old JNI library loader.

karllessard: I've pushed a new version using `Class.forName` as you suggested @Craigacp . Of course it makes that code more verbose than with my previous version but since it's only being called internally, I think it's fine.

Craigacp (approved):
LGTM

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
saudet: Loader.load() was designed to be called multiple times. It already has the equivalent of an isLoaded flag.

PR #324, opened 2021-05-23
Quick fix for spotless

rnett:
Removes `.mvn/jvm.config` since the `--add-exports` doesn't work on JDK 8.  Build won't work on JDK 16 unless you add the noted exports yourself (they are noted in CONTRIBUTING.md).  Alternatively, we could use 11 in the CI and rely on profiles to set the target.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,5 +1,0 @@
---add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
---add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
---add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
---add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
---add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,52 +2,71 @@
 
 ## Building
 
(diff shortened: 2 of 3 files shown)


karllessard:
Please @rnett , can you remove the reformatting changes in the `CONTRIBUTING.md` of your PR? Thanks

Also since it won't work with JDK16+, I would protect the execution of the spotless plugin only for the lower versions of the JVM, similar to [what I did](https://github.com/tensorflow/java/blob/ceae4897095218b9293bdce5b107f0ab02aa3e72/pom.xml#L187) for error-prone, so users compiling on it won't face issues.

We will eventually jump the CI build to use 11 but it is for now yet.

rnett:
Should be fixed.

> Also since it won't work with JDK16+, I would protect the execution of the spotless plugin only for the lower versions of the JVM, similar to what I did for error-prone, so users compiling on it won't face issues.

That's what I did for now.  Unfortunately, there's no way to do it via profiles since the args need to be set for the maven JVM, but it's noted so you'll just have to add it yourself on 16.

karllessard:
But right now the plugin is executed no matter which JDK you use, right? So building on JDK 16 will fail, unless you add the exports manually?

If so, my point is that it would be better to prevent the plugin to be executed on 16+ using Maven profile activation based on the JDK version, at least for now. This plugin is just a nice-to-have anyway and should not block our users

rnett:
Ah, I see.  It won't auto-run on JDK 16+ unless the `format` profile is enabled, but the plugin tasks will still be there, I'm assuming that if you're running the tasks manually you know to add the exports (or will when you get the error message).

karllessard (approved):

PR #327, opened 2021-05-31
Function ops

rnett:
Generates ops that take functions as parameters.


karllessard:
@rnett any way you can split that in multiple smaller PRs?

rnett:
Yikes, I'm not sure why it changed all the op classes.

rnett:
Ok @karllessard, the big change is in #328, I haven't applied it.  If you merge that PR I'll do another one with the generation, and then this one will be smaller after rebasing.

rnett:
Also @JimClarke5 this is what I was talking about wrt the runtime initialization checks, you should be able to use `ifOp` for that without too much trouble.  Although I'm not sure it's a great idea, and I'm working on the init scope PR (and variables soonish after that).

karllessard:
@rnett can you please rebase this PR now that we are on TF2.5? I guess new ops might also be added?

rnett:
I'm going to wait until #328 if you think you can merge it quickly, since this one needs to be rebased on top of it anyways.

rnett:
Ok, it's rebased.  The number of files is smaller, but still pretty big since the generation is in there.  I can try to make it smaller if you want, but if you need to check for groups it might be required anyways.

karllessard:
The number of files if fine, since most of the are generated (so I don't need to spend time reviewing them). Some ops needs to be reclassified into the right package though (like `ReduceDataset`).

karllessard:
@googlebot I consent.

Issue #202, opened 2021-01-31
Framework: Move Ops parameter to call method where possible


rnett:
I'd like to move the `Ops` parameters of framework classes to the call method, where possible.  This is primarily for Kotlin interop, but has a few other benefits as well.  It won't be possible for stateful classes (Metrics, Optimizers), but should be possible for most, as far as I can tell (Initializers, Activations, Losses).  I'm going to use Losses as a standin for all 3 for my examples.
1. Kotlin interop w/ `@FunctionInterface`.  If we do this, we can then define new losses like `Loss{ tf, x -> tf.stuff(x) }` or pass lambdas to methods that take losses.  This is very nice for layers, where we might have something Keras-like `Layer(activation=ReLU())` but want to replace it with something custom.
2. Re-use of objects.  Currently, losses create any subsequent calls in the same scope as their first call.  That means if it's called inside a sub-scope, including device ones, it ignores the scope.  This is somewhat expected, but not ideal.  It also causes further issues if we use `Ops` for (eager) tensor lifetime management, which has been suggested and is something I'd like to do (it's easy enough to make a long-lived copy of the initial scope, but then the tensors created in the call methods live forever, and the framework classes need to be closable).
3. Passing configs, like Keras.  In Keras, if you have a activation or loss that requires some parameters, you can pass it to a layer like `Layer(activation=LeakyReLU(alpha=0.3))`.  I expect this will be common with our API, as well.  Currently, this runs into the above issue w/ scoping, and prevents you from passing activations (or losses) from scopes that don't have an Ops instance available.

I'd look at having the stateful classes take `Ops` in `call` as well, and only using the constructor ops for initializing state.  This works better with scoping and lifetimes as mentioned above.

karllessard:
@JimClarke5 , we've been discussing Ryan and I about this and would like to revive this proposal, whenever you're ready

JimClarke5:
Actually, this may help with some of the stuff I am doing for Model.  The Model creates the graph, so it is difficult to create an Optimizer first, until the Model does this. Right now I have Model using Lambda functions to create Loss, Metrics and Optimizers after the Model creates its Graph/Ops. 

I am open to a way to defer setting the Ops (or whatever) later on in the Object's life cycle.

BTW: I think we should revisit Optimizers to be consistent with the rest of the framework packages.

rnett:
[...]

PR #65, opened 2020-05-29
Adding layers (based on keras) supporting multiple outputs

dhruvrajan:
This draft PR is tracking the implementation of keras layers in Java; will update with a README once the API is complete, and mark for review when it's ready.

PR #334, opened 2021-06-10
Reworked Layers Phase 1 

JimClarke5:
These are the first set of model layers, fairly simple ones without back propagation.

I have also revamped the `TestSession`, `GraphTestSession` and `EagerTestSession` to use lambdas for common operations between the `Graph` and `Eager` modes. This will need to be updated with #272, "Tensor.dataToString" when that is merged.

This PR supersedes #65, "Adding layers (based on keras) supporting multiple outputs"

This PR is dependent  #191 "TensorFormat enum". I temporary created a copy of TensorFormat until #191 is merged.

The layers framework supports multiple inputs and outputs, but some layers are restricted to single input/output by design,
(e.g. Dense).

This revision moved the `Ops` parameter out of the CTORs, and creates a new method, `init(Ops)`, and adds an `Ops` parameter to the call method  [e.g. `call(tf, ...)`]. see issue #202.

SequentialLayersTest tests chains layers together like in the Sequential Model.


JimClarke5:
@karllessard the `ops.pb` conflict is because I moved the `nn.raw `ops to `nn`, and moved the higher level `nn` ops to `framework`. This is a generated file so I am not sure how to resolve this conflict.

karllessard:
@deansher , any chance you can take a look at this work?

deansher:
Thank you, @karllessard, for inviting me to review. I have lately been exploring different aspects of deep learning that don't leave me the time to participate in tensorflow/java for now.

karllessard:
@JimClarke5 there is a lot of conflict in this PR, I think it's because we have merged a few of your other PRs before it, can you please rebase so I can start to take a look at it?

PR #335, opened 2021-06-11
SavedModelBundle leak fix

Craigacp:
This PR adds the deallocators to the `TF_Graph` and `TF_Session` that live inside a `SavedModelBundle` as previously they didn't free their memory on `SavedModelBundle.close()`.

We may want to backport this to the 0.3.x branch. I think the main conflict there is that there are formatting changes in `master` not present in `r0.3` but we could just write a similar patch, it's only 5 lines.

cc @saudet

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
@@ -15,6 +15,7 @@ limitations under the License.
 */
 package org.tensorflow;
 
+import static org.tensorflow.internal.c_api.global.tensorflow.TF_NewGraph;
 import static org.tensorflow.internal.c_api.global.tensorflow.TF_SetConfig;
 
 import com.google.protobuf.InvalidProtocolBufferException;
@@ -504,7 +505,7 @@ public class SavedModelBundle implements AutoCloseable {
       TF_Buffer runOpts = TF_Buffer.newBufferFromString(runOptions);
 
       // load the session
-      TF_Graph graph = TF_Graph.newGraph();
+      TF_Graph graph = TF_NewGraph();
       TF_Buffer metagraphDef = TF_Buffer.newBuffer();
       TF_Session session =
           TF_Session.loadSessionFromSavedModel(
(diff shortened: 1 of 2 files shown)


Craigacp:
Turns out `Graph` and `Session` use different idioms for closing their respective native object, and this difference broke the first version of this fix. We should probably make them use the same idiom, either both using JavaCPP's deallocators, or both directly calling the relevant TF C API functions on close. The split is confusing.

karllessard (approved):

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -510,21 +508,18 @@ private static SavedModelBundle load(
       TF_Graph graph = TF_NewGraph();
```
saudet, line 508: We probably should call `TF_Graph.newGraph()` here, but I'm not sure...

Craigacp: I tried that initially, but `Graph.close()` doesn't use `handle.releaseReference()` like `Session.close()` does, and so it doesn't call the `TF_Graph` deallocator. I think you're the last person to touch both `Graph.delete` and `Session.delete` and I wasn't sure if they were different for a reason.

As I said in the comment above, I think they should be made consistent, but that's going to be a bigger change than just the one which fixes this leak. For example if we switch `Graph` over to use `handle.releaseReference()` and then use `TF_Graph.newGraph` everywhere it'll mean that the `TF_Graph` inside `Graph` is now sensitive to enclosing `PointerScope`s which it currently isn't.

saudet: Is it that the TF_Graph becomes managed by the TF_Session and we shouldn't touch it anymore after that? It sounds like that from the documentation of the C API. In that case I'd put the call to TF_NewGraph() in TF_Session.loadSessionFromSavedModel() and forget about it.

Craigacp: No, I think the graph is separate from the session, or at least that's my reading of the doc here - https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h#L1229. The `TF_NewSession` doc above it suggests the `TF_Graph` is kept alive by the `TF_Session`, but it doesn't say that the session owns the graph, so I think we're still responsible for closing it.

killeent-stripe:
@Craigacp Would definitely a +1 to say that backporting this fix to a `0.3.x` release would be much appreciated!

PR #338, opened 2021-06-16
Init Scope

rnett:
Reworks how initialization is handled (somewhat, mostly the API).  I was originally planning on doing name-based for interop with tensorflow, but they don't do name based either so I kept the current method.

The API has changed: instead of adding init ops yourself, they will automatically be added when created with an init scope, which you get using `Ops.initScope()` or `Scope.initScope()`.  This allowed for better error checking (i.e. init ops can't depend on non init ops), eliding control dependencies on them (since they are created at init time), but required making `Scope`s part of OpBuilders (which was a good thing mostly, and prevents future "forgot to call apply" bugs).  I also added currently unused methods to do initialization in a different execution environment, which will be used by functions.

So for an example, creating a variable w/ an initial value becomes `tf.initScope().variable(tf.initScope().constant(4f))`, which automatically registers it for initialization by sessions.

I also added helpers to `Session` to create and initialize and a requirement that if the graph has init ops, initialization must b ran before running anything else.  I don't really like having separate `initialized` factory methods, thoughts on having it be initialized on construction by default but having `Session(Graph, boolean)` constructors where you can control it?

I have yet to update `framework` to use this, does it work for what you were working on w/ variables @JimClarke5?

The new generated op changes also aren't committed yet, for size reasons.


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -163,7 +166,14 @@ public Scope withNameAsSubScope(String defaultName) {
    * @return a new Scope that uses opName for operations.
    */
   public Scope withDevice(DeviceSpec deviceSpec) {
-    return new Scope(env, nameScope, controlDependencies, deviceSpec);
+    return new Scope(env, nameScope, controlDependencies, deviceSpec, isInit);
+  }
+
+  // TODO stop gradient recording in init scopes (once we have gradient recording)
+
+  /** Get an extension of this scope that generated initialization ops. */
```
Craigacp, line 174: s/generated/generates/

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -141,6 +141,46 @@ public void close() {
...
+        Runner runner = runner();
+        initializers.forEach(runner::addTarget);
+        runner.runNoInit();
+      }
+      hasInitialized = true;
+    }
+  }
+
+  /** Create a session and initialize it. */
+  public static Session initialized(Graph graph) {
```
Craigacp, line 171: These methods should have a better name. Maybe `createAndInitializeSession`? Though that's rather verbose.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -808,7 +815,7 @@ synchronized SaverDef saverDef() {
   private SaverDef saverDef;
   private final Scope baseScope;
 
-  private final List<Op> initializers = new ArrayList<>();
+  private final Set<Operation> initializers = new LinkedHashSet<>();
```
Craigacp, line 818: Do we guarantee that Operations don't have hashCode and equals implemented anywhere? This requires that we're using the identity hashcode for correct operation of the `isInitOp` check, so we might want to use `IdentityHashMap` to enforce that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperationBuilder.java
```diff
@@ -83,16 +85,24 @@
    */
   @Override
   public GraphOperation build() {
+    scope.apply(this);
```
Craigacp, line 88: So this adds all control dependencies from the scope to everything built by that ops? Isn't this a behaviour change?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -141,6 +141,46 @@ public void close() {
     }
   }
 
+  /**
+   * Execute the graph's initializers.
+   *
+   * <p>This method is equivalent to {@code session.run(Ops.create(session.graph).init())}.
```
Craigacp, line 147: This comment is incorrect now as `Init` doesn't exist.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -373,6 +413,15 @@ public Runner setOptions(RunOptions options) {
       return this;
     }
 
+    private void checkInitialization() {
```
Craigacp, line 416: This check worries me. Initializers aren't persisted in the GraphDef, and so if you deserialize a graph you can't find them. In Tribuo we have to run two sets of initializers to get around this problem, one for the graph that the user supplied, and one for the gradient, loss and outputs that we add to the graph before training.

I don't think this check is going to cause issues right now because it's not persisted, but we need to think carefully about how it interacts with persistence.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Init.java
```diff
@@ -1,110 +0,0 @@
...
-import org.tensorflow.Operation;
-import org.tensorflow.OperationBuilder;
-import org.tensorflow.op.Op;
-import org.tensorflow.op.RawOp;
-import org.tensorflow.op.Scope;
-import org.tensorflow.op.annotation.Endpoint;
-import org.tensorflow.op.annotation.Operator;
-
-@Operator
-public final class Init extends RawOp {
```
Craigacp, line 14: After this class has been removed how do I find the initializers of a Graph when I load it from a GraphDef? Before I could find the name of this op and save it out with the GraphDef, allowing me to run the initialization when all I have is a GraphDef. Without it I'm going to have to mirror the logic into user code and then have my own function which makes a `NoOp` with the required control dependencies.

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/SavedModelBundleTest.java
```diff
@@ -325,9 +324,11 @@ private static Signature buildGraphWithVariables(Ops tf, Shape xShape) {
     Placeholder<TFloat32> x = tf.placeholder(TFloat32.class, Placeholder.shape(xShape));
     Variable<TFloat32> y =
         tf.withName("variable")
-            .variable(tf.random.randomUniform(tf.constant(xShape), TFloat32.class));
+            .variable(
+                tf.initScope()
```
Craigacp, line 328: This change seems ugly. Is there no way we can grab the op and force it into the initializer scope?

Users are going to forget that they need to put the variables into the init scope.

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/ClassGenerator.java
```diff
@@ -1,19 +1,19 @@
 /*
-  Copyright 2021 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 2: Where's the actual change in this file? It looks like it's all formatting?

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/GradientDescentTest.java
```diff
@@ -22,7 +22,6 @@
 import org.tensorflow.op.Ops;
 import org.tensorflow.op.core.Assign;
 import org.tensorflow.op.core.Constant;
-import org.tensorflow.op.core.Init;
```
Craigacp, line 25: The disabled determinism test uses `Init`, so you should fix the reference there too.

Craigacp (changes requested):
This change is going to break Tribuo in ways that are hard to fix, because it removes the init op we use to find initializers in serialized GraphDefs and there doesn't seem to be a way to get the initialization op name back out. I think you should consider how this interacts with serialized GraphDefs which are the only way to persist a graph structure that we've got at the moment. The current init mechanism mirrors what TF Python does.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -633,6 +674,8 @@ Graph graph() {
   private final Object nativeHandleLock = new Object();
   private TF_Session nativeHandle;
   private int numActiveRuns;
+  private final Object initLock = new Object();
```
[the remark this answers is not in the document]
rnett: I needed to use it in `Runner` and forgot about `Session.this`.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: `GraphOperation` implements `equals` and `hashCode` based on the native pointer, which is fine for this (and better than identity, if you somehow have two Java objects pointing to the same native op, they should register the same as init ops).  `EagerOperation` doesn't but should, I'll add that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/GraphOperationBuilder.java
rnett: It was always applied manually in the generated code, i.e:
```java
OpBuilder builder = scope.env().opBuilder("", "");
builder.addInput(..);
builder = scope.apply(builder);
```
and not doing so would lead to bugs with name collisions (there was one in `Constant` at one point).

It would be a behavior change if you used the op builder manually without every applying the scope, but that would usually be a bug and wasn't the case in any of the non-test usages (Constant and ConcreteFunction, and ConcreteFunction's will be removed soon).

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/SavedModelBundleTest.java
rnett: Yeah, Python lifts variable initializers automatically, we can too but I'd need #302 first.  Perhaps something like `withInitScope(Function<Ops, T>)` for now?

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/op/generator/ClassGenerator.java
rnett: I changed how the op builder is gotten, in `buildFactoryMethods`.

tensorflow-framework/src/test/java/org/tensorflow/framework/optimizers/GradientDescentTest.java
rnett: Yeah I haven't updated `framework` yet.

rnett:
The exporting and importing is a problem, can you link where Tribuo does the exporting?  The intent was that when importing Python models, you would find the Restore op and use `Graph.registerRestoreOp`, I just need to add something similar for Java import and export.  I could swear I saw code for that somewhere but I can't remember where, do you know if we have something like that already?

rnett:
Looking more into `SaverDef`, I'm not going to be able to use that.  I'm planning to create essentially a `tf.init()` op with all the init ops as control dependencies and a set name pattern on export, and on import find that op and use `Graph.registerRestoreOp`.  This would mean that using `importGraphDef` would load all of the graph's initializers if the graph was exported from Java, which I think would solve your issue?

I also made `Session.restore` count as initialization for the checks.

Craigacp:
Tribuo uses `GraphDef` as it's representation of a graph before it's been trained, as its trainers are designed to be thread-safe so there can be multiple concurrent calls to train in flight. That means we need to be able to recreate the graph independently for each call to train. The `Graph` is only valid inside a call to train, then it's destroyed and recreated inside the `Model` object. All the logic is in the train call here - https://github.com/oracle/tribuo/blob/main/Interop/Tensorflow/src/main/java/org/tribuo/interop/tensorflow/TensorFlowTrainer.java#L466. It starts with a `GraphDef`, instantiates a `Graph` containing that `GraphDef`, adds the appropriate output and gradient update operations to that graph, runs init for the graph, runs init for the Tribuo added operations (which basically just inits the gradient slots), then starts training the model and stepping the gradient optimiser. The two phase init is because I can't easily aggregate all the necessary init variables from the supplied `GraphDef` with the ones Tribuo adds because I can't find them.

rnett:
Ok, GraphDef import/export is now supported, see https://github.com/tensorflow/java/blob/master/tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/GraphTest.java#L68-L68

tensorflow-core/tensorflow-core-api/src/test/java/org/tensorflow/SavedModelBundleTest.java
rnett: Well, I'd need #302 for functions, I can do this fine for now.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
rnett: What do you think about making it part of the constructor by default, but having a `Session(Graph graph, boolean initialize)` so that it can be disabled?  We're doing native calls anyways, and it should avoid people running into the init check by accident.

rnett: Another option would be to by default, run `init` before the first `run` call, with a boolean to disable doing that.

rnett:
@JimClarke5 I've only done enough updates to `framework` to get the tests to pass, you might want to do a full pass once this is merged.

karllessard:
@rnett , can you rebase this PR or fix the conflicts please?

rnett:
Ok, done.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -534,23 +561,47 @@ public void importGraphDef(GraphDef graphDef, String prefix) throws IllegalArgum
...
+    if (op.type().equals(Placeholder.OP_NAME)) {
+      throw new IllegalArgumentException("Can not make a placeholder an init op.");
+    }
+
+    graphOp.controlInputs().forEach(this::registerInitOp);
+    graphOp.inputs().forEach(x -> registerInitOp(x.op()));
+
+    initializers.removeAll(subgraphToOps(Collections.singleton(graphOp)));
+    initializers.add(op);
+    newInitializers = true;
```
karllessard, line 589: What happens with such sequence of calls:
```
graph.registerInitOp(x);
graph.toGraphDef();
graph.registerInitOp(y);
graph.toGraphDef();
```

`toGraphDef()` will add two op `INIT_OP_BASE_NAME`, which should fail right? Is that the expected behaviour?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
karllessard: @rnett , I think you've done that now so we can close this comment?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
karllessard: I think having a boolean `initialize` in the constructor sounds good, but let's keep the version without it with a default value to false, to mimic actual behavior?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -211,46 +248,81 @@ private Scope(
...
 
   /** Returns device string from the scope. */
   public String getDeviceString() {
     return deviceSpec.toString();
   }
+
+  /** Get whether this scope is building init ops. */
+  public boolean isInit() {
+    return isInit;
+  }
```
karllessard, line 327: Nit: order of `private` and `public` endpoints are getting a bit messy, can you move this method (and `getDeviceString()` while at it) before the private fields?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -211,46 +248,81 @@ private Scope(
...
+   *
    * @param builder OperationBuilder to add control inputs and device specification to
    */
   public OperationBuilder apply(OperationBuilder builder) {
     builder.setDevice(deviceSpec.toString());
-    return applyControlDependencies(builder);
+    for (Operation control : controlDependencies) {
+      if (isInit || !env.isInitOp(control)) {
+        builder.addControlInput(control);
+      }
```
karllessard, line 297: I have difficulties to understand, so it can occurs that init ops are being added to non-init scopes?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8143,6 +8076,26 @@ public Ops withSubScope(String childScopeName) {
     return new Ops(scope.withSubScope(childScopeName));
   }
 
+  /**
+   * Returns an API that builds init operations.
+   * <p>Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well, and are never used as control dependencies.
+   * Additionally, this scope drops all of its control dependencies.  If an input can not be made an init op (i.e. a Placeholder), will error on op creation.
+   */
+  public Ops initScope() {
```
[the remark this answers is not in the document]
rnett: It's `initScope` in Python and I didn't want to name it the same as the op it's replacing (could cause silent failures), but I could.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -522,9 +515,43 @@ public void importGraphDef(GraphDef graphDef, String prefix) throws IllegalArgum
...
+    }
+
+    operations()
+        .forEachRemaining(
+            op -> {
+              if (op.name().startsWith(initPrefix)) {
+                registerInitOp(op);
+              }
+            });
+  }
```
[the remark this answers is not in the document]
rnett: Kind of, Python handles its initialization via variable restores, which are already handled in `SavedModelBundle` afaik.  I wanted to be able to pull out the python ops, but they don't mark them at all so it's not easily possible.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerSession.java
```diff
@@ -333,6 +332,16 @@ public Scope baseScope() {
...
 
+  // initialization is meaningless for eager sessions
+
+  @Override
+  public boolean isInitOp(Operation op) {
+    return false;
+  }
+
+  @Override
+  public void registerInitOp(Operation op) {}
```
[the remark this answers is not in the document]
rnett: It won't be meaningless for functions (which will almost surely need their own ExecutionEnvironment class), which is why it's here.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: It won't, since the Scope will uniquify it.  I'm pretty sure I test that somewhere.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
rnett: This is saying that non-init ops shouldn't have control dependencies on init ops, since they will be done in initialization.  However, init ops should still be able to have control dependencies to other init ops.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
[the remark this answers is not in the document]
karllessard: yeah, in fact ideally it would match the name pattern used for other endpoints of `Ops` that starts a new scope, e.g. `withSubScope`, `withName`, `withDevice`, etc. I'm not sure I like it be following this logic, it would be `withInit` or `withInitScope`? Not sure it makes sense...

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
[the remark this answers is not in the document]
karllessard: @Craigacp any additional thoughts on that?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
karllessard: I guess then that it should not be allowed to do `tf.withControlDependencies(controls)` with init ops in the `controls` list, unless `tf` is within an init scope? Not sure if mixing init and non-init ops in the control list is a valid use case. Can we validate it upfront then and throw an exception in case of incoherence?

rnett: Yeah.  I don't want to throw an exception if you use an init op as a control dependency, since it's valid semantically, it's just always executed beforehand so we don't need to add anything more.  I will throw if you try to depend on a non-init op in an init scope though.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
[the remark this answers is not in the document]
rnett: `withInitScope` seems fine to me.

rnett:
I'm considering a semantics change to fix some of the issues with framework that will likely appear elsewhere:  right now, if you run `session.initialize()`, then add more initializers, they will be ignored.  What I want to do is have sessions track un-ran initializers and run them before any run.  This better matches the meaning of python's `init_scope`.  It would make Session configuration a bit more complicated though.  I think the best way to handle that would to enable the auto-init by default (it is very similar to as if you had control dependencies on the init vars, they would be ran together) but provide an option to not, which would throw an exception if there are any un-ran inits.

JimClarke5:
How does this work if you need to init a variable after `session.initialize` is called?  My main focus is on initializing variables created later.

rnett:
That's exactly what I want to change, right now you would have to not use `initScope` and manage control dependencies manually.  But with this change it would work fine.  It guarantees that all `init` ops added will be ran before each `run`.  I'm planning on removing `initialize()` altogether, so each `run` call would check for any un-ran init ops and run them.

JimClarke5:
Great. 

@karllessard  I think I’ll wait on layers until this is committed because I’ll need this for `Dense` `kernel` and `bias`.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8143,6 +8076,26 @@ public Ops withSubScope(String childScopeName) {
     return new Ops(scope.withSubScope(childScopeName));
   }
 
+  /**
+   * Returns an API that builds init operations.
+   * <p>Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well, and are never used as control dependencies.
+   * Additionally, this scope drops all of its control dependencies.  If an input can not be made an init op (i.e. a Placeholder), will error on op creation.
+   */
+  public Ops withwithInitScope() {
```
karllessard, line 8084: Oops ;)

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
```diff
@@ -15,21 +15,19 @@
  */
 package org.tensorflow.framework.optimizers;
 
+import java.util.*;
```
Craigacp, line 18: Where'd this star import come from?

Craigacp (changes requested):

rnett:
Note to self: unsynchronize, add note that changing graph from another thread will cause issues.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8143,6 +8076,26 @@ public Ops withSubScope(String childScopeName) {
     return new Ops(scope.withSubScope(childScopeName));
   }
 
+  /**
+   * Returns an API that builds init operations.
+   * <p>Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well, and are never used as control dependencies.
+   * Additionally, this scope drops all of its control dependencies.  If an input can not be made an init op (i.e. a Placeholder), will error on op creation.
```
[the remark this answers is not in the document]
rnett: Does "ignores any control dependencies" sounds better to you (instead of drops)?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8143,6 +8076,26 @@ public Ops withSubScope(String childScopeName) {
...
+   */
+  public Ops withInitScope() {
+    return new Ops(scope.withInitScope());
+  }
+
+  /**
+   * Make {@code op} an init operation, doing the same for all of it's inputs (and control inputs).
+   * <p>Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well, and are never used as control dependencies.
```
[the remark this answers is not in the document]
rnett: It just ignores them, does "and are ignored when used as control dependencies" sound better?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -141,6 +171,41 @@ public void close() {
...
+  }
+
+  /**
+   * Execute the graph's initializers, regardless of whether the session has been initialized.
+   *
+   * <p>This runs any ops that have been created with an init scope.
+   *
+   * @return this
+   */
+  public synchronized Session forceInitialize() {
```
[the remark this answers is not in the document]
rnett: It's if you need to reset the session state.  Not something you should be using probably, but there just in case, since there's no way to do it otherwise.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -211,46 +249,88 @@ private Scope(
...
-    return new Scope(env, nameScope, controls, deviceSpec);
+
+    return new Scope(env, nameScope, toAdd, deviceSpec, isInit);
   }
 
   /**
    * Applies device specification and adds each Operand in controlDependencies as a control input to
    * the provided builder.
    *
+   * <p><b>Should only be used from {@link OperationBuilder} implementations</b>
```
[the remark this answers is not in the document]
rnett: Unfortunately not, it's used in the operation builder classes which are in the base package.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/Scope.java
```diff
@@ -211,46 +249,88 @@ private Scope(
...
-   * Adds each Operand in controlDependencies as a control input to the provided builder.
+   * Handle op creation, like registering it as an init op if the scope is init.
    *
-   * @param builder OperationBuilder to add control inputs to
+   * <p><b>FOR INTERNAL USE ONLY</b>
    */
-  public OperationBuilder applyControlDependencies(OperationBuilder builder) {
-    for (Op control : controlDependencies) {
-      builder = builder.addControlInput(control.op());
+  public void onOpCreated(Operation op) {
```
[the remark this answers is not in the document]
rnett: Same as `apply`, the builders are in the base package.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Helpers.java
```diff
@@ -22,32 +22,30 @@
...
+   *
+   * @param scope current scope
+   * @param init The op to use to initialise this variable.
+   * @param options carries optional attributes values
+   * @return a new instance of Variable
+   */
+  @Endpoint(name = "variable")
+  public static <T extends TType> Variable<T> createVariableWithInit(
+      Scope scope, Operand<T> init, Variable.Options... options) {
+    Variable<T> newVar = Variable.create(scope, init.shape(), init.type(), options);
```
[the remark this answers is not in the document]
rnett: Why?  It's not required, you could create and initialize a variable in non-init scopes.  Probably not a good idea, but it's valid.  And we don't force the normal variable that way, either.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
rnett: Not sure, but it passes style checks.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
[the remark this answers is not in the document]
rnett: Yeah, you should be able to just get the op by name and then register it as an init op (`Ops.liftToInitScope` or `ExecutionEnviroment.registerInitOp`).  And I can use `Init` as the name, I just wanted to make sure it didn't clash with names from other TF lbiraries.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -534,23 +561,47 @@ public void importGraphDef(GraphDef graphDef, String prefix) throws IllegalArgum
...
+    GraphOperation graphOp = (GraphOperation) op;
+
+    if (op.type().equals(Placeholder.OP_NAME)) {
+      throw new IllegalArgumentException("Can not make a placeholder an init op.");
+    }
+
+    graphOp.controlInputs().forEach(this::registerInitOp);
+    graphOp.inputs().forEach(x -> registerInitOp(x.op()));
+
+    initializers.removeAll(subgraphToOps(Collections.singleton(graphOp)));
```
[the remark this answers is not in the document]
rnett: It doesn't matter, since `subgraphToOps` doesn't include the outputs, but if that changes at some point (or there's a bug) it's less likely to remove itself.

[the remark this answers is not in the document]
rnett: It doesn't actually need to add the inputs though, just check them.

rnett:
Snapshot builds appear to be broken, the mac artifact is missing from https://oss.sonatype.org/content/repositories/snapshots/org/tensorflow/tensorflow-core-api/0.4.0-SNAPSHOT/

cc @karllessard

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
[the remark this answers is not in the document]
Craigacp: Yep, that works. "drops" sounds like it mutates something.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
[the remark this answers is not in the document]
Craigacp: Yep

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Helpers.java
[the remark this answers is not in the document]
Craigacp: The initialization won't be executed though. There's no access to the assign operation so you'd have to figure out what it's name was and run it by hand. That's going to be very difficult to figure out.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
Craigacp: Well we should fix that because they are prohibited under the Google Java Style (https://google.github.io/styleguide/javaguide.html#s3.3-import-statements).

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8077,19 +8077,29 @@ public Ops withSubScope(String childScopeName) {
...
    * Make {@code op} an init operation, doing the same for all of it's inputs (and control inputs).
-   * <p>Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well, and are never used as control dependencies.
-   * Additionally, this scope drops all of its control dependencies.  If an input can not be made an init op (i.e. a Placeholder), will error on op creation.
-   * @throws IllegalArgumentException if the op or one of its inputs can't be made an init op.
+   * <p>
+   * Init operations will be initialized at session creation, will have their inputs (and control inputs) made init ops as well,
+   * and are ignored when used as control dependencies.
+   * Additionally, this scope ignores any control dependencies.
+   * <p>
+   * If an input can not be made an init op (i.e. a Placeholder), will throw an {@link IllegalStateException} on op creation.@see ExecutionEnvironment#registerInitOp(Operation)
```
Craigacp, line 8100: Put the `see` on a separate line, it'll be formatted like that in Javadoc anyway.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -75,7 +76,8 @@
  * <p><b>WARNING:</b>A {@code Session} owns resources that <b>must</b> be explicitly freed by
  * invoking {@link #close()}.
  *
- * <p>Instances of a Session are thread-safe.
+ * <p>Instances of a Session are thread-safe. Modifying the graph from another thread <b>is not
+ * safe</b>.
```
Craigacp, line 80: I think this should be more explicit. Modifying the graph concurrently while another thread is using a session from that graph is not thread safe.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -715,7 +719,7 @@ Graph graph() {
   private int numActiveRuns;
 
   private final boolean autoInit;
-  private Set<Operation> ranInits = new LinkedHashSet<>();
+  private Set<Operation> ranInits = Collections.synchronizedSet(new LinkedHashSet<>());
```
Craigacp, line 722: Why is this a synchronized set? It's either empty, or set to `graph.initializers()` (which is an unmodifiable view on a synchronized set). As far as I can tell the object allocated here is never mutated.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -545,18 +545,21 @@ private synchronized void addInitOp() {
...
     initializers.forEach(builder::addControlInput);
     builder.build();
-
     newInitializers = false;
   }
 
   /**
    * Generate a representation of the Graph.
    *
+   * <p>Will create an initialization operation that depends on all registered init ops.
```
Craigacp, line 561: Maybe "If there are newly registered initializers on this graph this call adds an initialization operation to this graph that depends...". The current text isn't explicit that it mutates the graph, nor does it state when it does that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Helpers.java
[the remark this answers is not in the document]
rnett: Yeah, that makes sense, I guess it did add it to `Init` originally.  I don't like the difference between variable and variable-with-init, but I suppose that will be fixed when I do the `tf.Variable` API anyways.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
rnett: Huh, it's not applying to generated files.  I'm going to make another PR for that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/core/Helpers.java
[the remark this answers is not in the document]
Craigacp: I don't particularly like the distinction either, especially given the poor interaction with eager mode, but it's all we've got at the moment.

tensorflow-framework/src/main/java/org/tensorflow/framework/optimizers/Optimizer.java
Craigacp: That file isn't generated though, I wrote that file.

rnett: I mixed up `Ops` and `Optimizers`, that's even weirder.  Let me look.

rnett: Huh, apparently it's not supported by the formatter, that's annoying.  See https://github.com/google/google-java-format/issues/113.  I'll see if I can find a way to do it via spotless.

rnett: Fixed it manually for now.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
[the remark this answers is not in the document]
karllessard: Just following up on this, so do we plan to use "Init" or not? I guess the problem by doing is if we import a graph with an op named "Init" that is actually not an init op. At the same, I might increase our chance to find the init op to register from models trained on another platform than Java. It is kind of a blind guess... 

I'm personally betting that the latter is more probable. Also, I think doing so will be more backward compatible with our previous implementation of the Init op, right? So yeah, my vote goes for "Init" too.

JimClarke5:
@rnett I am trying to build on macOS, pulling `rnett:rn_init_scope`, and I keep getting these errors:

```
[ERROR]   SavedModelBundleTest.exportFunctionWithVariables:208 expected: <17.781294> but was: <18.12133>
[ERROR]   SavedModelBundleTest.exportMultipleFunctions:127 expected: <4.093658> but was: <5.931101>
[ERROR]   SessionTest.saveAndRestore:234 expected: <org.tensorflow.internal.types.TFloat32Mapper$DenseTFloat32@2024886b> but was: <org.tensorflow.internal.types.TFloat32Mapper$DenseTFloat32@8a1de29>
```
Any suggestions.

rnett:
@JimClarke5 It should, although I'm not sure we want to purposefully introduce that kind of eager-like graph modification to session semantics.  Imo it would be odd to intentionally have a Java function that's not a session run affect the session state, the "run all new initializers" was intended more as a failsafe.  You could run into the thread safety issues w/ modifying a graph while the sessions is open, too.  Ideally, I think the solution would be to have things like Metrics use Eager variables to store their state, but that makes managing them more complicated.

JimClarke5:
I have totally refactored `Metrics` to take advantage of this PR and it's a resounding **success**.
Good job @rnett 

I have removed the `Ops` parameter in the CTORs and now pass it in the new `Metric` interface  calls.
There is no longer a need for code to create control op dependencies for initialization nor a need to initialize the 
metrics objects once they are created.
This will help with the `Model` class where the `Ops` is created after the metrics are created.
Also, the `resetStates` method returns an `Op` that will need to be run explicitly either alone or maybe as a control op 
to another `Op` graph structure. (I decided to take @rnett's advice not to treat it within an `initScope`.)

Here is the main interface for `Metric` now:

```
interface Metric {

  List<Op> updateStateList(
      Ops tf, Operand<? extends TNumber> values, Operand<? extends TNumber> sampleWeights);

  List<Op> updateStateList(
      Ops tf,
      Operand<? extends TNumber> labels,
      Operand<? extends TNumber> predictions,
      Operand<? extends TNumber> sampleWeights);

  <T extends TNumber> Operand<T> result(Ops tf, Class<T> type);

  Op resetStates(Ops tf);

  Op updateState(
      Ops tf, Operand<? extends TNumber> values, Operand<? extends TNumber> sampleWeights);

  Op updateState(
      Ops tf,
      Operand<? extends TNumber> labels,
      Operand<? extends TNumber> predictions,
      Operand<? extends TNumber> sampleWeights);

  <T extends TNumber> Operand<T> callOnce(
      Ops tf,
      Operand<? extends TNumber> values,
      Operand<? extends TNumber> sampleWeights,
      Class<T> type);
}
```

Once this PR is merged I will create another PR  for these metrics' changes.

My next goal is to rework  `Layers`.

JimClarke5:
I have finished converting `layers`, and `optimizers` and everything works.  I have refactored both to not include `Ops`/`Graph` in the CTORs. Once this PR is merged, I will create separate PRs to check in my changes.

Except for these failures in the overall build, I am good to go

```
[ERROR] Failures: 
[ERROR]   SavedModelBundleTest.exportFunctionWithVariables:208 expected: <17.614902> but was: <17.720098>
[ERROR]   SavedModelBundleTest.exportMultipleFunctions:127 expected: <4.6954327> but was: <5.659465>
[ERROR]   SessionTest.saveAndRestore:234 expected: <org.tensorflow.internal.types.TFloat32Mapper$DenseTFloat32@ab941f5f> but was: <org.tensorflow.internal.types.TFloat32Mapper$DenseTFloat32@91507d4d>
```

rnett:
Can you rebase on this branch?  The latest commit here fixed those for me.

JimClarke5:
@rnett I have pulled your latest code and all is good now. I am ok with this PR.

karllessard (approved):
LGTM, @Craigacp any additional comment before I merge this?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -607,10 +611,16 @@ public boolean isInitOp(Operation op) {
...
+   * #registerInitOp(Operation)}.
+   *
+   * <p>Note that NoOps aren't included in this list, since any inputs or control dependencies are
+   * guaranteed to also be in this list, and including the no-ops wouldn't change the initialization
+   * result.
    */
   public Set<Operation> initializers() {
-    return Collections.unmodifiableSet(initializers);
+    return initializers.stream()
+        .filter(x -> !x.type().equals(NoOp.OP_NAME))
```
Craigacp, line 622: I'm a little worried by this, as the old init code generated a `NoOp` to gather up the previous initializers which didn't have a specific name, so it might ignore inits in older TF-Java models (or TF-Python graphs).

Craigacp:
> LGTM, @Craigacp any additional comment before I merge this?

Apart from the NoOp issue that I just mentioned it all looks fine. 

I am worried about the level of change this is to the codebase and I can't quite see the reason for it, but that doesn't mean there isn't a good one.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
rnett: The register init op method adds all inputs and control inputs as init ops, even if you're adding a no-op, so it should work fine.  The adding still adds stuff for init no-ops, the just aren't returned here since they don't do anything.  If you add a "collection" init op that is a NoOp that just runs it's dependencies, and those dependencies are already init ops (i.e. what is done during graph def export), we don't want to then re-run session init or anything else depending on `initializers()`, since there isn't anything new to run.

rnett:
@Craigacp It's mostly to support functions, which are Graph's with Eager init scopes, which wasn't possible to implement using fake graph ops like we were using for init previously.

karllessard:
> I am worried about the level of change this is to the codebase and I can't quite see the reason for it, but that doesn't mean there isn't a good one.

According to @JimClarke5 [comment](https://github.com/tensorflow/java/pull/338#issuecomment-898771320), the improvements in the framework are substantial. Personally, I think if we can lift the burden of initialization from the users, it's a plus but saved models were already handling this for inference so changes are just impacting users that were doing their training in Java (@rnett, am I right here?)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
Craigacp: Only if the init ops in the graph def have the right prefix. I don't think old ones we generated will do. If this is going to leave behind those old graphs we should note that somewhere. Similarly I don't know if this will correctly load in graphs that use the TF 1 style global variable initializer.

rnett:
> According to @JimClarke5 [comment](https://github.com/tensorflow/java/pull/338#issuecomment-898771320), the improvements in the framework are substantial. Personally, I think if we can lift the burden of initialization from the users, it's a plus but saved models were already handling this for inference so changes are just impacting users that were doing their training in Java (@rnett, am I right here?)

Yeah, the SavedModel handles it for Tf2 models.  This also makes it a big easier to handle TF1 models and we can export our initializers easier (I'm going to make a new PR for those).

JimClarke5:
This PR will be beneficial whenever a model is created  by the developer, which includes all the major model actions `fit`, `evaluate` and `predict`. Actions, such as `predict`, may not necessarily require that the model be saved first.

Craigacp (approved):
Ok, I think I'm fine with this being merged.

I do wonder if we could get away with just tracking the init ops better as we own all of them without creating a privileged init scope that people could use incorrectly (and isn't type safe), but I don't understand the eager/graph function use case well enough to know if that would cause problems.

karllessard:
Thanks @rnett !

PR #340, opened 2021-06-16
Added .bazelversion to be consitent with the TF build process

JimClarke5:
Added `.bazelversion` to `tensorflow-core-api` to be consistent with the TF Build process defined at [Build from source](https://www.tensorflow.org/install/source).  The TF build source recommends using  [Bazelisk](https://github.com/bazelbuild/bazelisk) which runs the  `bazel` version given in `.bazelversion`. Otherwise, the exact matching `bazel` version will need to be installed for each TF version.

This file needs to be kept in sync with `@org_tensorflow/.bazel_version`.


Craigacp:
Could you add a note to `CONTRIBUTING.md` in the native build section reminding us to keep the `.bazelversion` in sync with tensorflow/tensorflow? That way we might remember to do it. Or at least have somewhere to point people at when things go wrong.

CONTRIBUTING.md
```diff
@@ -44,6 +44,11 @@ configuration script (`./configure`), and copy the resulting
 `.tf_configure.bazelrc` to `tensorflow-core-api`. This overrides the default options, and you can add to it manually (i.e. adding `build --copt="-g"`
 to build with debugging info).
 
+The `tensorflow-core/tensorflow-core-api/.bazelversion` file must  be kept in sync with `@org_tensorflow/.bazel_version`. 
+This allows using [Bazelisk](https://github.com/bazelbuild/bazelisk) which runs the bazel version given in .bazelversion instead of having to 
+physically reinstall a specifc `bazel` version each time the TensorFlow version changes.
```
karllessard, line 49: "specifc" -> "specific"

CONTRIBUTING.md
```diff
@@ -44,6 +44,11 @@ configuration script (`./configure`), and copy the resulting
 `.tf_configure.bazelrc` to `tensorflow-core-api`. This overrides the default options, and you can add to it manually (i.e. adding `build --copt="-g"`
 to build with debugging info).
 
+The `tensorflow-core/tensorflow-core-api/.bazelversion` file must  be kept in sync with `@org_tensorflow/.bazel_version`.
```
karllessard, line 47: "must  be" -> "must be"

karllessard (changes requested):

CONTRIBUTING.md
JimClarke5: OK

CONTRIBUTING.md
JimClarke5: OK

karllessard (approved):

PR #342, opened 2021-06-20
Redo Framework Ops based on the current version of Master.

JimClarke5:
This is a reopen of #255 with a new branch using the latest master, as the old PR was way out of sync with the current master.

Added org.tensoflow.framwork.op.FrameworkOps to tensorflow-frameworks.
For now, these are not generated, but hard coded.

These are higher level ops that may invoke core ops. Higher level Ops may perform the
operation solely in the TensorFlow framework or do preprocessing of the Operands before invoking
a core level Op.

As part or this PR,tf.nn.raw generated package was removed and those ops are now generated directly into org.tensorflow.op.nn undertensorflow-core-api.
org.tensorflow.op.NnOps uses the core (formerly raw) ops for
SoftmaxCrossEntropyWithLogits and SparseSoftmaxCrossEntropyWithLogits.

FrameworkOps now contains the high level ops, sigmoidCrossEntropyWithLogits, softmaxCrossEntropyWithLogits, and sparseSoftmaxCrossEntropyWithLogits.

Also I moved SetsOps to org.tensorflow.framwork.op and l2Normalize to org.tensorflow.framwork.op.MathOps.
There are more framework ops when layers are checked in.

The easiest way to use it, for example is:

FrameworkOps fops = FrameworkOps.create(tf);
Operand<T> result =  fops.math.l2Normalize(tLabels, axis);
...
Operand<T> diffResult = fops.sets.difference(weightsShape2D, validDims);


tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/global/tensorflow.java
```diff
@@ -3150,7 +3150,7 @@ public static native void TF_RegisterFilesystemPlugin(
 // TF_InitKernel to do op/kernel registration.
 // Plugin should implement TF_InitKernel to register kernels. This function
 // should register all kernels in a plugin.
-public static native void TF_InitKernel();
+
```
karllessard, line 3153: Note that I know where this change comes from :) I'm skipping this C API endpoint that was causing trouble on TF2.5 but I forgot to commit the updated generated bindings so let's keep it here

karllessard (approved):
Looks good @JimClarke5 , thanks!

PR #348, opened 2021-07-05
Multi-release jar support (Java 9 to start)

rnett:
Adds support for multi-release jars.  Only Java 9 to start with, but it's easy to extend.

Adam or Karl, can you add the CI build tag?  I need to check that native artifacts are still built correctly.

TODO:
- [ ] Add tests (will need a CI task to test w/ Java 9+).
- [ ] Add binary compat tests?  See https://github.com/hboutemy/maven-jep238/blob/master/multirelease-base/pom.xml#L22

### DIFF
--- a/tensorflow-core/tensorflow-core-api-java9/pom.xml
+++ b/tensorflow-core/tensorflow-core-api-java9/pom.xml
@@ -7,7 +7,7 @@
     <artifactId>tensorflow-core</artifactId>
     <version>0.4.0-SNAPSHOT</version>
   </parent>
-  <artifactId>tensorflow-core-java9</artifactId>
+  <artifactId>tensorflow-core-api-java9</artifactId>
   <packaging>jar</packaging>
 
   <name>TensorFlow Core Java 9 API Classes</name>
--- a/tensorflow-core/tensorflow-core-api-java9/pom.xml
+++ b/tensorflow-core/tensorflow-core-api-java9/pom.xml
@@ -1,0 +1,31 @@
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.tensorflow</groupId>
+    <artifactId>tensorflow-core</artifactId>
+    <version>0.4.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>tensorflow-core-java9</artifactId>
+  <packaging>jar</packaging>
+
+  <name>TensorFlow Core Java 9 API Classes</name>
+  <description>Java 9 classes for the Tensorflow Core API</description>
+
+  <properties>
+    <java.module.name>org.tensorflow.core.api.java9</java.module.name>
+    <maven.compiler.source>9</maven.compiler.source>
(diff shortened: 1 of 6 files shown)


Craigacp:
You should start with Java 11, there aren't any supported releases of Java 9. Plus the CI is already set up with Java 11 support.

tensorflow-core/tensorflow-core-api-java9/pom.xml
```diff
@@ -0,0 +1,31 @@
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
```
karllessard, line 1: Are you creating a new module to avoid errors in the IDE? Because I think there is a way to add `java9+` source files in the same module (`tensorflow-core-api`) but in a separate folder, I'd personally prefer that.

rnett: Yeah, it's for IDE issues.

karllessard: I don't think we should design our project based on IDE's "bugs", they might come up with a fix later (if none already exists?).

I'd be curious to give it a try if you place the JDK9 sources in, let say, `src/java9`

rnett: Yeah, in theory, but I think this one was pretty severe.  If I remember right @JimClarke5 tried this (w/ source folders) but put it off because of the IDE issues?  I'm going to try it.

This is also fully supported by maven (see https://maven.apache.org/plugins/maven-compiler-plugin/multirelease.html) so it's not like it's a hack.

JimClarke5: Maven handles them nicely, but InteliJ complains of duplicate classes for those classes with different implementations as it doesn't seem to understand the difference between the 2 Java versions.

There is an open issue at:
[Add support for building Java 9 multi-release jars](https://youtrack.jetbrains.com/issue/IDEA-168946?_ga=2.232548127.1822431595.1625598749-216887102.1625598749 )

and
[IntelliJ support to multi-release jars](https://youtrack.jetbrains.com/issue/IDEA-221747)

As of Feb 2021, these still seems to be an open issues.

rnett: Testing it, it seems like you only get that error on the declarations, it's not that bad to just ignore.  It keeps everything in the same module and has less maven configuration.  There's no way to allow the source folder to use Java 9+ (or whatever) though, so using any 9+ APIs there will result in false positive errors.

karllessard: I personally find the solution with multiple directories in the same module way cleaner, it requires almost no additional Maven configuration, nor any inter-module manual assembly, and it works like a charm outside the IDE.

I think it depends on how much code we want to put in these version-specific classes. If we just add the code that differs between JDK versions, then I'd be comfortable to just leave the `src/main/java9` as a non-source directory by default and there will be no error. Of course the IDE won't be as efficient to assist you writing the code under that folder but it should be rare cases. But if we are heading to do a full duplicates of the core classes (e.g. `EagerSession`), then it may become more painful. I would not recommend doing that though.

@Craigacp , any thoughts on this?

rnett: The only thing I had planed to put in there was some sort of Cleaner for resource scopes, so that should be fine.

rnett:
Moved to #351.

PR #353, opened 2021-07-15
Fix dead url

aviperl:
Current URL is a 404; replace with updated URL.


aviperl:
@googlebot I signed it!

karllessard (approved):
Thanks @aviperl !

Issue #313, opened 2021-05-03
Large latency on Tensor allocation


nebulorum:
**System information**
- TensorFlow version: 0.3.1
- Doc Link: https://www.tensorflow.org/jvm/install

**Describe the documentation issue**

Not sure this is the correct forum but I would like to some guidance on how to setup sessions and resource management would be interesting. 

After two weeks trying to understand why latencies in 0.3.1 were completely uncontrollable (as compared to official 1.3.1) I ran into #208. This matches my observations. 

We are trying to run Prediction on models with thousands of data points in different tensor per prediction. Memory allocation on the threads are in 20MB/s and there seems to be a sync between JavaCCP Allocation thread and our worker threads. In addition to this allocation using Size(1) tensor seems to be very slow (in the 7ms range). 

After reading #208, it seems we are doing everything wrong. But I don't really have a clear picture of how it should be done: Would `EagerSession` help? Could I use a `Session` per HTTP request? Should I allocate larger multi-dimensional  tensors instead of a single one? How should I configure thread pools?  I understand that the API is work in progress, but current documentation is very light on this kind of documentation.

I don't think this is a bug, but I can convert into some other sort of issue.

Craigacp:
`Session` should be long lived as recreating it involves a bunch of allocations and other things. If you can batch up the incoming requests then that will allow the TF native runtime to perform more efficient batched matrix multiplies/convolutions (in many cases). I think the `Session` object should be thread safe so you can use it concurrently from multiple threads.

We should probably turn off the JavaCPP memory allocation size check and let the GC clear it up, if you're creating lots of Java side objects GC should run frequently enough to clear things out anyway. Are you hitting the thread.sleep inside JavaCPP's allocator?

Note if you're making lots of sessions out of saved model bundles you might be hitting some memory leaks inside TF's C API (which the TF Java API uses to interact with the native library). I opened this issue (https://github.com/tensorflow/tensorflow/issues/48802) upstream but we've not had a response yet.

nebulorum:
Running the JVM under the profiler I never see the JavaCCP deallocation thread sleep. It is either in wait, monitor or running state. On lower load most of the time on wait and then monitor/running sync up with the worker threads. 

We  have a simple model, that can run in 3ms under low load. Our overall call should happen under 40ms, but when load increases we can clearly see waits in our telemetry. Model evaluation is still quick but something is causing contention on the JVM. 

[...]

PR #354, opened 2021-07-16
Resource scopes draft

rnett:
This is a rough draft of resource scopes.  It's not fully implemented or used anywhere, but the deallocation behavior and API are there, and that's mostly what I want comments on.  Synchronization can probably be done better, too, but not to much should change (suggestions on that are welcome if you notice anything that sticks out).

There's four variants from two variables: implicit/explicit and strong/weak.  Implicit/explicit controls whether the scope itself is GC'd (unlike panama, you can still manually close implicit scopes).  Strong/weak controls whether the resources will be GC'd.  A strong scope holds on to all of it's references and dereferences them on close, while a weak scope only holds onto them weakly, and thus only dereferences the ones that are still reachable at close.  This can result in some resources being deallocated after their scope closes, which is why I provide the option for strong scopes, but it's mostly fine.

I've considered getting rid of implicit/explicit and just making all scopes be closed on GC, but I assume there's a reason panama didn't do that.  I don't think we would run into the same issues since we're reference counting: even if a scope is GC'd, its resources would only be deallocated if there were no other scopes referencing them, but I'm not sure enough to remove it.

I also have `addDependency` which works like [panama's](https://inside.java/2021/05/28/taming-resource-scopes/) in that a scope can't be closed until any scopes with dependencies on it are.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/resource/Deallocator.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/resource/Deallocator.java
@@ -1,0 +1,23 @@
+/*
+  Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ =======================================================================
+
+ */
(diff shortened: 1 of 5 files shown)


rnett:
The WeakIdentityHashMap is taken from https://github.com/plume-lib/plume-util and is mostly a placeholder atm.

rnett:
I'm planning on doing the same thing as discussed in https://inside.java/2021/09/16/finalizing-the-foreign-apis/: registering all scopes w/ the cleaner, but also allowing them to be closed manually.

saudet:
> I'm planning on doing the same thing as discussed in https://inside.java/2021/09/16/finalizing-the-foreign-apis/: registering all scopes w/ the cleaner, but also allowing them to be closed manually.

That's fine, as long as there is a way to disable this behavior, see issue #313.

rnett:
I'm also going to add some way to add "on added" and "on removed" handlers at a scope level, to support stuff like the new gradient API's memory management (see the "Some details on memory management" section above [here](https://github.com/tensorflow/community/blob/master/rfcs/20201201-cpp-gradients.md#use-cases)).

Issue #357, opened 2021-07-30
Bug in String deallocation


brychcy:
When trying to upgrade to 0.3.1 we got junit failures in a class with multiple
tests that load and use saved models with string inputs and outputs.

Tensorflow was for some tests throwing an exception with a message that definitely was wrong.

When the tests were run individually, no error occured.

When the tests were reordered, different tests were failing.

So this smells likely a memory corruption, and I've looked at the changes in  0.3.1 (w.r.t. 0.3.0),
namely #253.

I think I've understood the issue: The problem appears if the underlying TensorBuffers are shared, in which case
the strings should not be deallocated.

I'll send a PR which uses TF_TensorMaybeMove (which in its current implementation returns null if then TensorBuffer is shared, and the tensor itself, if it is not) to detect and avoid the situation (the PR fixes the problem for us)

brychcy:
See PR #358

karllessard:
Thanks for the PR @brychcy , we'll take a look at it.

Can you also please retry with 0.3.2, which was released a few weeks ago? We have fixed a memory leak related to loading/unloading saved models, that might prevent other issues as well

brychcy:
Happens also with 0.3.2.
Actually I directly tried to upgrade to that but then noted that the problem already appears with 0.3.1.

brychcy:
This has been fixed on the master and in 0.3.3 so this bug can be closed

PR #358, opened 2021-07-30
Don't dealloc strings if the TensorBuffer is shared (#357)

brychcy:


saudet:
There's no more straightforward way to get the information? It doesn't sound like TF_TensorMaybeMove() is always going to return null when the content is being shared.

brychcy:
> There's no more straightforward way to get the information? It doesn't sound like TF_TensorMaybeMove() is always going to return null when the content is being shared.

If there is, I haven't found it.

BTW, what I really wonder is: Shouldn't string deallocation be done automatically on the native Tensorflow level when the last tensor that uses the TensorBuffer is released?

saudet:
No, that doesn't happen. We need to allocate the strings manually, and deallocate them manually or we end up with memory leaks, see issue #251.

karllessard (approved):
Even if we are not 100% sure that this fix will cover all cases, it seemed to work for @brychcy so let's merge it to give it a try and see. Thanks @brychcy !

karllessard:
Oops, it looks like the modified code is not formatted correctly, @brychcy  can you please run `mvn spotless:apply` locally before pushing a newer version of this PR?

brychcy:
> Oops, it looks like the modified code is not formatted correctly, @brychcy can you please run `mvn spotless:apply` locally before pushing a newer version of this PR?

Done.

At first it didn't do anything for me because origin/master pointed to my version and 
<ratchetFrom>origin/master</ratchetFrom>
is used.

As the whole file was reformatted, I have done it in a separate commit.

karllessard (approved):
Thanks a lot @brychcy for fixing this!

PR #359, opened 2021-07-30
Apply spotless to generated files, exclude c api files

rnett:
Does what it says in the title.  I can include the C API files if you want, but most of them (especially the presets) have special formatting by @saudet and I'm not sure if we want to override that.


rnett:
Also, when I run generation locally I get a bunch of new op classes like `CacheDatasetV2` that are in `ops.pb` but don't seem to be generated, can anyone reproduce this?  Did we at some point not push the new ops?

rnett:
I also added a `.gitattributes` to set the line endings of most files to `LF`, since I was running into issues with it w/ the formatter when switching between Windows and Linux.  Most of the generation code assumes that anyways.

PR #360, opened 2021-07-31
ConcreteFunction fix and performance improvements

rnett:
Fixes using input types instead of output types and adds some performance improvements for `call`.

cc @karllessard


johnynek:
can we add a test of using two outputs which failed prior to this change?

That will help with a regression in the future.

karllessard:
Oops, I've just merged [that PR](https://github.com/tensorflow/java/pull/363) that fixes the output data type bug. @rnett , can you please rebase this one?

karllessard:
Ok as Ryan is out of town this week, he asked me to take over some of his PR if we need to, so I've rebased it and pushed it in [another PR](https://github.com/tensorflow/java/pull/364), including the test case you've been asking for @johnynek . So I'll close this one.

PR #361, opened 2021-08-03
Upgrade JavaCPP to just released version 1.5.6

### LINKED ISSUE #313: Large latency on Tensor allocation

saudet:
Fixes #313 

/cc @dennisTS @nebulorum


Craigacp:
Do we need to regenerate the preset?

saudet:
The build does that by default, but it won't change anything significant,
but yeah let's push that too.

karllessard (approved):
Thanks @saudet

PR #363, opened 2021-08-04
Fix outputDtypes in ConcreteFunction

### LINKED ISSUE #362: Unable to produce a ConcreteFunction with two outputs
When making a signature with two outputs, I get the following error:

```
[error] (run-main-10) org.tensorflow.exceptions.TensorFlowException: Function instantiation included invalid output index: 1 not in [0, 1).
[error] org.tensorflow.exceptions.TensorFlowException: Function instantiation included invalid output index: 1 not in [0, 1).
[error]         at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:101)
[error]         at org.tensorflow.GraphOperationBuilder.finish(GraphOperationBuilder.java:415)
[error]         at org.tensorflow.GraphOperationBuilder.build(GraphOperationBuilder.java:88)
[error]         at org.tensorflow.GraphOperationBuilder.build(GraphOperationBuilder.java:67)
[error]         at org.tensorflow.op.core.StatelessPartitionedCall.create(StatelessPartitionedCall.java:94)
[error]         at org.tensorflow.op.core.PartitionedCall.create(PartitionedCall.java:65)
[error]         at org.tensorflow.ConcreteFunction.call(ConcreteFunction.java:244)
[error]         at org.tensorflow.op.core.Function.call(Function.java:42)
[error]         at org.tensorflow.op.Ops.call(Ops.java:1214)
[error]         at org.tensorflow.ConcreteFunction.call(ConcreteFunction.java:327)
```

I believe the issue is that here the code claims to be asking for output data types, but it passes the input data types:

https://github.com/tensorflow/java/blob/50f7efa5821877668c2645886b0e4addef7306b6/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java#L247

see the call site expects output data types:
https://github.com/tensorflow/java/blob/50f7efa5821877668c2645886b0e4addef7306b6/tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/core/PartitionedCall.java#L42

this is with a snapshot version of 0.4.0 on August 3, 2021.

Here is an example that also throws an error related to wrong number of outputs, but not the exact same string:
```scala
    def dbl(tf: Ops) = {                                               
      val x: Placeholder[TInt32] = tf.placeholder(classOf[TInt32]);    
      val dblX = tf.math.add(x, x);    
      val tripX = tf.math.add(x, dblX)    
      Signature.builder().input("x", x).output("dbl", dblX).output("tripX", tripX).build();             
    }                                                                     
                                                                       
    val cf = ConcreteFunction.create(new Function[Ops, Signature] {    
      def apply(o: Ops) = dbl(o)    
    })                                                               
                                                                                       
    val out = cf.call(Map("x" -> (TInt32.scalarOf(2): Tensor)).asJava)    
    println(out.get("dbl").asInstanceOf[TInt32].getInt())                                    
    println(out.get("tripX").asInstanceOf[TInt32].getInt())  
```

which prints:
```
2021-08-03 16:19:19.303110: F external/org_tensorflow/tensorflow/core/framework/op_kernel.cc:973] Check failed: index < outputs_.size() (1 vs. 1)
zsh: abort
```

johnynek:
close #362 

this is a more minimal change than #360

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
@@ -244,7 +244,7 @@ public class ConcreteFunction implements AutoCloseable, TensorFunction {
         PartitionedCall.create(
                 scope,
                 inputList,
-                Arrays.stream(inputDtypes)
+                Arrays.stream(outputDtypes)
                     .map(x -> TensorTypeRegistry.find(x).type())
                     .collect(Collectors.toList()),
                 this)


johnynek:
I should add a test that shows the bug first, then add the fix.

johnynek:
@googlebot I fixed it.

karllessard (approved):
Thanks @johnynek , funny enough I've found the same bug just a few days ago and mentioned @rnett about it, so Ryan you won't need to include that fix anymore!

PR #364, opened 2021-08-05
ConcreteFunction fix and performance improvements

karllessard:
Signed-off-by: Ryan Nett <JNett96@gmail.com>

Rebase of https://github.com/tensorflow/java/pull/360

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
@@ -19,10 +19,10 @@ import static org.tensorflow.internal.c_api.global.tensorflow.TF_FunctionSetAttr
 import static org.tensorflow.internal.c_api.global.tensorflow.TF_GraphToFunction;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


karllessard:
@rnett , looks like you'll need to consent my new PR...

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -241,31 +241,31 @@ public String toString() {
...
+    } else {
+      Map<String, Operand<?>> namedOutputs = new LinkedHashMap<>(signature().outputNames().size());
+
+      Iterator<String> outputNames = signature().outputNames().iterator();
+      for (int i = 0; outputNames.hasNext(); i++) {
+        String outputName = outputNames.next();
+
+        if (i > outputList.size()) {
+          throw new IllegalStateException(
+              "Somehow, not all required outputs were returned from the function");
```
Craigacp, line 260: Might be useful if this exception returned both `i` and `outputList.size()`. Though I suspect it's very unlikely to happen.

Craigacp (approved):
A couple of small things but I'm fine if this is committed as is.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
```diff
@@ -218,7 +218,7 @@ public String toString() {
    * @return the outputs of the function
    */
   public Map<String, Operand<?>> call(Scope scope, Map<String, Operand<?>> arguments) {
-    List<Operand<?>> inputList = new ArrayList<>(signature.inputNames().size());
+    List<Operand<?>> inputList = new ArrayList<>(signature().inputNames().size());
```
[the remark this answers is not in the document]
karllessard: I've looked and didn't see any reason for calling `signature()` instead of `signature`, since there are no class extending `ConcreteFunction` that I'm aware of. I've marked the class as `final` to make sure of that.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/ConcreteFunction.java
karllessard: Done. I've also move that sanity check outside the loop, it could be done beforehand.

Craigacp (approved):
LGTM.

PR #375, opened 2021-08-28
[Quick Fix] Update op generation

rnett:
I forgot to re-run generation before merging init scopes.  This fixes that.  There's also a few ops that apparently weren't generated last time we added some?  Not sure why, but they are in `ops.pb`, so I'm guessing they just weren't committed.


karllessard:
A lot of changes here are undesirable, many ops that were classified before (e.g. under `data` or `xla`) now ends up in the core. That's the result of some changes in the ops-generator?

rnett:
No, that's the thing, I didn't change anything, just re-ran op generation locally.  Can you run it on your machine and compare it to what is here?  `ops.pb`, the api defs, and the op generation code is all unchanged, afaik.

rnett:
Might be best to just do this as part of #378, and make sure to commit the newly generated files.

karllessard:
Yes I was also thinking about cleaning that up while I'm working on #378, we can close this one

PR #376, opened 2021-08-28
Init exporting and loading

### LINKED ISSUE #365: TFFailedPreconditionException with BigGAN-deep model from TensorFlow Hub
**System information**
- OS: macOS Big Sur 11.3.1
- TensorFlow-java version: 0.3.1 
- Java version: I tried both 1.8 and 11

**Issue**

I am able to load [this](https://tfhub.dev/deepmind/biggan-deep-128/1) biggan-deep model into java with the SavedModelBundle.Loader but when I run it, I get

> TFFailedPreconditionException: Error while reading resource variable prev_truncation from Container: localhost. This could mean that the variable was uninitialized. Not found: Container localhost does not exist. (Could not find resource: localhost/prev_truncation)

The model works in Python. I initially encountered the issue when I used [DJL](https://djl.ai/) to load and run the model. I tried it with TF-Java to see if the issue is in DJL or TF-Java.

**Describe the expected behavior**

The model should run fine just like in python.

**Java code to reproduce the issue**

The main method and the `loadModel(...)` are the important ones, the rest are just tensor operations.

```
public static void main(String[] args) {
    SavedModelBundle model = loadModel("src/test/resources/biggan-deep-128_1", new String[0]);

    int[] input = {100, 207, 971, 970, 933}; // image classes

    Tensor y = oneHot(input, 1000);
    Tensor z = truncNormal(input.length, 128);

    Tensor result =
            model.session().runner()
                    .feed("y", y)
                    .feed("z", z)
                    .feed("truncation", TFloat32.scalarOf(0.5f))
                    .fetch("G_trunc_output").run().get(0);
}

private static SavedModelBundle loadModel(String dir, String[] tags) {
    SavedModelBundle.Loader loader = SavedModelBundle.loader(dir);

    try {
        Field field = SavedModelBundle.Loader.class.getDeclaredField("tags");
        field.setAccessible(true);
        field.set(loader, tags);
    } catch (ReflectiveOperationException e) {
        throw new AssertionError(e);
    }

    return loader.load();
}

private static Tensor truncNormal(int row, int col) {
    float[][] dist = new float[row][col];
    Random random = new Random();

    for (int i = 0; i < row; i++) {
        for (int j = 0; j < col; j++) {
            double sample = random.nextGaussian();
            while (sample < -2 || sample > 2) {
                sample = random.nextGaussian();
            }
            dist[i][j] = (float) sample;
        }
    }
    return toTensor(Shape.of(row, col), dist);
}

private static Tensor oneHot(int[] input, int numCategories) {
    float[][] dist = new float[input.length][numCategories];
    for (int i = 0; i < input.length; i++) {
        float[] row = new float[numCategories];

        Arrays.fill(row, 0.0f);
        row[input[i]] = 1.0f;

        dist[i] = row;
    }
    return toTensor(Shape.of(input.length, numCategories), dist);
}

private static TFloat32 toTensor(Shape shape, float[][] data) {
    FloatNdArray mat = NdArrays.ofFloats(shape);
    for (int i = 0; i < data.length; i++) {
        mat.set(TFloat32.vectorOf(data[i]), i);
    }
    return TFloat32.tensorOf(mat);
}
```

build.gradle:

```
dependencies {
    compile group: 'org.tensorflow', name: 'tensorflow-core-platform', version: '0.3.1'
}
```

**Other info / logs**

Console output: 

```
> Task :Main.main()
2021-08-04 22:07:22.105450: I external/org_tensorflow/tensorflow/cc/saved_model/reader.cc:32] Reading SavedModel from: src/test/resources/biggan-deep-128_1
2021-08-04 22:07:22.295033: I external/org_tensorflow/tensorflow/cc/saved_model/reader.cc:55] Reading meta graph with tags {  }
2021-08-04 22:07:22.295055: I external/org_tensorflow/tensorflow/cc/saved_model/reader.cc:93] Reading SavedModel debug info (if present) from: src/test/resources/biggan-deep-128_1
2021-08-04 22:07:22.295136: I external/org_tensorflow/tensorflow/core/platform/cpu_feature_guard.cc:142] 
This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2021-08-04 22:07:23.514662: I external/org_tensorflow/tensorflow/cc/saved_model/loader.cc:277] SavedModel load for tags {  }; Status: success: OK. Took 1409219 microseconds.
Exception in thread "main" org.tensorflow.exceptions.TFFailedPreconditionException: Error while reading resource variable prev_truncation from Container: localhost. 
This could mean that the variable was uninitialized. Not found: Container localhost does not exist. (Could not find resource: localhost/prev_truncation)
	 [[{{node Equal/ReadVariableOp}}]]
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:95)
	at org.tensorflow.Session.run(Session.java:691)
	at org.tensorflow.Session.access$100(Session.java:72)
	at org.tensorflow.Session$Runner.runHelper(Session.java:381)
	at org.tensorflow.Session$Runner.run(Session.java:329)
	at ai.tf.Main.main(Main.java:32)

> Task :Main.main() FAILED

Execution failed for task ':Main.main()'.
> Process 'command '/Library/Java/JavaVirtualMachines/amazon-corretto-8.jdk/Contents/Home/bin/java'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

rnett:
This adds proper initializer exporting to `SavedModelBundle`, and fixes #365.

Won't be so big once #375 is merged.


rnett:
Currently having issues since python does it's variable checkpoint restore, *then* runs init ops, which makes no sense.

I can manually run restore afterwards [here](https://github.com/rnett/java/blob/rn_init_exporting/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java#L601), but that will run it twice, and doesn't help if you load the model in python.

rnett:
I'm leaning towards not exporting our init ops at all (except maybe as bookkeeping).  Any state initialized or changed within them will be handled by variable restoration anyways.  And the only solution I can think of is removing `Assign` ops from the init ops before exporting or something similar, which is not great.

karllessard:
@rnett , can you please rebase your PR since we have just upgraded to TF2.6.0 and all ops were updated as well?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -260,13 +269,29 @@ public void export() throws IOException {
...
       functions.forEach((k, f) -> metaGraphDef.putSignatureDef(k, f.signature().asSignatureDef()));
 
+      if (!functions.containsKey(JAVA_INIT_OP_SIGNATURE_KEY)) {
+
+        metaGraphDef.putSignatureDef(
+            JAVA_INIT_OP_SIGNATURE_KEY,
+            SignatureDef.newBuilder()
+                .putOutputs(
+                    JAVA_INIT_OP_SIGNATURE_KEY,
+                    TensorInfo.newBuilder().setName(initOp.name() + ":0").build())
```
karllessard, line 291: You can now use `initOp.output(0).name()` instead

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -63,6 +67,11 @@
 public class SavedModelBundle implements AutoCloseable {
 
   public static final String DEFAULT_TAG = "serve";
+  private static final String JAVA_INIT_OP_SIGNATURE_KEY = "__saved_model_java_init_op_tracker";
+  private static final String INIT_OP_SIGNATURE_KEY = "__saved_model_init_op";
+  private static final String MAIN_OP_COLLECTION_KEY = "saved_model_main_op";
+  private static final String LEGACY_INIT_OP_COLLECTION_KEY = "legacy_init_op";
+  private static final String TABLE_INITIALIZERS_COLLECTION_KEY = "table_initializer";
```
karllessard, line 74: Can you please provide some details of what each of these keys is used for?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -365,7 +390,10 @@ public Session session() {
 
   /** Return the signature of all functions available in this saved model. */
   public List<Signature> signatures() {
-    return functions.values().stream().map(f -> f.signature()).collect(Collectors.toList());
+    return functions.values().stream()
+        .map(SessionFunction::signature)
+        .filter(signature -> !signature.key().equals(INIT_OP_SIGNATURE_KEY))
```
karllessard, line 395: Maybe comment why this signature should be filtered out?

karllessard:
> Currently having issues since python does it's variable checkpoint restore, _then_ runs init ops, which makes no sense.
> 
> I can manually run restore afterwards [here](https://github.com/rnett/java/blob/rn_init_exporting/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java#L601), but that will run it twice, and doesn't help if you load the model in python.

That is strange behavior if this is what is happening, indeed. Can you please provide more details on the purpose of this PR and summarize which what is the new logic flow in the init sequence when exporting/importing a model? I'm not sure to totally understand, thanks

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
rnett: I can't, since `initOp` is a NoOp, so trying to get an output that doesn't exist will error.

rnett:
> > Currently having issues since python does it's variable checkpoint restore, _then_ runs init ops, which makes no sense.
> > I can manually run restore afterwards [here](https://github.com/rnett/java/blob/rn_init_exporting/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java#L601), but that will run it twice, and doesn't help if you load the model in python.
> 
> That is strange behavior if this is what is happening, indeed. Can you please provide more details on the purpose of this PR and summarize which what is the new logic flow in the init sequence when exporting/importing a model? I'm not sure to totally understand what are the impacts of these changes, thanks

Yeah, I'm not sure why it's done that way.  Nothing really changes on our end for initialization, things are just remembered better when using SavedModels.  I.e. init ops will be correctly added to the init scope when loading a SavedModel, regardless of whether they were added from Python or Java.  It also adds table init ops from some hub models as init ops, so that they are ran (see #365).

PR #377, opened 2021-08-28
Release candidate for 0.3.3

karllessard:
Contains fixes for:

* Incompatibility between Java saved models and Python
* Memory leaks with String tensors


karllessard:
Hi @brychcy , we want to do a intermediary release of TF Java that includes your [fix](https://github.com/tensorflow/java/commit/709f5703f60fa17e42d8bb4345b6bb0b6162d4c0) for the memory leak in the string tensors. Can you please make GoogleBot happy by consenting again the CLA? (see above). Thanks a lot!

brychcy:
@googlebot I consent

Craigacp:
@googlebot I consent.

Craigacp (approved):
LGTM, but we should probably comment the place I marked.

karllessard:
@MarkDaoust , we have just released version 0.3.3 and we want the Java install page to reflect that new version.

The [install.md](https://github.com/tensorflow/java/blob/master/docs/install.md) file in our master branch has been updated accordingly. Can you please update the website? Also, I used (if I remember correctly) to create an issue in `tensorflow/docs` to notify you of this but it looks like can't create issues anymore:

image

has the process changed? Also, anyway to update this automatically when we push a change to the `docs/` folder of our repo?

MarkDaoust:
> 0.3.3

Okay, that will be published in a couple of minutes.

> if I  I remember correctly

We've always had the issues tab disabled...

karllessard:
> We've always had the issues tab disabled...

Thanks @MarkDaoust , so I really don't recall how we were doing it then :D I can add to our release process that we need to ping you directly on GitHub to refresh the page, unless you have something better to suggest?

MarkDaoust:
SGTM. This works fine. If I miss the email you can always make a bug on tensorflow/tensorflow with a "docs" tag, and add the rest of the team from tensorflow/docs.

PR #378, opened 2021-08-30
Upgrade for TensorFlow 2.6.0

saudet:
@karllessard This builds and runs fine, at least on my Linux machine, but as usual you'll probably want to take a look at the ops before merging this!


karllessard:
On it!

karllessard:
@saudet , I've rebased the PR and it looks like you need to reconsent now.

Also, I've noticed that there are a lot of changes in the `.bazelrc` file of TensorFlow that we did not import in our copy, I'm not sure if we should copy it/merge it at each upgrade, wdyt?

saudet:
@googlebot I consent.

saudet:
> @saudet , I've rebased the PR and it looks like you need to reconsent now.
> 
> Also, I've noticed that there are a lot of changes in the `.bazelrc` file of TensorFlow that we did not import in our copy, I'm not sure if we should copy it/merge it at each upgrade, wdyt?

Not sure what to say, this Bazel stuff is all really annoying.

karllessard:
@googlebot I consent.

karllessard:
> > @saudet , I've rebased the PR and it looks like you need to reconsent now.
> > Also, I've noticed that there are a lot of changes in the `.bazelrc` file of TensorFlow that we did not import in our copy, I'm not sure if we should copy it/merge it at each upgrade, wdyt?
> 
> Not sure what to say, this Bazel stuff is all really annoying.

Yeah same here, it's hard to say what are the real impact of updating/not-updating this file... Ok let's merge it as is and we can update it later if we think we need to. Thanks @saudet!

karllessard (approved):

PR #379, opened 2021-08-30
Fix for SavedModelBundle with an empty tag set

Craigacp:
Saved models can use the empty string as a tag set (e.g. https://tfhub.dev/google/elmo/3) which we had previously treated as invalid.


karllessard:
This has been merged to master from the 0.3.3 release

PR #385, opened 2021-09-20
Fix GPU builds

rnett:
Fixes GPU builds by updating `.bazelrc` and CI CUDA and CUDNN versions.

@karllessard can you tag this for CI build?

I just overwrote the `.bazelrc` file, if there's anything we want to add we should be doing it in a separate file, anyways.


rnett:
I'm testing this locally, but it's still building.

rnett:
I could copy the tensorflow bazelrc from the remote repo, and allow us to use our own for whatever we need.  It would make it much harder to edit though (you'd need to use a patch).  Thoughts?

rnett:
This is working on my machine, so we should be good to merge.  `tensorflow.bazelrc` is now the bazelrc from the tensorflow repo, and `.bazelrc` can be whatever we want (I have readonly build cache in there for now, setting it to write in the CI will take precedence).

rnett:
Would be good to figure out where the `QuantizedConcatV2` on my machine is coming from first though.

rnett:
The CI timing out since it can't reuse the cache.

karllessard:
> The CI timing out since it can't reuse the cache.

Yes, so let's merge it, anyway we know that the current build is not working properly on GPU devices, thanks @rnett

karllessard (approved):

rnett:
It's timing out on `master` too, I think you have to manually re-run it a couple times to build the cache.

karllessard:
Finally got the snapshots to build, I think it took 4 passes this time... you can try it out now

Issue #387, opened 2021-10-13
Error saving  SavedModel Using the Exporter API


HichemMaiza:
<em>Please make sure that this is a bug. As per our [GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md), we only address code/doc bugs, performance issues, feature requests and build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04 x86\_64): Windows 10
- TensorFlow Python installation :  pip install tensorflow==2.4.1
- TensorFlow Java version : 2.3.1 
- Java version (i.e., the output of `java -version`): Java 8 
- Python version (if transferring a model trained in Python): python 3.8 
- Bazel version (if compiling from source):
- GCC/Compiler version (if compiling from source):
- CUDA/cuDNN version: No Cuda
- GPU model and memory: No GPU

**Describe the current behavior**

I keep getting the Error `No Operation named [StatefulPartitionedCall_2:0] in the Graph` 
when Using `SavedModelBundle.exporter` to save the model

```java
ConcreteFunction serveFunction = savedModel.function("serve_model");
SavedModelBundle.exporter(exportDir)
                .withFunction(serveFunction)
                .export();
```

To access and inspect Graph operations, i can see the `StatefulPartitionedCall_2` 
But without the `:` at the end of the operation name.  

```java
Iterator<Operation> operationIterator  = serveFunction.graph().operations();
while(operationIterator.hasNext()){
    System.out.println(operationIterator.next().name());
}
```
code snippet output 

```shell
Adam/iter
Adam/iter/Read/ReadVariableOp
Adam/beta_1
Adam/beta_1/Read/ReadVariableOp
Adam/beta_2
...
...
...
train_model_labels
StatefulPartitionedCall_1
saver_filename
StatefulPartitionedCall_2
StatefulPartitionedCall_3
```

Works fine when invoking directly the Op from session.runner()

```java
String checkpointPath = "...";
session.runner()
        .feed("saver_filename:0", checkpointPath)
        .fetch("StatefulPartitionedCall_2:0").run() ;
```

The model was saved using a regular tf.saved_model.save function

````python
def save_module(module, model_dir):
    tf.saved_model.save(module, model_dir,
                        signatures={
                            'serve_model':
                                module.__call__.get_concrete_function(signature_dict),
                            'train_model':
[...]

PR #388, opened 2021-10-15
Allow output targets in Session runner

karllessard:
This solves issue #387 , allowing operation and output names to be passed as session targets.


karllessard:
Thanks @Craigacp

PR #399, opened 2021-11-30
Remove ndarray from the docs generator.

MarkDaoust:
`ndarray` is not part of the repo anymore. 
This is preventing me from updating the docs to v0.4.

We need to either remove it here or add code to this script to clone that repo and copy-in the code.

### DIFF
--- a/tools/build_java_api_docs.py
+++ b/tools/build_java_api_docs.py
@@ -68,8 +68,6 @@ def main(unused_argv):
             merged_source/'java/org/tensorflow')
   shutil.copytree(REPO_ROOT/'tensorflow-framework/src/main/java/org/tensorflow/framework',
                   merged_source/'java/org/tensorflow/framework')
-  shutil.copytree(REPO_ROOT/'ndarray/src/main/java/org/tensorflow/ndarray',
-                  merged_source/'java/org/tensorflow/ndarray')
 
   gen_java.gen_java_docs(
       package='org.tensorflow',


Craigacp:
I think we should clone that repo and merge it in. We want the ndarray docs to be in the main docs as otherwise a bunch of the Javadoc will be dead links (and rather confusing to read at that).

karllessard:
@MarkDaoust , NdArray have been moved here: https://github.com/tensorflow/java-ndarray

They can both be released independently now (e.g. we've just released 0.4.0 of TF Java but NdArray is still at 0.3.3, which is fine). How can we can keep all this in sync on the website?

karllessard (approved):

MarkDaoust:
> They can both be released independently now (e.g. we've just released 0.4.0 of TF Java but NdArray is still at 0.3.3, which is fine). How can we can keep all this in sync on the website?

I'm not exactly sure yet 😅.

IDK anything about java package management. Do you know a good command I can use to check maven  for the latest version of the two packages? If I can automate the version number check I could download the correct branch for both, splice the source together, and generate the docs.

karllessard:
> 
> IDK anything about java package management. Do you know a good command I can use to check maven for the latest version of the two packages? If I can automate the version number check I could download the correct branch for both, splice the source together, and generate the docs.

But wait... thinking of it, TensorFlow speaking, we want to show on tensorflow.org probably the documentation of the NdArray version that is actually used by the Java client and not necessarily the latest one. 

`mvn dependency:tree | grep ndarray` can give you that version. You can then clone the whole NdArray repo and checkout that version using Git. For example, if the first command gives you the following output: 
```
[INFO] +- org.tensorflow:ndarray:jar:0.3.3:compile
[INFO] |  \- org.tensorflow:ndarray:jar:0.3.3:compile
[INFO] |  \- org.tensorflow:ndarray:jar:0.3.3:compile
```
You can checkout that version of NdArray `git checkout tags/v0.3.3 -b branch`. Would that work for you?

Craigacp:
We've updated the script to clone the java-ndarray repo as part of the 1.0 release.

PR #407, opened 2022-01-13
Fix NullPointerException issue.

### LINKED ISSUE #406: NullPointerException when calling matrixDiagV3
The following code works fine with 0.3.1, but now I get NPE in 0.4.0:

```
        EagerSession eagerSession = EagerSession.options().async(true).build();
        Ops tf = Ops.create(eagerSession);
        Operand<TFloat32> diagonal = tf.fill(tf.constant(new long[]{2}), tf.constant(1f));
        Operand<TFloat32> value = tf.fill(tf.constant(new long[]{1}), tf.constant(0f));

        Operand<TFloat32> output =
                tf.linalg.matrixDiagV3(
                        diagonal, tf.constant(0), tf.constant(2), tf.constant(2), value);
        Tensor tensor = output.asTensor(); // exception here
```

I got exception on both macOS and Linux.

```
java.lang.NullPointerException
	at org.tensorflow.EagerOperation.resolveTensorHandle(EagerOperation.java:193)
	at org.tensorflow.EagerOperation.resolveTensor(EagerOperation.java:166)
	at org.tensorflow.EagerOperation.tensor(EagerOperation.java:152)
	at org.tensorflow.Output.asTensor(Output.java:97)
	at org.tensorflow.Operand.asTensor(Operand.java:65)
```

frankfliu:
This fixes #406, With this fix, it will throw a TFInvalidArgumentException instead of NullPointerException.

The error message looks like the following:

org.tensorflow.exceptions.TFInvalidArgumentException: padding_value must be a scalar
	Encountered when executing an operation using EagerExecutor. This error cancels all future operations and poisons their output tensors.

Before the fix, the stacktrace is quite confusing:

java.lang.NullPointerException
	at org.tensorflow.EagerOperation.resolveTensorHandle(EagerOperation.java:193)
	at org.tensorflow.EagerOperation.resolveTensor(EagerOperation.java:166)
	at org.tensorflow.EagerOperation.tensor(EagerOperation.java:152)
	at org.tensorflow.Output.asTensor(Output.java:97)
	at org.tensorflow.Operand.asTensor(Operand.java:65)

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/EagerOperation.java
@@ -190,9 +190,9 @@ class EagerOperation extends AbstractOperation {
     requireTensorHandle(handle);
     try (PointerScope scope = new PointerScope()) {
       TF_Status status = TF_Status.newStatus();
-      TF_Tensor tensor = TFE_TensorHandleResolve(handle, status).withDeallocator();
+      TF_Tensor tensor = TFE_TensorHandleResolve(handle, status);
       status.throwExceptionIfNotOK();
-      return RawTensor.fromHandle(tensor, session).asTypedTensor();
+      return RawTensor.fromHandle(tensor.withDeallocator(), session).asTypedTensor();
     }
   }
 


Craigacp (approved):
LGTM

Craigacp:
Could you put an example of the error message difference as a comment here in case we've got this mistake elsewhere so we can find it with search.

frankfliu:
> Could you put an example of the error message before & after as a Github comment here in case we've got this mistake elsewhere so we can find it with search.

I updated the description

karllessard:
Thanks @frankfliu !

PR #412, opened 2022-02-04
Bumping to TF 2.7.1, protobuf 3.19.4, error-prone 2.10.0

Craigacp:
Bumping to the latest point release of TF.


karllessard:
@Craigacp , it looks like the build passed on all platform, should we merge it now then?

karllessard (approved):

PR #415, opened 2022-02-12
Upgrade for JavaCPP 1.5.7 and TensorFlow 2.8.0

saudet:
Everything builds and all tests pass for me, even though I almost didn't have to update anything, but a lot changes in the ops as usual, so that probably needs to be looked into before we can merge this. @karllessard


Craigacp:
Did the framework tests pass without my PR from today (https://github.com/tensorflow/java/pull/414)? I had assumed it was an overbroad backport that broke the RNG functionality, but if it worked in TF 2.8 then that's even stranger.

saudet:
Yes, all tests pass, including everything from tensorflow-framework as is in this branch, without pull #414. I'm only seeing a "skipped" one from org.tensorflow.framework.optimizers.GradientDescentTest.

saudet:
@karllessard I'm not sure that creating a new "rawops" package with only a couple of ops is better than just leaving everything in "core", but that's what your tool picked up, so I'll leave you deal with this!

karllessard:
Thanks for trying @saudet but I think the API import tooling is not working as good as before since the last few versions of TensorFlow. The number of new ops seems relatively small, I'll go through them and update their package manually.

karllessard:
Ok @saudet , I just pushed the following changes to your branch:

- Upgrade to Bazel 4.2.1 (since that's what TF 2.8.0 is build with)
- Update `tensorflow.bazelrc` with the version found in TF 2.8.0 sources
- Reclassify the new ops

For the latter, I tried quickly to fix the `java_api_import` tool but without to much success, they might have change the structure of their golden API that this tool is parsing, so I reclassify them by hand, which wasn't too bad in this case (only 7 ops).

karllessard:
@saudet , any idea what this error is about? https://github.com/tensorflow/java/runs/5311689706?check_suite_focus=true#step:5:1595

saudet:
> ```
> Auto-Configuration Error: Couldn't find undname.exe under C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC, please check your VC installation and set BAZEL_VC environment variable correctly.
> ```

I don't think TF supports Visual Studio 2022, try to revert to 2019. If that doesn't work, try to revert to the previous version of Bazel since that could as well be caused by a regression in there.

rnett:
> I don't think TF supports Visual Studio 2022, try to revert to 2019.

I doubt the bazel version we're using does either, `4.0.0` didn't for me.

karllessard:
We’ll see if the native build succeeds but. eventually it will fail after when compiling the Java stuff because while we’ve already migrated to jdk 11 in ndarray, we haven’t done it yet in TF Java

saudet:
>```
>  C:\tmp\_bazel_runneradmin\mg3pkz7r\execroot\tensorflow_core_api\external\org_tensorflow\tensorflow\core\common_runtime\isolate_placer_inspection_required_ops_pass.cc : fatal error C1083: Cannot open compiler generated file: '': Invalid argument
>```
That usually means there's a path somewhere that's longer than MAX_PATH. We might need to hack the build by creating a temporary junction for `d:\a\java\java\tensorflow-core\tensorflow-core-api\` to a shorter path like `d:\tf` this way:
https://github.com/bytedeco/javacpp-presets/blob/1.5.7/scipy/cppbuild.sh#L178-L187

karllessard:
@saudet , I wanna see how this PR builds after upgrading to JDK11 (#427) so I've rebased your branch on master.

karllessard (approved):

PR #417, opened 2022-02-13
Add TF Python Config support to deserialize from Models saved with the TF Engine outside of Java. 

JimClarke5:
Added `GELU` activation with support in `FrameworkOpss`

Added maps in `Activation` interface to resolve to an `Activation` object via either a TF Engine activation name or from a `Map<> `config object. TF Python only serializes to the activation name string and does not serialize values passed to activation function, like the `GELU` `approximate` parameter.

Added `getConfig() `to serialize the activation. Added `getName() `to return the TF Python name for the activation.

Added static methods to call the activation directly, for example 
`ReLU.relu(tf, input)` for default settings
`ReLU.relu(tf, input, 2.0f /*alpha*/, 999.0f /*maxValue*/, 0.01f /*threshold*/) `
`Linear.linear(tf, input)`
**Note:** Python allows a function for the activation  method (similar to 'C' function pointer), 
but each function may have its own set of arguments.
It would be nice to handle this in Java, but I could not find a good way to do this with a `java.lang.FunctionalInterface`.


tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -14,24 +14,104 @@
...
+        }
+      };
+
+  /**
+   * Creates an Activation based on a name as known to the TensorFlow engine.
+   *
+   * @param name the activation name
+   * @return the Activation
+   */
+  static Activation create(String name) {
```
Craigacp, line 72: The doc should note that it defaults to linear if the name is null and throws IAE if the name is unknown.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -14,24 +14,104 @@
...
+   *     <pre>{@code
+   * "name" : String - this is used to locate the class, this is the TensorFlow Engine's class name
+   *
+   * }</pre>
+   *
+   * @return the Activation
+   */
+  static Activation create(Map<String, Object> config) {
+    String className = config == null ? null : (String) config.get("name");
+    if (className == null) {
```
Craigacp, line 96: Is there a reason we want this to default to linear rather than throwing IAE for both null and unknown names?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -42,19 +44,21 @@
...
+public class ELU extends AbstractActivation {
+
+  /** The activation name as known by TensorFlow */
+  public static final String NAME = "elu";
 
   private static final double ALPHA_DEFAULT = 1.0;
 
-  /** A scalar, slope of negative section. */
-  private final double alpha;
+  private double alpha;
```
Craigacp, line 61: Why are the activations mutable now? (Also its lost the javadoc).

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -72,16 +76,96 @@ public ELU(double alpha) {
...
-  /** {@inheritDoc} */
-  @Override
-  public Operand<T> call(Ops tf, Operand<T> input) {
+  /**
+   * Creates a new ELU from a configuration Map
+   *
+   * @param config the configuration map, if the map contains an entry for {@code alpha} that value
+   *     is used, otherwise {@link #ALPHA_DEFAULT} is used.
+   */
+  public ELU(Map<String, Object> config) {
```
Craigacp, line 85: What happens if you supply the configuration for e.g. `ReLU` to this constructor? I would expect that to be an error, but I think in this case it would silently ignore it and default alpha. That doesn't seem like ideal behaviour.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
```diff
@@ -72,16 +76,96 @@ public ELU(double alpha) {
...
+   *
+   * <p><code>x</code> if <code>x &gt; 0</code> and <code>alpha * (exp(x) -
+   *  * 1)</code> if <code>x &lt; 0</code>.
+   *
+   * <p>The ELU hyperparameter <code>alpha</code> controls the value to which an ELU saturates for
+   * negative net inputs. ELUs diminish the vanishing gradient effect.
+   *
+   * <p>ELUs have negative values which pushes the mean of the activations closer to zero. Mean
+   * activations that are closer to zero enable faster learning as they bring the gradient closer to
+   * * the natural gradient. ELUs saturate to a negative value when the argument gets smaller.
```
Craigacp, line 102: I think the `*` got doubled here, and in one of the code sections above).

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/AbstractActivation.java
```diff
@@ -14,18 +14,44 @@
...
   protected AbstractActivation() {}
 
+  /**
+   * Gets a configuration map, this default implementation returns a singleton Map, with {@code
+   * "name"} as the key, and the @code name} parameter as its value;
+   *
+   * @param name the name of the Activation as known by TensorFlow engine.
+   * @return the configuration map
+   */
+  protected Map<String, Object> getConfig(String name) {
```
Craigacp, line 37: Should this be no-args and call `getName()` to get the name?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
```diff
@@ -31,16 +32,59 @@
...
+   *      Operand&lt;TFloat32&gt; input = &#46;&#46;&#46;;
+   *      Operand&lt;TFloat32&gt; result = Exponential.exponential(tf, input);
+   * </pre>
+   *
+   * @param tf the TensorFlow Ops
+   * @param input the input
+   * @param <T> the data type for the input
+   * @return the exponential activation: {@code exp(x)}.
+   */
+  public static <T extends TNumber> Operand<T> exponential(Ops tf, Operand<T> input) {
```
Craigacp, line 69: Why is the bound `TNumber` rather than `TFloating`?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/GELU.java
```diff
@@ -0,0 +1,157 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 1: Copyright 2022?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
```diff
@@ -93,9 +95,66 @@ public ReLU(float alpha, float maxValue, float threshold) {
...
+   *           #ALPHA_DEFAULT} is used.
+   *       <li>if the map contains an entry for {@code max_value} that value is used, otherwise
+   *           {@link #MAX_VALUE_DEFAULT} is used.
+   *       <li>if the map contains an entry for {@code threshold} that value is used, otherwise
+   *           {@link #THRESHOLD_DEFAULT} is used.
+   *     </ul>
+   */
+  public ReLU(Map<String, Object> config) {
+    this(
+        ((Number) config.getOrDefault("alpha", ALPHA_DEFAULT)).floatValue(),
```
Craigacp, line 113: Why cast to `Number` and get floatValue rather than casting to `Float`?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -14,24 +14,104 @@
...
+          put("relu", ReLU::new);
+          put("selu", SELU::new);
+          put("sigmoid", Sigmoid::new);
+          put("softmax", Softmax::new);
+          put("softplus", Softplus::new);
+          put("softsign", Softsign::new);
+          put("swish", Swish::new);
+          put("tanh", Tanh::new);
+        }
+      };
```
karllessard, line 64: Why not using Enums instead of Strings? I think it's more idiomatic to Python to use strings everywhere like this, but using Enums in Java would be relieving, if it's not just for the sake of auto-completeness.

We just need to make sure that our Enums are named based on these labels so we can match back a String from Keras config if needed (or we add an attribute to the enums that matches this label).

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
karllessard: Or maybe change the scope of this constructor to be protected or package-private?

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
JimClarke5: New code throws NPE if null, else throws IAE if name is not a known activation

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ELU.java
JimClarke5: I just did a clean up of the java doc for activations

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/AbstractActivation.java
JimClarke5: `getConfig(name)` is a helper that creates a singleton Map with a "name" key.
while` getConfig() `is an abstract method that may create a more complex config Map.
Perhaps change to `getDefaultConfig` and populate it using `getName()`.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
JimClarke5: TF Python, tf.math.exp requires floating, so I will fix.

JimClarke5: Using `TNumber` works, but `TFloating` creates an error.
`'exponential(org.tensorflow.op.Ops, org.tensorflow.Operand<T>)' in 'org.tensorflow.framework.activations.Exponential' cannot be applied to '(org.tensorflow.op.Ops, org.tensorflow.Operand<T>)'`

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/ReLU.java
JimClarke5: The number is converted to Double in the (GSON)  json parser.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
JimClarke5: Changed to enums

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Exponential.java
JimClarke5: Actually the problem here is that `public <T extends TNumber> Operand<T> call(Ops tf, Operand<T> input) `
is part of the functional interface for all activations. I have this method calling the underlying static method, `exponential()`, using the state from the Exponential class. So if the static method,`exponential()`, is `TFloating`, then it is out of sync with `TNumber` in the `call() `method. 

For now, I have the  `call()` method invoking `tf.math.exp()` directly,  the way it used to be,  and I changed the signature for `exponential()` to use `TFloating`.

JimClarke5:
@karllessard @Craigacp  I just made the changes to bring this branch into sync with master. It is not ready to be reviewed.

tensorflow-framework/src/main/java/org/tensorflow/framework/activations/Activation.java
```diff
@@ -14,24 +14,137 @@
...
+   * @param <T> the data type of the input and the result
    */
-  Operand<T> call(Ops tf, Operand<T> input);
+  <T extends TNumber> Operand<T> call(Ops tf, Operand<T> input);
+
+  /**
+   * The Enumerations for creating Activations based an activation name, with either an empty
+   * constructor or a constructor that takes a Map object that contains the Activation's state.
+   */
+  enum ActivationType {
```
karllessard, line 76: Instead of having `Activation.ActivationType`, let's move this to an outer class called `Activations`

JimClarke5: OK

PR #431, opened 2022-03-11
Unpack native artifact in dev mode

karllessard:
This modification prevents having the following error when developing locally in IntelliJ (and maybe other IDEs) in dev mode: `Could not find jnitensorflow in class, module, and library paths.` 

The issue was happening when inside the same project, you add `tensorflow-java` as a module plus other modules that want to use it, like test applications. Since IntelliJ finds `tensorflow-core-api` inside the same project structure, it only relies on the binaries found under its `target` build directory and ignore any of its Maven artifact, including the native one we were depending on to run the unit tests with the `dev` profile.

So instead of adding a dependency to the native jar, we now unpack it under `target/native` to retrieve the same binary structure as we would have if we would do a full native build.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/pom.xml
+++ b/tensorflow-core/tensorflow-core-api/pom.xml
@@ -78,8 +78,8 @@
   <profiles>
     <!--
     Developing Profile
-      This profile skips the tedious native build for local development, fetching the artifacts as a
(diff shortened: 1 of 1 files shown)


Craigacp (approved):
LGTM

PR #439, opened 2022-04-04
Fix broken @see blocks.

MarkDaoust:
@see lines try to link the whole line to the name, it tries to lookup these URLs as names in the API and fails:

`@see name description`

https://www.tensorflow.org/jvm/api_docs/java/org/tensorflow/DeviceSpec

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DeviceSpec.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/DeviceSpec.java
@@ -7,7 +7,7 @@ import java.util.Objects;
  *
  * <p>This is a Java port from python device_spec.py.
  *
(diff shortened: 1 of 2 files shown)


Craigacp (approved):
LGTM

karllessard:
Thanks @MarkDaoust! Before merging, is your intention is to generate the documentation for 0.4.1 that we released last week? If so, the doc must be generated from the`r0.4` branch, just want to check if we need to merge this PR in that branch instead.

MarkDaoust:
I was just planning on "eventual consistency", let's not make this more complicated than it needs to be.

karllessard:
Well, looks like it will be more complicated than we were hoping for anyway, because Google's Java style checker did not liked it 😄 (see [failing status](https://github.com/tensorflow/java/runs/5820981329?check_suite_focus=true) in GitHub)

what about keeping `@see` and just passing the straight URL, with no markup nor label?

PR #448, opened 2022-05-13
Add more factory methods for TFE_TensorHandle

### LINKED ISSUE #447: [Bug] TensorFlow JavaCPP GPU memory leak
During the investigation from our customer issue: https://github.com/deepjavalibrary/djl/issues/1552, we found that TensorFlow JavaCPP may have memory leak while copying from CPU to GPU devices. We managed to build a minimum reproducible code in the attachment. We appreciate for your help! :)

## How to reproduce?

All you need to do, is to find a GPU device, and run the followings after you download the attachment:

```
./gradlew run
```
and you will be able to reproduce this in less than 45 seconds.

We tested on the release of 0.4.0

## Problematic point

The copyToDevice operation will produce a new TensorHandle from the original one. However, it seemed not closed even if you do .close() operation on it. .deallocate() also doesn't help here.

```
        try (PointerScope ignored = new PointerScope()) {
            TF_Status status = TF_Status.newStatus();
            TFE_TensorHandle newHandle =
                tensorflow.TFE_TensorHandleCopyToDevice(
                    handle, eagerSessionHandle, deviceName, status);
            status.throwExceptionIfNotOK();
            return newHandle.retainReference();
        }
```

## Does CPU also have this issue?

No, we tested CPU with close for more than 3 hours and did not observe a significant memory increment.

## Output logs and test setup

You can run the project under any instances with GPU. After 13k rounds, the process will crash with OOM on GPU.

[output.log.zip](https://github.com/tensorflow/java/files/8673374/output.log.zip)
[TestTF.zip](https://github.com/tensorflow/java/files/8673375/TestTF.zip)

saudet:
Specifically for TFE_TensorHandleCopyToDevice() and TFE_TensorHandleCopySharingTensor()

Fixes #447

/cc @lanking520

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/AbstractTFE_TensorHandle.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/c_api/AbstractTFE_TensorHandle.java
@@ -1,19 +1,19 @@
 /*
- Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-     http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

Craigacp (approved):
LGTM

Craigacp:
Can you run spotless on it so it passes the formatting check?

saudet:
> Can you run spotless on it so it passes the formatting check?

I didn't make any changes to the parts it's complaining about. If we want to "fix" this, it should be run against all the source code. Having only some files "correctly" formatted isn't helpful.

Craigacp:
It's set up to be incremental, every time we touch a file that hasn't been formatted spotless requires the file to be formatted. It means we don't need one massive commit that formats everything and trashes all the history. So please apply the formatting to that file.

Craigacp (approved):
LGTM.

Issue #461, opened 2022-06-14
Performance Issue when upgrading from 0.3.3 to 0.4.1


mattdornfeld:
<em>Please make sure that this is a bug. As per our [GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md), we only address code/doc bugs, performance issues, feature requests and build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
Code is based one examples in tests found in Tensorflow 0.3.3
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04 x86\_64):
Method profiling is done on MacOS 12.4 x86_64. Resource profiling is done on Ubuntu 18.04 x86_64
- TensorFlow installed from (source or binary):
Installed from binary using Gradle
- TensorFlow version (use command below):
Tensorflow 2.4.1
- Java version (i.e., the output of `java -version`):
Compiled with Java 11 run with Java 17
- Java command line flags (e.g., GC parameters):
-XX:+UseG1GC
- Python version (if transferring a model trained in Python):
Python 3.7

**Describe the current behavior**
I'm trying to upgrade from Tensorflow Java 0.3.3 to 0.4.1 and am seeing significant performance degradation when doing the upgrade. I've done a good amount of profiling to help diagnose the issue and I was hoping someone might be able to help me understand what's going on. On a high level these performance tests are constructing a HashMap of tf.Tensor objects from input data and feeding them to a trained Logistic Regression model for inference. Note that this profiling shows a downgrade from 0.4.1 back to 0.3.3 (so earlier is the newer version and later is the older versioon)

When downgrading from 0.4.1 to 0.3.1 I see a significant decrease in p50 latency (measured in ms)
![image](https://user-images.githubusercontent.com/4313686/173674740-25e0218f-85e1-48fc-ae3b-011e28264a1a.png)

I also see increased CPU usage
![image](https://user-images.githubusercontent.com/4313686/173674802-2d2fafa8-5b73-4c2b-b94b-a907ab01881c.png)

Decreased memory usage
![image](https://user-images.githubusercontent.com/4313686/173674841-f5df307a-7601-4066-be5a-c3a1d7543cd2.png)

And increased Young Generation Garbage collection
![image](https://user-images.githubusercontent.com/4313686/173674909-7379926f-b4aa-42da-9938-ce72bfd49340.png)

[...]

PR #466, opened 2022-07-15
Simplify graph initializers

karllessard:
This PR simplifies how TF Java handles automatic graph initialization. Actual implementation might be too complex for what users need and, as we see in https://github.com/tensorflow/java/issues/461, might also be a cause of additional latency when running model inference.

In a nutshell, the refactoring consists of:
- removing endpoints or functionalities that we do not judge essential for the users
- preserve a single 'Init' op for running all initializations, by creating or mutating its protodef on export
- limit overhead of checking the initialization state of a session when running inference

CC\ @rnett

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
@@ -516,7 +516,7 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
     importGraphDef(graphDef, "");
   }
 
-  static final String INIT_OP_BASE_NAME = "Init";
+  static final String INIT_OP_NAME = "Init";
 
   /**
    * Import a representation of a TensorFlow graph.
@@ -539,12 +539,12 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
     String initPrefix;
     if (!prefix.isEmpty()) {
       if (prefix.endsWith("/")) {
-        initPrefix = prefix + INIT_OP_BASE_NAME;
+        initPrefix = prefix + INIT_OP_NAME;
       } else {
-        initPrefix = prefix + "/" + INIT_OP_BASE_NAME;
+        initPrefix = prefix + "/" + INIT_OP_NAME;
       }
     } else {
-      initPrefix = INIT_OP_BASE_NAME;
+      initPrefix = INIT_OP_NAME;
     }
 
     operations()
@@ -1065,8 +1065,8 @@ public final class Graph implements ExecutionEnvironment, AutoCloseable {
       return graphDef;
     }
     var graphDefWithInitBuilder = graphDef.toBuilder();
-    var initNode = graphDefWithInitBuilder.getNodeBuilderList().stream().filter(n -> n.getName().startsWith(INIT_OP_BASE_NAME)).findFirst().orElseGet(() -> {
-      return graphDefWithInitBuilder.addNodeBuilder().setName(INIT_OP_BASE_NAME).setOp(NoOp.OP_NAME);
+    var initNode = graphDefWithInitBuilder.getNodeBuilderList().stream().filter(n -> n.getName().equals(INIT_OP_NAME)).findFirst().orElseGet(() -> {
+      return graphDefWithInitBuilder.addNodeBuilder().setName(INIT_OP_NAME).setOp(NoOp.OP_NAME);
     });
     initializers.stream().skip(newInitializersMarker).forEach(op -> initNode.addInput("^" + op.name()));
     return graphDefWithInitBuilder.build();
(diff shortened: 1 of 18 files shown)


karllessard:
The PR is in a draft state as I would like to add more unit tests + clarify a few points with Ryan (original author of the actual implementation of the initializers) before merging.

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/processor/operator/OperatorProcessor.java
```diff
@@ -562,25 +562,7 @@ private static TypeSpec buildTopClass(OpsSpec spec) {
...
-            .addModifiers(Modifier.PUBLIC)
-            .addParameter(T, "op")
-            .returns(T)
-            .addStatement("scope.env().registerInitOp(op.op())")
-            .addStatement("return op")
-            .addJavadoc(
-                "Make {@code op} an init operation, doing the same for all of it's inputs (and control inputs).\n"
-                    + initScopeComment
-                    + "\n@see ExecutionEnvironment#registerInitOp(Operation)\n"
-                    + "\n@throws IllegalStateException if the op or one of its inputs can't be made an init op.")
```
karllessard, line 583: @rnett I took the liberty of removing this endpoint, assuming that if a user is building a graph and wants an op to be added as a initializer, then he should simply use the `withInitScope()` when creating this op. Wdyt?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
```diff
@@ -591,11 +550,13 @@ private static SavedModelBundle fromHandle(
           .getValueList()
           .forEach(
               node -> {
-                graph.registerInitOp(graph.operationOrThrow(node));
+                graph.registerInitializer(graph.operationOrThrow(node), false);
               });
     }
 
-    return new SavedModelBundle(graph, session, metaGraphDef, functions);
+    return new SavedModelBundle(graph, session, metaGraphDef, functions.entrySet().stream()
```
Craigacp, line 557: Why'd this transformation move out of the constructor?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -1091,6 +1060,18 @@ private static GraphDef toGraphDef(TF_Graph handle) {
     }
   }
 
+  private GraphDef addOrUpdateInit(GraphDef graphDef) {
+    if (newInitializersMarker < 0) {
+      return graphDef;
+    }
+    var graphDefWithInitBuilder = graphDef.toBuilder();
+    var initNode = graphDefWithInitBuilder.getNodeBuilderList().stream().filter(n -> n.getName().startsWith(INIT_OP_BASE_NAME)).findFirst().orElseGet(() -> {
```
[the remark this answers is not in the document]
karllessard: The first one will collect all new initializers that has been registered, which should be enough I think. But maybe we should try to *match exactly* a single op called `Init` and if there are none, we create it?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/SavedModelBundle.java
karllessard: Bah, just a little preference of keeping constructors as simple as possible and moving any logic to the caller (so if that logic throws an exception, it won't be thrown from the constructor). I really don't mind reverting this, don't even recalled changing it :)

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -1091,6 +1059,28 @@ private static GraphDef toGraphDef(TF_Graph handle) {
...
+            .orElseGet(
+                () -> {
+                  return graphDefWithInitBuilder
+                      .addNodeBuilder()
+                      .setName(INIT_OP_NAME)
+                      .setOp(NoOp.OP_NAME);
+                });
+    initializers.stream()
+        .skip(newInitializersMarker)
+        .forEach(op -> initNode.addInput("^" + op.name()));
```
Craigacp, line 1080: What's the `^` mean in this case?

karllessard: This is how you mark an input as a control dependency in the proto file. But I agree that is worth commenting it out :)

Craigacp: Sounds good, and yes it definitely needs a comment.

Craigacp (approved):
LGTM

Issue #429, opened 2022-03-09
Version 0.4.0 not loading models from S3


dennisTS:
Dear TF team,

we are experiencing a weird issue with the latest version of TensorFlow for Java - after upgrade it stopped loading models from S3. 

Here's an **example setup**:

Java 8, a single dependency:
```
    <dependency>
      <groupId>org.tensorflow</groupId>
      <artifactId>tensorflow-core-platform</artifactId>
      <version>0.4.0</version>
    </dependency>
```

A single class:

```
import org.tensorflow.SavedModelBundle;

public class App {

  public static void main(String[] args) {
    SavedModelBundle.load("s3://some/location", "serve");
  }
}
```

Environment variable: `AWS_REGION=eu-central-1`

**Outcome**:
```
2022-03-09 20:53:24.893857: I external/org_tensorflow/tensorflow/cc/saved_model/reader.cc:43] Reading SavedModel from: s3://some/location
2022-03-09 20:53:24.893896: I external/org_tensorflow/tensorflow/cc/saved_model/loader.cc:283] SavedModel load for tags { serve }; Status: fail: NOT_FOUND: Could not find SavedModel .pb or .pbtxt at supplied export directory path: s3://some/location. Took 49 microseconds.
Exception in thread "main" org.tensorflow.exceptions.TensorFlowException: Could not find SavedModel .pb or .pbtxt at supplied export directory path: s3://some/location
	at org.tensorflow.internal.c_api.AbstractTF_Status.throwExceptionIfNotOK(AbstractTF_Status.java:101)
	at org.tensorflow.SavedModelBundle.load(SavedModelBundle.java:623)
	at org.tensorflow.SavedModelBundle.access$000(SavedModelBundle.java:67)
	at org.tensorflow.SavedModelBundle$Loader.load(SavedModelBundle.java:97)
	at org.tensorflow.SavedModelBundle.load(SavedModelBundle.java:357)
	at App.main(App.java:6)
```

Either of the changes below "fixes" the issue:
* changing from S3 to local
* changing from `0.4.0` to `0.3.3`

**Question**: has anything changed since the previous version? Are we missing some important point that would make integration with S3 possible?

Thank you!

Craigacp:
[...]

PR #468, opened 2022-08-19
Adding a hook for loading TF IO native libraries

Craigacp:
This exposes the C API method to load in a TF IO filesystem plugin. We can't test it without adding a bunch of binaries to the build, but it seems to load TF IO 2.9 without incident. I checked it by loading `<python-3.8-venv>/lib/python3.8/site-packages/tensorflow_io/python/ops/libtensorflow_io_plugins.so`.

This will help users with issues like #429, though they'll still need to download and unpack TF IO separately.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
@@ -1,4 +1,4 @@
-/* Copyright 2019-2021 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2019-2022 The TensorFlow Authors. All Rights Reserved.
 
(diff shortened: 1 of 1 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
```diff
@@ -108,6 +109,20 @@ public static OpList loadLibrary(String filename) {
     }
   }
 
+  /**
+   * Loads the filesystem plugin from fielname and registers all the filesystems it supports.
```
karllessard, line 113: fielname -> filename

karllessard (approved):
Approved but there is a small typo in the comments, feel free to merge it right away after fixing it

karllessard (approved):

PR #469, opened 2022-08-31
Add TUint16 type

joeyearsley:
In medical imaging lots of the datatypes are Uint16, however, to decode pngs to this type we first need to add the Type definition. 

This MR adds the type definition to the core API such that users don't need to re-implement this type in their pkgs.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TUint16Mapper.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TUint16Mapper.java
@@ -1,0 +1,136 @@
+/*
+ *  Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *  =======================================================================
+ */
+package org.tensorflow.internal.types;
+
+import org.bytedeco.javacpp.PointerScope;
+import org.tensorflow.RawTensor;
+import org.tensorflow.SparseTensor;
+import org.tensorflow.TensorMapper;
+import org.tensorflow.internal.buffer.TensorBuffers;
+import org.tensorflow.ndarray.buffer.ByteDataBuffer;
+import org.tensorflow.ndarray.impl.dense.ByteDenseNdArray;
+import org.tensorflow.ndarray.impl.sparse.ByteSparseNdArray;
+import org.tensorflow.proto.framework.DataType;
+import org.tensorflow.types.TInt64;
+import org.tensorflow.types.TUint16;
+
+/**
+ * Maps memory of {@link org.tensorflow.proto.framework.DataType#DT_Uint16} tensors to a
+ * n-dimensional data space.
+ */
+public final class TUint16Mapper extends TensorMapper<TUint16> {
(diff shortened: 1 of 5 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TUint16Mapper.java
```diff
@@ -0,0 +1,136 @@
...
+    return new DenseTUint16(tensor, buffer);
+  }
+
+  @Override
+  protected SparseTensor<TUint16> mapSparse(
+      TInt64 indices, TUint16 values, TInt64 denseShape, PointerScope tensorScope) {
+    return new TUint16Mapper.SparseTUint16(indices, values, denseShape, tensorScope);
+  }
+
+  private static final class DenseTUint16 extends ByteDenseNdArray implements TUint16 {
```
Craigacp, line 49: I think this should probably extend `ShortDenseNdArray` rather than `ByteDenseNdArray`, with similar replacements elsewhere in the code.

Craigacp (dismissed):

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TUint16Mapper.java
```diff
@@ -0,0 +1,136 @@
+/*
+ *  Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
joeyearsley: done

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/internal/types/TUint16Mapper.java
joeyearsley: Thanks for catching this, I've updated all byte references to shorts.

Craigacp:
Looks fine to me. @karllessard do you want to have a quick look through?

karllessard:
Yep just did, looks good, thanks a lot @joeyearsley ! 

Though the CI job is failing, one of the unit test did not pass:

karllessard:
@joeyearsley , I took a look and strangely, `tf.math.add` does not support UINT16. To make the test pass, can you change it to `tf.math.sub` (which supports all types) and update the expected results accordingly?

joeyearsley:
Thanks @karllessard , I've ran `mvn spotless:apply` and altered the `add` to `sub` along with the expected results. 
I'm having trouble running the tests on my laptop so I apologise in advance if there are any silly mistakes.

karllessard:
Oh, @joeyearsley you have one unused import now in `NumericTypesTestBase.java` and spotless complains about it. After that, I think will be good to go!

```
Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.20.2:check (default-cli) on project tensorflow-core-api: The following files had format violations:
Error:      src/test/java/org/tensorflow/types/NumericTypesTestBase.java
Error:          @@ -26,7 +26,6 @@
Error:           import?org.tensorflow.ndarray.index.Indices;
Error:           import?org.tensorflow.op.Ops;
Error:           import?org.tensorflow.op.core.Constant;
Error:          -import?org.tensorflow.op.math.Add;
Error:           import?org.tensorflow.op.math.Sub;
Error:           import?org.tensorflow.types.family.TNumber;
Error:           
Error:  Run 'mvn spotless:apply' to fix these violations.
Error:  -> [Help 1]
```

joeyearsley:
Thanks @karllessard, that should do it now.

Out of interest, do either of you build on a Mac and have you ever come across the following build issue:
```
external/jsoncpp_git/src/lib_json/json_value.cpp:1385:35: error: exception specification in declaration does not match previous declaration
Value::Comments& Value::Comments::operator=(Comments&& that) noexcept {
```

It seems like a C++ version issue but not my area of expertise to easily pinpoint a solution.

Craigacp:
For Java side work like this you should run the builds with the maven dev profile (`-Pdev`) which downloads the snapshot binaries rather than building them. Building the native binary takes ages, and doesn't work too well on M1 or M2 Macs.

karllessard (approved):
Thanks @joeyearsley !

Issue #252, opened 2021-03-24
Does TensorFlow-Java support Apple Silicon Macs?


Taiyx:
<em>Please make sure that this is a bug. As per our [GitHub Policy](https://github.com/tensorflow/tensorflow/blob/master/ISSUES.md), we only address code/doc bugs, performance issues, feature requests and build/installation issues on GitHub. tag:bug_template</em>

**System information**
- Have I written custom code (as opposed to using a stock example script provided in TensorFlow):
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): **MacOs Big Sur ver11.0.1,  M1 slices** 
- Mobile device (e.g. iPhone 8, Pixel 2, Samsung Galaxy) if the issue happens on mobile device:
- TensorFlow installed from (source or binary):  Maven
- TensorFlow version (use command below): Old tensorflow version(1.4.0/1.5.0), and new  tensorflow Java Version 0.2.0
- JVM version:  1.8.0_162
-  No GPU

**Describe the current behavior**
I just try to run tensorflow java offical example, and get tensorflow version for test. but It dosen't work. 
I have test different versions of tensorflow java interface, and only ver 1.13.1 works well. 
And all other versions can not work,  for example old tensorflow version(1.4.0/1.5.0), and new  tensorflow Java Version 0.2.0/0.3.0(tensorflow ver2.3.1/2.4.1) .

The Error shows below:

> A fatal error has been detected by the Java Runtime Environment:
> 
> SIGILL (0x4) at pc=0x00000001290edc15, pid=6333, tid=0x0000000000001a03
> 
> JRE version: Java(TM) SE Runtime Environment (8.0_162-b12) (build 1.8.0_162-b12)
> Java VM: Java HotSpot(TM) 64-Bit Server VM (25.162-b12 mixed mode bsd-amd64 compressed oops)
> Problematic frame:
> C  [libtensorflow_framework.2.dylib+0x14c15]  tensorflow::monitoring::MetricDef<(tensorflow::monitoring::MetricKind)1, long long, 2>::MetricDef<char [11], char [7]>(absl::lts_2020_02_25::string_view, absl::lts_2020_02_25::string_view, char const (&) [11], char const (&) [7])+0x125
> 
> Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
> 
> An error report file with more information is saved as:
>  /***/tf_test/hs_err_pid6333.log
> 
[...]

PR #481, opened 2022-12-16
Upgrade to TF 2.10.1, with macosx-arm64 support

karllessard:
This PR does (a bit unfortunately) two different things:

- Upgrade to TensorFlow 2.10.1
- Enable building from sources on Apple Silicon machines (M1)

I tried to upgrade to the now released TF 2.11.0 but hit a bunch of new problems which I want to deal with later, let's get at least MacOX ARM64 build working with a still recent version of TensorFlow (2.10.1)

Addresses #394 and #252


tensorflow-core/pom.xml
```diff
@@ -141,6 +143,7 @@
         <javacpp.platform.linux-x86>${javacpp.platform}</javacpp.platform.linux-x86>
         <javacpp.platform.linux-x86_64>${javacpp.platform}</javacpp.platform.linux-x86_64>
         <javacpp.platform.macosx-x86_64>${javacpp.platform}</javacpp.platform.macosx-x86_64>
+        <javacpp.platform.macosx-arm64>${javacpp.platform}</javacpp.platform.macosx-arm64>
```
saudet, line 146: It'd be more consistent with the linux-* ones to put anything macosx-arm64 before macosx-x86_64.

CONTRIBUTING.md
```diff
@@ -61,6 +60,17 @@ To build for GPU, pass `-Djavacpp.platform.extension=-gpu` to maven. By default,
 for more info. If you add `bazelrc` files, make sure the `TF_CUDA_COMPUTE_CAPABILITIES` value in them matches the value set elsewhere, as it will take
 precedence if present.
 
+### Apple Silicon
+
+The TensorFlow Java project relies on [GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)
+to build and distribute the native binaries for TensorFlow. Unfortunately until this day, GitHub Actions still does not support runners with a
```
Craigacp, line 66: "until this day" -> "at the moment"?

CONTRIBUTING.md
```diff
@@ -61,6 +60,17 @@ To build for GPU, pass `-Djavacpp.platform.extension=-gpu` to maven. By default,
...
+Apple Silicon chip (such as M1). Therefore, we cannot distribute the binaries for this platform, so they must be compiled and installed locally on such systems.
+
+Please follow the present [procedure](CONTRIBUTING.md#building) for building TensorFlow Java from sources.
+
+:warning: Until this day (12-16-2022), TensorFlow fails to build on XCode command line tools version 14+. If you have such version installed, it might
```
Craigacp, line 71: "Until this day (12-16-2022)" -> "As of 12-16-2022"

tensorflow-core/tensorflow-core-api/build.sh
```diff
@@ -16,12 +17,17 @@ else
...
 fi
 
+# Add platform specific flags
+if [[ "${PLATFORM:-}" == macosx-arm64 ]]; then
+  BUILD_FLAGS="$BUILD_FLAGS --config=macos_arm64"
+fi
+
 if [[ "${EXTENSION:-}" == *mkl* ]]; then
-    export BUILD_FLAGS="$BUILD_FLAGS --config=mkl"
+    BUILD_FLAGS="$BUILD_FLAGS --config=mkl"
```
Craigacp, line 26: Should we strip all the MKL bits at this point?

tensorflow-core/tensorflow-core-api/src/bazel/op_generator/op_export_main.cc
```diff
@@ -93,32 +93,30 @@ int main(int argc, char* argv[]) {
   tensorflow::port::InitMain(tensorflow::java::kUsageHeader, &argc, &argv);
   std::vector<tensorflow::string> api_dirs;
 
-  if(argc < 4) {
-      std::cerr << "Must specify <library_path> <binary_output> <text_output>" << "\n";
+  if(argc < 3) {
```
Craigacp, line 96: Why isn't this an equality check?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
```diff
@@ -8097,6 +8156,45 @@ public While whileOp(Iterable<Operand<?>> input, ConcreteFunction cond, Concrete
...
+   * @param args A list of {@code Tensor} with possibly different types to be passed as arguments
+   *  to the HLO module.
+   * @param module A serialized computation, a text representation of mlir.Module.
+   * @param Sout List of output tensor shapes.
+   * @param Tout List of output tensor data types.
+   * @param dimArgsSpec the specification for the dimension arguments, one for each
+   *  dimension argument. In absence of dynamic shapes this list is empty.
+   * @return a new instance of XlaCallModule
+   */
+  public XlaCallModule xlaCallModule(Iterable<Operand<?>> args, String module, List<Shape> Sout,
```
Craigacp, line 8193: Given the big warnings all over the docs for this op, should it be visible?

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/Node.java
```diff
@@ -159,5 +159,5 @@ public class Node extends Pointer {
 
   // Called after an incident non-control edge has changed. Does nothing if not
   // all input edges are defined.
-  public native void RunForwardTypeInference();
+
```
Craigacp, line 162: Why'd the method disappear but not the comment?

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/core/ApproxTopK.java
```diff
@@ -0,0 +1,326 @@
+/* Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
Craigacp, line 1: Do we auto insert the copyright year when generating? If so, maybe it's worth updating the year to be "2018 - <current year>".

Craigacp:
As 2.10/2.11 get merged in we might want to think about how to expose the new model fingerprint support, and if we can generate similar fingerprints from TF-Java - https://github.com/tensorflow/community/blob/master/rfcs/20220610-saved-model-fingerprinting.md.

karllessard:
Linux GPU was failing to build, looks like this is a known problem and I've applied a patch found [here](https://github.com/tensorflow/tensorflow/issues/57826#issuecomment-1328911560) that might fix it, we'll see.

tensorflow-core/tensorflow-core-api/build.sh
karllessard: As being discussed [here](https://github.com/tensorflow/java/issues/394#issuecomment-1354030460), I'm planning to do that in a separate PR, since this one already include enough changes.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java
karllessard: Good catch!

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/internal/c_api/Node.java
karllessard: I think that's what happen when you skip a method in JavaCPP preset, I had to do this with this one since the symbol was found nowhere in the native libs, see line 327 in `tensorflow.java` patch.

tensorflow-core/tensorflow-core-api/src/gen/java/org/tensorflow/op/core/ApproxTopK.java
karllessard: Yep looks like that coming from us, I've updated it. Also I did the same change for the `Op` annotation processor (the one that generates the `*Ops.java` classes)

Craigacp (approved):
LGTM assuming we get a clean CI build.

karllessard:
There is an issue with the `java_op_exporter` in Linux that does not find the TF shared libraries now that I'm linking to it. Should be easy to fix with `LD_LIBRARY_PATH`, I'll test it out and push new changes

karllessard:
Ok, works in Linux and MacOS now, but for Windows I need to merge it first so the builds can continue over 6 hours. I'll repair after if needed.

Issue #482, opened 2022-12-29
Failure when a tensorflow variable has no gradients


ramon-garcia:
Tensorflow java fails when a variable is hidden from gradients using tf.stopGradient.

Variables without gradients are used to have persistent state between runs. In my case, I want to have some customizable values in the graph. With a constant, these values cannot be changed. For instance:

For instance, I want to normalize input values so that their square deviation is (more or less) 1. So one introduces a reasonable value of square deviation via a Tensorflow variable defined with

var signalVar = tf.withName("signalVar").variable(tf.constant(1.f));

it is initialized as 1, but can be modified with variable assignment (omitted here). Then use it to divide the input signal. But we don't want to train it.

var signalNormalized = tf.math.div(signal, tf.stopGradient(signalNormalized));

Then, when calling optimizer.minimize an IllegalStateException "close() has been called on the Graph this Operation was a part of".

var optimizer = new Adam(graph, learningRate, betaOne, betaTwo, epsilon);
optimizer.minimize(loss, "train"); // IllegalStateException throw here.

The reason is that when the optimizer code scans for variables, it checks the gradients and their data types, but a variable protected with stopGradient() has no gradient TF_Operation initialized.

PR #483, opened 2022-12-29
Fix fail when a variable receives zero gradient  #482

ramon-garcia:
Fixes bug #482 . A check is added in the optimizer code that the gradient of a variable is valid, and otherwise it is left out.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
@@ -147,7 +147,10 @@ public final class Output<T extends TType> implements Operand<T> {
     return operation.getUnsafeNativeHandle(index);
   }
 
-  /** Returns whether the underlying operation has no valid handle. Makes the opposite check as GraphOperation.requireHandle **/
+  /**
+   * Returns whether the underlying operation has no valid handle. Makes the opposite check as
+   * GraphOperation.requireHandle *
+   */
   public boolean isClosed() {
     Pointer handle = operation.getUnsafeNativeHandle(index);
     return handle == null || handle.isNull();
--- a/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
+++ b/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
@@ -149,7 +149,7 @@ public final class Output<T extends TType> implements Operand<T> {
 
   public boolean isClosed() {
     Pointer handle = operation.getUnsafeNativeHandle(index);
-    return handle== null || handle.isNull() ;
+    return handle == null || handle.isNull();
   }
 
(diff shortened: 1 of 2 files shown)


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Output.java
```diff
@@ -147,6 +147,11 @@ Pointer getUnsafeNativeHandle() {
     return operation.getUnsafeNativeHandle(index);
   }
 
+  public boolean isClosed() {
```
Craigacp, line 150: Does this behave correctly on an `EagerSession`? Also it needs some javadoc and I think it'll fail the spotless check.

ramon-garcia: Fixed the formatting problem.

I tried the following small example with EagerSession, and it appears to reply correctly false:

```
import org.tensorflow.EagerSession;
import org.tensorflow.op.Ops;
import org.tensorflow.framework.initializers.Zeros;
import org.tensorflow.types.TFloat32;

public class App {
    public static void main(String[] args) {
        try (EagerSession s = EagerSession.create()) {
            Ops tf = Ops.create(s);
            Zeros<TFloat32> zeroInit = new Zeros<>();
            // y = a*x + b
            var x = tf.constant(new float[] {1.0f, 3.0f, 4.0f});
            var y = tf.constant(new float[] {2.5f, 6.5f, 8.5f});
            var b = tf.constant(1.0f);
            var a = tf.constant(1.8f);
            var ypred = tf.math.add(tf.math.mul(a, x), tf.stopGradient(b));
            var loss_gen = new org.tensorflow.framework.losses.MeanSquaredError();
            var los = loss_gen.call(tf, ypred, y);
            System.out.println("IsClosed " + los.asOutput().isClosed());
        }
    }
}

```
Running it, the output is:
2023-01-06 16:32:31.701434: I external/org_tensorflow/tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
**IsClosed false**

So it appears to work correctly.

Craigacp: Ok. Can you add the javadoc to the method?

ramon-garcia: Yes, I have just added the javadoc.

Craigacp (approved):

karllessard:
Thanks @ramon-garcia for your contribution!

ramon-garcia:
Thank you for accepting this contribution. It will make my work easier, now that I can stop gradient propagation.

PR #502, opened 2023-09-06
[linux-arm64] set "mkl_aarch64" bazel config for linux-arm64 platform

snadampal:
This is to enable TensorFlow core library bazel build with mkldnn (oneDNN) backend for linux-arm64 platforms.

### DIFF
--- a/tensorflow-core/tensorflow-core-api/build.sh
+++ b/tensorflow-core/tensorflow-core-api/build.sh
@@ -22,6 +22,11 @@ if [[ "${PLATFORM:-}" == macosx-arm64 ]]; then
   BUILD_FLAGS="$BUILD_FLAGS --config=macos_arm64"
 fi
 
+# Add platform specific flags
+if [[ "${PLATFORM:-}" == linux-arm64 ]]; then
+  BUILD_FLAGS="--config=mkl_aarch64"
+fi
+
 if [[ "${EXTENSION:-}" == *mkl* ]]; then
     BUILD_FLAGS="$BUILD_FLAGS --config=mkl"
 fi


Craigacp:
We're currently working on removing the bazel build and replacing it with the binaries built by TF for Linux & macOS platforms, so while we can merge this, it'll likely be replaced before the next release.

snadampal:
Hi @Craigacp , thanks for the review! When you say replacing the bazel build with binaries, does it mean we can build TF core outside the java repo and able to import the binaries? in that case it will automatically cover linux-arm64 platform as well, right? because it is fully supported in the TF repo.

Craigacp:
We'll repackage the TF binaries from the TF python wheels on several platforms. The Java binding itself still needs to be compiled so we're likely still limited to things we can easily build on GitHub actions as we don't have other build resources.

We're currently deciding if it will still be possible to build TF-Java entirely from scratch without downloading the binaries as we might still need to do a full build for Windows as that is packaged differently.

saudet:
There is no problem cross compiling JNI bindings generated with JavaCPP, so given binaries for the C++ API of TF Core itself, it should already pretty much work.

snadampal:
Thanks, @saudet and  @Craigacp. What is the tentative timeline to get the new build system support? if it is only few weeks away I'm fine with directly trying linux-aarch64 in the new system, otherwise I see value in having this PR on the mainline to enable source builds at least.

Craigacp:
We don't have a timeline on the new build support, nor are we planning on setting up cross compiling for Linux arm64 at the moment.

Craigacp (approved):

snadampal:
Thanks for approving the PR, @Craigacp . Please let me know what my next step is to get this merged.

Craigacp:
The CI is blocked as we're still using ubuntu-18.04 which doesn't work anymore. Once we've merged the other PR to bump the ci I'll run it on this and then we can merge it.

karllessard:
I've just merged both #503 and this one

snadampal:
Thanks, @karllessard !

PR #511, opened 2024-01-08
Add C API extension for custom gradient functions

karllessard:
This PR introduces a new C API extension directly edited from TF Java sources to enable capabilities not being exposed by the official TensorFlow C API. This simplifies the generation of the Java bindings for the C API when using JavaCPP since we now avoid parsing complex, internal and unstable C++ data structures.

The first capability enabled by this extension is the custom gradient Java functions, excluded so far in the `bazelcism` development branches.

The PR also introduces a series of small minor changes.


karllessard:
@saudet if you can also review the native and JavaCPP part, that would be great, thank you

tensorflow-core/tensorflow-core-native/src/main/resources/org/tensorflow/internal/c_api/include/tensorflow_adapters.h
```diff
@@ -1,42 +0,0 @@
...
-    operator T*()             { size = arr.size(); return (T*)arr.data(); }
-    operator Span<T>&() { return arr; }
-    operator Span<T>*() { return ptr ? &arr : 0; }
-    T* ptr;
-    typename Span<T>::size_type size;
-    void* owner;
-    Span<T> arr2;
-    Span<T>& arr;
-};
-
```
saudet, line 42: You probably don't want to erase all of that?

karllessard: We don't need it anymore, Git history will be there for reference later if needed. All of this was only required to parse C++ gradient func stuff.

saudet:
> @saudet if you can also review the native and JavaCPP part, that would be great, thank you

Looks alright, although I put "native" stuff in src/main/resources so that it is available to end users so they can use it for their own native libraries, but I don't think this is going to happen here so, I guess it doesn't matter

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/AbstractGradientAdapter.java
```diff
@@ -0,0 +1,95 @@
...
+  private static TF_Output toNativeOutputs(List<Operand<?>> outputs) {
+    var nativeOutputs = new TF_Output(outputs.size());
+    for (int i = 0; i < outputs.size(); ++i) {
+      var output = outputs.get(i).asOutput();
+      var nativeOutput = nativeOutputs.getPointer(i);
+      nativeOutput.oper(((GraphOperation)output.op()).getUnsafeNativeHandle());
+      nativeOutput.index(output.index());
+    }
+
+    // Native outputs must be freed by native invoker of this method, prevent deallocation from JVM
```
karllessard, line 89: @saudet something I was not sure about is this: is it the right way to tell JavaCPP to not free up a native array dynamically allocated from the JVM? Looks like I need to it in two steps. And I'm freeing up later the array pointer in the native code [here](https://github.com/tensorflow/java/pull/511/files#diff-0d6c014ab16494544e1c94bdabc96deb8d6cbb6d6b6f086117582a1791015ad5R67), using `free()`.

saudet: In this case, you should allocate with Pointer.malloc()

karllessard: That seems to work great, thanks! Pushed the changes

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/AbstractGradientAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+import org.bytedeco.javacpp.Pointer;
+import org.bytedeco.javacpp.PointerPointer;
+import org.bytedeco.javacpp.PointerScope;
+import org.tensorflow.internal.c_api.TF_GradFuncAdapter;
+import org.tensorflow.internal.c_api.TF_GraphId;
+import org.tensorflow.internal.c_api.TF_Operation;
+import org.tensorflow.internal.c_api.TF_Output;
+import org.tensorflow.internal.c_api.TF_Scope;
+
+/** Helper base class for custom gradient adapters <b>INTERNAL USE ONLY</b> */
```
Craigacp, line 31: Should we try to put it in a different package to delineate internal vs external API or does it need protected access into `Graph`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/TensorFlow.java
```diff
@@ -15,26 +15,29 @@
...
+import org.tensorflow.op.annotation.OpMetadata;
 import org.tensorflow.proto.OpList;
 
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.tensorflow.internal.c_api.global.tensorflow.*;
```
Craigacp, line 39: Star import.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/NativeScope.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Override
+  public NativeScope withDevice(DeviceSpec newDevice) {
+    return new NativeScope(
+        TF_NewScopeWithDevice(nativeScope, newDevice.toString()), graph, newDevice.toString());
+  }
+
+  @Override
+  public Scope withInitScope() {
+    throw new IllegalStateException("Can't add init operations in a gradient scope");
```
Craigacp, line 68: s/gradient/native/ ?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/NativeScope.java
```diff
@@ -0,0 +1,145 @@
...
+  public Scope withControlDependencies(Iterable<Op> controls) {
+    return withControlDependencyOps(StreamSupport.stream(controls.spliterator(), false)
+            .map(Op::op)
+            .collect(Collectors.toList()));
+  }
+
+  @Override
+  public Scope withControlDependencyOps(Iterable<Operation> controls) {
+    List<Operation> controlDeps =
+        StreamSupport.stream(controls.spliterator(), false).collect(Collectors.toList());
```
Craigacp, line 95: This juggling feels annoying given it's exactly the same juggling a few lines above. Maybe we should change the API (in a different PR) to accept a `List`?

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/NativeScope.java
```diff
@@ -0,0 +1,145 @@
...
+  @Override
+  public OperationBuilder apply(OperationBuilder builder) {
+    return builder;
+  }
+
+  @Override
+  public String getDeviceString() {
+    if (device == null) {
+      throw new UnsupportedOperationException(
+          "Can't get device string for gradient scope unless it has been explicitly set");
```
Craigacp, line 119: s/gradient/native/

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/TypedGradientAdapter.java
```diff
@@ -17,77 +17,41 @@
...
+
+  private final CustomGradient<T> gradient;
+  private final Class<T> opInputClass;
+  private final Constructor<T> ctor;
+
+  TypedGradientAdapter(CustomGradient<T> gradient, Class<T> opInputClass) {
+    super();
+    this.gradient = gradient;
+    this.opInputClass = opInputClass;
+    //noinspection unchecked
```
Craigacp, line 42: Shouldn't this be `@SuppressWarnings("unchecked")`?

tensorflow-core/tensorflow-core-native/src/main/java/org/tensorflow/internal/c_api/presets/tensorflow.java
```diff
@@ -58,31 +58,12 @@
...
-//            "tensorflow_adapters.h", // Keep this at the bottom, since it is not a real tensorflow header but one that we maintain (see main resources)
-// ---------------------------------------------------------
-// END OF CUSTOM GRADIENT
-// ---------------------------------------------------------
+            // Following are C API extensions maintained within TF Java, see src/main/native.
+            // Binding directly the C++ API with JavaCPP turned out to be too precarious between different releases,
+            // so it is simpler to write our own C API only exposing what we need from it.
+            "graph.h",
+            "scope.h",
+            "gradients.h",
```
Craigacp, line 66: Should we name these `tfj_gradients.h` etc?

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/gradients_impl.cc
```diff
@@ -0,0 +1,98 @@
...
+
+namespace tensorflow {
+    namespace java {
+        using namespace tsl;
+        using namespace std;
+
+        unordered_map<string, TF_GradFuncAdapter> g_grad_func_adapters;
+
+        template <typename T, typename U> T* struct_cast(U* ptr) {
+            return static_cast<T*>(static_cast<void*>(ptr));
```
Craigacp, line 33: This seems like crimes are occurring. Is the type information not present in the headers?

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/graph.h
```diff
@@ -0,0 +1,40 @@
...
+/// Returns the unique identifier of the graph `g`
+TF_CAPI_EXPORT extern TF_GraphId TF_GetGraphId(const TF_Graph* g);
+
+/// Remove an operation from the name map of the graph `g`, so that it cannot be reversely looked up by name.
+/// This is particularly useful for preventing custom gradient operations to pollute the graph namespace.
+TF_CAPI_EXPORT extern void TF_UnmapOperationName(TF_Graph* g, TF_Operation* operation);
+
+#ifdef __cplusplus
+} /* end extern "C" */
+#include "graph_impl.cc"
```
Craigacp, line 37: My C++ is very rusty, so is it worth a comment as to why we're including the cc file into the header?

Craigacp (changes requested):
A few small things to look at.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Graph.java
```diff
@@ -71,6 +60,8 @@
 import org.tensorflow.types.TString;
 import org.tensorflow.types.family.TType;
 
+import static org.tensorflow.internal.c_api.global.tensorflow.*;
```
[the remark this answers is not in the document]
karllessard: Yeah, I noticed them and deliberately let them there because... star imports are fun sometimes :) But ok ok, I'll revert this back

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/AbstractGradientAdapter.java
karllessard: It needs access to protected method in `Graph`, yes. That is also why the abstract class is not in the same package as other gradient stuff (`org.tensorflow.op`).

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/NativeScope.java
karllessard: Yes, I did a few "optimization" around but tried to keep the Java implementation as much as it was before in this PR.

tensorflow-core/tensorflow-core-native/src/main/java/org/tensorflow/internal/c_api/presets/tensorflow.java
karllessard: I'm questioning this too... `tf_` or `tfj_`.. I kind of like that our extensions _blends_ with the original C API, that is why I left it the class name with the `TF_` prefix, though it makes it more explicit it is custom to TF Java using a different prefix (though do users care?)

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/gradients_impl.cc
karllessard: Haha! I actually got that from the TensorFlow source code [here](https://github.com/tensorflow/tensorflow/blob/9d637f69f699c0c422716b56153a8b27b681891a/tensorflow/c/c_api.cc#L658). Now, it does not mean that if it is in TensorFlow code that it is not evil ;) Actually it is used to carry out C++ pointers into a C struct that is left undefined in the headers, yes. That's how the C API does it too.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/AbstractGradientAdapter.java
Craigacp: Every so often I wish for friend classes in Java. Oh well.

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/graph.h
karllessard: That's the only way I found to include custom C++ code during JavaCPP compilation. @saudet , is there another way to provide non-headers files to be compiled with the JNI code other than including it as part of a header that is parsed?

Though I was thinking also that this could be useful if we want to exclude these CC files from the Windows build (as we still don't know if the dependent C++ API headers will be available).

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/gradients_impl.cc
Craigacp: Might be worth putting a comment pointing at the C API, because that cast is just nasty and I'd like to remember to blame someone else for it.

tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/op/TypedGradientAdapter.java
karllessard: That was there before...  I'll check

tensorflow-core/tensorflow-core-native/src/main/java/org/tensorflow/internal/c_api/presets/tensorflow.java
Craigacp: I guess it doesn't matter too much. They'd likely get new names if we tried to contribute it back to the TF C API anyway.

tensorflow-core/tensorflow-core-native/src/main/native/org/tensorflow/internal/c_api/graph.h
saudet: We could also do that by putting those files in `@Platform(include=...` but removing them in `init()` only when called from the Parser, although I don't think we'd have any problems just leaving them there, but anyway if you did those kinds of hacks are mostly for cases when we don't want to modify the original code. The way you did it is fine

karllessard: So @Craigacp you're also fine with this, I suggest we keep it like that for now

Craigacp: Sure, but it would be good to comment that it's due to the way JavaCPP builds it, in case we eventually get more control over that build so we can go back to a more traditional C++ layout.

karllessard:
Ok @Craigacp , pushed my changes. I've also switched to the TFJ_ prefix finally

Craigacp (approved):
LGTM

PR #512, opened 2024-01-15
Fix custom gradient lookup bug

karllessard:
The original code was doing the lookup on the node name, not the node type.

+ clean up in Bazel scripts


Craigacp (approved):
LGTM

PR #513, opened 2024-01-18
Revert Conv2d backprop to V1

karllessard:
V2 versions of Conv2dBackprop* operations adds unnecessary complexities to the API and seem to not be used (yet) by the Python API, so we'll make again the V1 version as the main API for these operations.


Craigacp (approved):
Will the pbtxt changes be overwritten by regenerating the ops?

karllessard:
> Will the pbtxt changes be overwritten by regenerating the ops?

No, these files are not overwritten when regenerating the op. The generator will only create them when they are missing for a new op, but after that we can edit them directly.

PR #514, opened 2024-01-19
Build core-platform only on deploy

karllessard:
Small PR to build `tensorflow-core-platform` only when the `deploying` profile is active (i.e. when building and pushing artifacts to Sonatype). This way, users can intuitively run `mvn install` locally and they won't get error that some artifacts are missing (right now the "patch" is to pass the `-Djava.platform.host` parameter to get rid of this error).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ For dependencies, we can use anything compliant with [this list](https://opensou
 
 ## Building
 
-To build all the artifacts locally, simply invoke the command `mvn install -Djavacpp.platform.host` at the root of this repository (or the Maven command of your choice). It is also
+To build all the artifacts locally, simply invoke the command `mvn install` at the root of this repository (or the Maven command of your choice). It is also
 possible to build artifacts with support for CUDA® by adding the `-Djavacpp.platform.extension=-gpu` argument to the Maven command.
 
 ### JDK 16+
@@ -120,7 +120,7 @@ Once these steps have been executed, you can run `mvn install` to build the new 
 ### Generating Java Bindings
 
 After upgrading the TensorFlow library, you need to regenerate all Java bindings that depends on the native code. That includes Java protos, C API bindings (JavaCPP) and
-operator classes. You can trigger the regeneration of these bindings with the Maven command `mvn clean install -Pgenerating -Djavacpp.platform.host`.
+operator classes. You can trigger the regeneration of these bindings with the Maven command `mvn clean install -Pgenerating`.
 
 This will trigger a small Bazel build of the TensorFlow sources to regenerate the Java protos, so make sure your [environment](CONTRIBUTING.md#native-builds) is setup properly.
 
--- a/tensorflow-core/pom.xml
+++ b/tensorflow-core/pom.xml
@@ -34,6 +34,7 @@
     <module>tensorflow-core-native</module>
     <module>tensorflow-core-generator</module>
     <module>tensorflow-core-api</module>
+    <module>tensorflow-core-platform</module>
   </modules>
 
   <properties>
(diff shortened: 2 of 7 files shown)


karllessard:
Another option though is to play with the list of native dependencies in the POM of `tensorflow-core-platform` itself to include only the current platform when `deploying` profile is not active. This way, you can product a `tensorflow-core-platform` locally and depend on it in your local projects. Thoughts?

Craigacp:
The only use of platform now is as a metaproject to pull together all of the different platforms' binaries, so having it occasionally only contain a single platform seems like it might cause more trouble. I think this solution is fine.

Craigacp (approved):

karllessard:
Ah oops :D I just did and submitted the other option... I though you might still want to include "tensorflow-core-platform" as a dependency in your local projects, so you don't need to include individually `-native` and `-api`... I can revert it to the original option, sorry I didn't saw your comment before

karllessard:
I still believe it is useful the new way, thinking of it... imagine you want to run locally some project that depends on `tensorflow-core-platform` that you don't own. But you are running on Linux Arm64 so you do a local build of TF-Java to be able to run it. If the build does not produce `tensorflow-core-platform`, it will continue to fail to run. Does that make sense?

Craigacp:
Sure.

Craigacp (approved):

saudet:
Deploying also happens for developers that work on Mac or Windows but that deploy to Linux machines. You're going to have users wondering why their deployments suddenly start failing. That's why I personally prefer having it download everything by default, because then we don't have to muck around with profiles that don't work in situations X or circumstances Y with tools Z, such as Gradle or sbt.

karllessard:
I just realized that I can't add dependencies in the profile like I'm doing now, unless I'm unwrapping them in `tensorflow-core-platform` to make it a uberjar of native libraries (which could work too). There are other ways to do it as well, I'll push a new version soon.

I'm not sure to understand your point @saudet . Users will still be able to develop on Mac or Windows and deploy on Linux when using the artifacts we are distributing. Here we just impact the users building TF Java locally. So chances are that they just want to be able to run their project locally on an unsupported platform (unless they also plan to have their own custom TF Java distribution).

Basically I'm doing the same as we are doing when are setting `-Djavacpp.platform.host`, but I now want this behaviour to be the default so that a simple intuitive `mvn install` on a local machine works, without the need of specifying anything else.

saudet:
Like I said, the big problem here is that there are ***no other tools than Maven that support profiles***. You will receive a ton of hate mail from Gradle and sbt users. The question is: Is this alright with you? If the answer is yes, then by all means, please go ahead! :) I'm not debating anything here. I'm just stating facts. Even Maven itself does not recommend using profiles for **anything**. If there is something you do not understand please tell me what that is. Just repeating over and over exactly the same thing about why using profiles is better isn't going to help move the conversation forward.

karllessard:
@Craigacp , I've pushed a new version, can you please take a look? I'm just making `javacpp.platform.host` the default behaviour.

Craigacp:
Looks fine to me.

Craigacp:
> Like I said, the big problem here is that there are _**no other tools than Maven that support profiles**_. You will receive a ton of hate mail from Gradle and sbt users. The question is: Is this alright with you? If the answer is yes, then by all means, please go ahead! :) I'm not debating anything here. I'm just stating facts. Even Maven itself does not recommend using profiles for **anything**. If there is something you do not understand please tell me what that is. Just repeating over and over exactly the same thing about why using profiles is better isn't going to help move the conversation forward.

This is about how our build works, it doesn't affect downstream users of TF-Java using other build systems as it will pull in published or deployed artifacts. We're not going to migrate TF-Java to use gradle or sbt, so that's not relevant.

saudet:
Yes this change will affect downstream users big time. When you're done not listening to me, a better solution that doesn't break everything for Gradle, sbt, etc would be to use os-maven-plugin: https://github.com/bytedeco/javacpp-presets/issues/846#issuecomment-1482762303

Craigacp:
Why will it affect them? They will continue to depend on `tensorflow-core-platform` as before, which will pull in all the binaries that we've published for that version along with `tensorflow-core-api`.

saudet:
No it won't! If you don't believe me, try it yourself

PR #518, opened 2024-02-12
Exclude TSL linkage to solve Windows build problem

karllessard:
While we are waiting for a fix to solve the [Windows TSL symbols problem](https://github.com/tensorflow/tensorflow/issues/62579), it turns out that the JavaCPP bindings that were generating linkage problem with the C Lib are not being used directly by our Java API. Skipping them allows us now to build TF Java successfully on Windows. See this run:

https://github.com/karllessard/tensorflow-java/actions/runs/7876348680


Craigacp (approved):
LGTM

PR #521, opened 2024-02-23
Refactoring native builds and Java binding generation

karllessard:
As part of a everlasting initiative, we are moving out of building TensorFlow natively using Bazel to downloading the official builds being distributed in the Python wheels.

As we shuffled things around, we took an opportunity as well to facilitate the TensorFlow upgrades and the generation of all the Java bindings, including protos and ops.

This releases introduces a few breaking changes, the major one being that the classified JARs containing the native binaries is now `tensorflow-core-native`, and `tensorflow-core-api` now becoming a pure Java API.

Version is updated to `1.0.0-SNAPSHOT` as we are targeting this codebase to become the first official release of TensorFlow Java (1.0.0) in a very near future.


Craigacp (approved):
So much less bazel.

karllessard:
Merged, thank you all!

PR #525, opened 2024-03-01
Pom profile changes

Craigacp:
Switch `tensorflow-core-platform` so it's only generated when the `deploying` profile is set. Lets the profiles default through so the platform has the correct values when pulled in without any profiles activated.


saudet:
As I pointed out before, this isn't going to work with other tools than Maven...

Craigacp:
It worked when we tested it with gradle earlier today as it defaults in the pom and platform pulls in all the dependencies.

tensorflow-core/pom.xml
```diff
@@ -84,588 +85,13 @@
       -->
       <id>javacpp-platform-host</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
         <property>
-          <name>javacpp.platform.host</name>
+          <name>!javacpp.platform</name>
```
karllessard, line 89: There is no need of having this profile if we set the default value of `javacpp.platform` to `${os.name}-${os.arch}` by default in the pom `properties`, and simply let user override it if needed

karllessard:
> As I pointed out before, this isn't going to work with other tools than Maven...

The new strategy here @saudet is that we'll build `tensorflow-core-platform` only when releasing TFJava as part of the CI/CD. The local build will skip it, hence removing the need of playing with platform properties like we were doing originally.

saudet:
Yeah, sure, feel free to experiment, but do test with other tools is all I can recommend. In the end, I think you'll find the the way I'm doing it in the presets is the least worst

tensorflow-core/pom.xml
```diff
@@ -74,6 +72,9 @@
       <properties>
         <javacpp.platform>${os.name}-${os.arch}</javacpp.platform>
       </properties>
```
karllessard, line 74: This should be removed as well, as it will be set in the top properties on this file

tensorflow-core/tensorflow-core-platform/pom.xml
```diff
@@ -41,25 +41,25 @@
       <groupId>org.tensorflow</groupId>
       <artifactId>tensorflow-core-native</artifactId>
       <version>${project.version}</version>
-      <classifier>${javacpp.platform.linux-x86_64.extension}</classifier>
+      <classifier>linux-x86_64</classifier>
```
karllessard, line 44: Maybe we should keep using properties but the with an extension here (e.g. `${javacpp.platform.linux-x86_64`), just so we know the values are consistent. Not strong opinion either way.

karllessard (approved):

PR #528, opened 2024-03-08
Adding casts to the if test so it passes on GPU

Craigacp:
`ops.math.neg` causes eigen to segfault on GPUs now we've upgraded to TF 2.15. This PR fixes the one test where we do that so that there are casts around it. Not sure how to indicate that users shouldn't use this op on GPU as the op definition accepts ints and it's autogenerated.


karllessard:
Was this reported to TF core? Or is it an eigen problem?

karllessard:
As we plan to upgrade to TF2.16, let see if the bug is still there before merging this

Craigacp:
The if test still fails on GPU with TF 2.16 (#531).

karllessard (approved):
Ideally we would wrap this `.neg` operation with our own that would take care of the casting. But let's merge it so at least the build won't fail on GPUs.

PR #531, opened 2024-03-22
Upgrade to TF 2.16.1

karllessard:


tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/generator/op/OpGenerator.java
```diff
@@ -335,15 +335,15 @@ private void createApiDef(OpDef opDef, File apiDefFile) throws IOException {
             visibility = ApiDef.Visibility.SKIP;
             apiDef.setVisibility(visibility);
             break;
-          case "D":
-          case "d":
+          case "B":
+          case "b":
```
Craigacp, line 339: Why did this option get renamed to `b`? It's still documented as `d` on line 319.

Craigacp (approved):
LGTM aside from the op gen option rename.

tensorflow-core/tensorflow-core-generator/src/main/java/org/tensorflow/generator/op/OpGenerator.java
karllessard: Oops, that was a left over, reverted it

PR #533, opened 2024-03-31
Make all TF operations visible via the Ops API

karllessard:
Many operations in TensorFlow are by default "hidden" to the public API. While this default visibility setting should be set agnostically of any language bindings, often it seems that it is hidden only because the Python API is wrapping it. Marking all operators as "visible" in the TF Java API definitions makes them more easily accessible to Java users.


tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/EstimatorOps.java
```diff
@@ -0,0 +1,607 @@
...
+import org.tensorflow.types.TInt64;
+import org.tensorflow.types.TString;
+import org.tensorflow.types.family.TType;
+
+/**
+ * An API for building {@code estimator} operations as {@link Op Op}s
+ *
+ * @see {@link Ops}
+ */
+public final class EstimatorOps {
```
Craigacp, line 61: Didn't estimators get removed in TF 2.16? Or is this a different meaning of estimator? It looks like it's about boosted trees and all that stuff got move out into tfdf.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/RiscOps.java
```diff
@@ -0,0 +1,798 @@
...
+import org.tensorflow.types.TInt32;
+import org.tensorflow.types.family.TNumber;
+import org.tensorflow.types.family.TType;
+
+/**
+ * An API for building {@code risc} operations as {@link Op Op}s
+ *
+ * @see {@link Ops}
+ */
+public final class RiscOps {
```
Craigacp, line 83: Any idea what this is for? It looks like it's just some core ops but for a different CPU, but we don't know what CPU.

Craigacp (commented):
Most of this looks fine, but some of them probably shouldn't be visible (e.g. EstimatorOps and RiscOps).

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/EstimatorOps.java
karllessard: Yeah... we have different options:
    1. Preserve the operator and the Ops classes but marking them as deprecated (the generator used to support that and I think that is still the case)
    2. Leave the operator classes but not the Ops classes
    3. Discard all of them

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/RiscOps.java
karllessard: Google (the search engine) tells me it is related to TensorFlow Lite on RISC-V: https://hackmd.io/@nctu-cas-lab/BJ75A0IsK

So is not making much sense for TFJava?

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/EstimatorOps.java
karllessard: Yep, looks like everything under `estimator` in Python is deprecated: https://www.tensorflow.org/versions/r2.15/api_docs/python/tf/estimator

Since we haven't released anything officially yet, I suggest that we skip them entirely

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/RiscOps.java
Craigacp: Yeah, I think we should hide those.

tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/EstimatorOps.java
Craigacp: Sounds good.

Craigacp (approved):

Issue #479, opened 2022-12-02
Make Graph accessible from Session


arianmaghsoudnia:
**System information**
- TensorFlow version : 0.4.0
- Are you willing to contribute it (Yes/No): Yes

**Describe the feature and the current behavior/state.**
I would like to have access to the [Graph attribute](https://github.com/tensorflow/java/blob/851fcf6b205a5ad5028b6f8afa4b2fe16a1cb0f9/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java#L706) of the Session. Is there a reason it is private?

**Will this change the current api? How?**
Yes. Makes an existing attribute public.

**Who will benefit with this feature?**
If you have a loaded session and need to access the detail of your graph.

karllessard:
I guess it is safe to expose it, yes, I don't see why not. Would you be interested to push a PR for this?

AyushABD-17:
code base is in java?

manojava98:
Hi, @karllessard I can raise PR for this one, kindly let me know if this is okay with you. Thanks in advance!

karllessard:
Yes @manojava98 , please do!

manojava98:
Hi @karllessard I have raised the PR #535  . Kindly, check

Craigacp:
Fixed in `master`.

PR #535, opened 2024-04-15
Graph is changed to public in Session.java

manojava98:
made graph as public in sessions.java  #479


tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/Session.java
```diff
@@ -707,7 +707,7 @@ Graph graph() {
     return graph;
   }
 
-  private final Graph graph;
```
karllessard, line 710: Please @manojava98 , leave the field private and add a public getter named `graph()` to return that instance instead.

karllessard:
Thanks @manojava98 !

PR #566, opened 2024-11-08
Release 1.0.0

karllessard:
This PR applies small changes in order to do the TF Java 1.0.0 release. 

It also includes a small change for fixing the native builds, that was failing due to missing absl headers. For now we will download these headers from the wheel.


Craigacp (approved):
LGTM

